PR #14, opened 2016-05-13
Bugfix/add documentation

zackurben:


randallknutson:
Not sure why tests are failing but this is only a documentation change so it won't affect the tests.

PR #41, opened 2017-02-13
Return correct URL for uploaded file.

douglaselee:
Return correct URL for uploaded file so image will display properly if that option was selected.  Also add JWT request header to POST.  This is useful if someone has added a route to the Formio server and is uploading files to it.

### DIFF
--- a/src/providers/storage/url.js
+++ b/src/providers/storage/url.js
@@ -1,4 +1,5 @@
 var Promise = require("native-promise-only");
+var Formio  = require("../../Formio");
 var url = function(formio) {
   return {
     title: 'Url',
@@ -29,7 +30,7 @@ var url = function(formio) {
             resolve({
               storage: 'url',
               name: fileName,
-              url: xhr.response.url,
+              url: xhr.responseURL + '/' + fileName,
               size: file.size,
               type: file.type
             });
@@ -49,6 +50,7 @@ var url = function(formio) {
         }
 
         xhr.open('POST', url);
+        xhr.setRequestHeader('x-jwt-token', Formio.getToken());
         xhr.send(fd);
       });
     },


travist:
Looks good! Thanks!

douglaselee:
Hi Travis.  Happy to help, thanks for Formio!

travist:
I just fixed this with version 1.5.3.  Thanks everyone.

douglaselee:
Thanks Travis.  Sorry for the screw up.

travist:
No worries @douglaselee. I appreciate your contribution.

douglaselee:
Hi Travis.  

The variation...
var Formio  = require("../../formio");
... is not working for me in Windows/Chrome.

When required as "../../formio" the Formio variable is a plain object with no getToken function.

Maybe because formio.js contains...
var Formio = function(path) {
...

Meanwhile the formio (lowercase) parameter passed to the url function is a Formio object but it does not contain a getToken funtion.

When required as "../../Formio" the Formio variable is a function(path) with lots of properties including the getToken function.

I don't know what to do about this so maybe we should just remove this part of the change for now (lines 2 and 53).  Do you want me to create a pull request for that?

Thanks again for your patience.

travist:
I would rather trudge forward and figure out why Windows is not liking the change vs. back it out. Once we figure that out, we can then get another pull request and release to fix it.

douglaselee:
That's a good long term strategy but meanwhile file uploads to URLs won't work because the getToken call will fail and the call to POST will be skipped entirely.

travist:
So are you having issues with the latest renderer getting these changes? It may be that what is happening is you are using a renderer that does not yet have these fixes in place. I am rolling out a new renderer as we speak so you can maybe try with that one which includes the latest formio.js.

douglaselee:
I'm up to date on formio.js and 3 commits behind ngFormio.  Looks like those 3 commits are about dates and times.

travist:
Right, but when I rebuild the renderer I pull in the latest formio.js. The last release of the renderer 2.9.9 included the broken version of formio.js. I just pushed up a new version 2.9.10 which includes the fixed version. Let me know if that works for you. If not, then we can go back to the drawing board on why it isn't working in IE.

douglaselee:
Building formio.js 1.5.3 locally and pulling that into ngFormio 2.9.9 also built locally.  Running Chrome on Windows.  What's IE?

travist:
I meant Windows.  Hmmm... I may have to pull out the ol' windows VM to try and reproduce and determine fix for this.

douglaselee:
Thanks Travis!

PR #54, opened 2017-03-20
Add class name in formio.form and formio.wizard

GuillaumeSmaha:
To add wizard support on ng2-formio, I need to know the current class name on the object `this.formio`


GuillaumeSmaha:
I miss your commits for wizard....

PR #59, opened 2017-03-23
Add missing pattern validator

GuillaumeSmaha:


travist:
Thanks @GuillaumeSmaha!

PR #63, opened 2017-03-24
Feature: Emit prev and next page events

GuillaumeSmaha:


src/formio.wizard.js
```diff
@@ -22,11 +22,21 @@ export class FormioWizard extends FormioForm {
...
   prevPage() {
     this.setPage(this.page - 1);
+    this.onPrevPage({page: this.page, submission: this.submission});
+  }
+
+  onPrevPage(submission) {
+    this.emit('prevPage', submission);
+  }
+
+  onNextPage(submission) {
```
travist: Why not just emit directly from the ```prevPage``` and ```nextPage``` methods without introducing two new superfluous methods.

GuillaumeSmaha: Ok, I will update it

GuillaumeSmaha:
@travist It's updated.

PR #70, opened 2017-03-28
Returns promises on change page functions and add wizard test

GuillaumeSmaha:


test/wizards/simple.js
```diff
@@ -1,3 +1,6 @@
+import _map from 'lodash/map';
```
travist: Thank you for adding tests!  👏

test/wizards/simple.js
```diff
@@ -181,6 +204,283 @@ module.exports = {
...
+          done(error);
+        })
+        .then(() => {
+          // Check start page
+          assert.deepEqual(form.data, {
+            a: '',
+            b: '',
+            c: '',
+            d: ''
+          });
```
GuillaumeSmaha: @travist I don't know if this a bug or not. But after cancel, I have to compare with an object containing empty string instead of an empty object.
I have the same issue at line https://github.com/formio/formio.js/pull/70/files/02c3a480e1f5bf0d4e7a0ff0db2a7cd7a22aca01#diff-2d594189a79b299c2731f90be3b29e25R349 .  I have to define `d` as an empty string

travist:
I just pulled in these changes to master.

PR #71, opened 2017-03-28
Fixed radio selects, and require libraries.

travist:


StephenOTT:
@travist does this fix: https://github.com/formio/ng2-formio/issues/79 ?

PR #72, opened 2017-03-29
Fix HTML component: check if attribute name is not empty

GuillaumeSmaha:


GuillaumeSmaha:
Should fix https://github.com/formio/ng2-formio/issues/80

travist:
Thank you!

PR #74, opened 2017-03-31
Add nextPage in panel components to allow conditional next page

GuillaumeSmaha:
Manage previous page button to keep the form flow, manage submit in intermediate page with null value in nextPage function and dynamicly update the next button into a submit button.

The server `form.io` will certainly need an update without the `Form Component` that you describe in the issue https://github.com/formio/formio/issues/296 to avoid to check a bypassed page which contains required fields.

### DIFF
--- a/src/components/panel/fixtures/comp1.js
+++ b/src/components/panel/fixtures/comp1.js
@@ -76,6 +76,7 @@ export const component = {
       "input": true
     }
   ],
+  "nextPage": null,
   "theme": "default",
   "title": "User Information",
   "input": false,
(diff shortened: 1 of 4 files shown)


travist:
Looks good. Is this reverse compatible with the previous wizard implementation?

GuillaumeSmaha:
Yes, the page management use the "nextPage" attribute and without it, the function always returns `page + 1` (https://github.com/formio/formio.js/pull/74/files#diff-e644f27b6fa752d5b2b05362b1e018bcR45 )

travist:
Looks good. Thanks!

Issue #79, opened 2017-04-18
differences between ng-formio and formio.js when loading a submission


andrewjfoster:
when using the full submission url for a form submission, the submission loads and renders fine in an angular 1 application using ng-formio.    When doing the same using formio.js (which this library wraps) the form and submission have issues.

1) when using drop down select boxes with JSON as the list type the drop down fails to populate unless the json is string encapsulated.
2) submission values fail to load in the select box fields and select list fields.

It appears the issues relate to formio.js.     the ng-formio library doesn't appear to wrap formio.js like ng2-formio does.

edenware:
With RC 19 this issue is just about resolved.

The core issue is the way that submissions are interpreted.

The JSON select list values fail to parse because they are already parsed.    a quick #hack was to add the following code after scripts are loaded (i.e. in the main client index.ts with the ng2-formio starter project)

let otbParse = JSON.parse;
JSON.parse = function(value: any) {
  if (typeof value === 'string') {
    return otbParse(value);
  }
  else
    return value;
}

adding this meant that all submitted form values matched when recalled apart from select boxes.   The select value recalled from a submission was already parsed as a single object with boolean properties of the key select box values.     To fix this altering the get value function to the following fixes the issue:

--- from SelectBoxes.js
/**
     * Set the value of this component.
     * @param value
     */

  }, {
    key: 'setValue',
    value: function setValue(value, noUpdate, noValidate) {
      var tempValue = [];
      var valueIsArray = (0, _isArray3.default)(value);

      (0, _each3.default)(this.inputs, function (input) {
        if (valueIsArray)
          input.checked = value.indexOf(input.value) !== -1;
        else {
          input.checked = value[input.value] == undefined ? false : value[input.value];
          tempValue.push(input.value);
        }
      });
      
      this.value = valueIsArray ? value : tempValue;
      
      if (!noUpdate) {
        this.updateValue(noValidate);
      }
    }
  }]);

I hope this helps in resolving these issues.

PR #100, opened 2017-06-03
Bugfix/submission load

edenware:
Resolved a couple of issues that occur when reloading submissions to a form.

1) Select Boxes fail to reload the selected values.
2) Select component fails to reload the selected values

see issue #79


dist/formio.js
```diff
@@ -39,10 +39,10 @@ require('whatwg-fetch');
 var EventEmitter = require('eventemitter2').EventEmitter2;
 var copy = require('shallow-copy');
 
-/**
- * The Formio interface class.
- *
- *   let formio = new Formio('https://examples.form.io/example');
```
travist: Why were all of these changed? Do you have a setting in your IDE that is changing each of these comments? Please fix before I can accept this pull request. Thank you for your contributions!

edenware: Sorry, I did gulp build before checking in the code changes.    I can't see any difference in the comments but somehow git does....   Anyway,    The main two files that contain the code changes are src/components/selectboxes/SelectBoxes.js and src/components/select/Select.js.    The other 8 files are the result of gulp build.    I'll revert them and then re-submit.

PR #102, opened 2017-06-03
Add SPAN around label text for better styling hooks

### LINKED ISSUE #101: Add SPAN around label text for better styling hooks
- [ ] `radio` inputs should have a SPAN tag for textNode
- [ ] `checkbox` inputs should have a SPAN tag for textNode

criticalmaas:
After looking to better enable styling radio and checkboxes for A11y and touch screens, I made a simple change to the rendering to wrap the label textNode with a SPAN tag. I submit for your consideration.

Closes #101

### DIFF
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -48,15 +48,15 @@ export class CheckBoxComponent extends BaseComponent {
     });
 
     // Create the SPAN around the textNode for better style hooks
-    this.labelText = this.ce('span', 'span');
+    this.labelSpan = this.ce('labelSpan', 'span');
 
     if (this.info.attr.id) {
       this.label.setAttribute('for', this.info.attr.id);
     }
     this.addInput(input, this.label);
     if (!this.options.inputsOnly) {
-      this.labelText.appendChild(this.text(this.component.label));
-      this.label.appendChild(this.labelText);
+      this.labelSpan.appendChild(this.text(this.component.label));
+      this.label.appendChild(this.labelSpan);
     }
     container.appendChild(this.label);
   }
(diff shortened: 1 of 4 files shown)


src/components/checkbox/Checkbox.js
```diff
@@ -46,12 +46,17 @@ export class CheckBoxComponent extends BaseComponent {
     this.label = this.ce('label', 'label', {
       class: 'control-label'
     });
+
+    // Create the SPAN around the textNode for better style hooks
+    this.labelText = this.ce('span', 'span');
```
travist: Can you change the first string here to be descriptive id of what this is. I would call it ```labelSpan``` so change to this please.

```
this.labelText = this.ce('labelSpan', 'span');
```

This label is used to hook into the rendering from an external library so that it can control every element that is created within the renderer. Having a descriptive label helps that external library know what it is changing.

src/components/checkbox/Checkbox.spec.js
```diff
@@ -10,10 +10,20 @@ describe('Checkbox Component', function() {
       for (let i=0; i < inputs.length; i++) {
         assert.equal(inputs[i].name, 'data[' + comps.comp1.key + ']');
       }
+      Harness.testElements(component, 'span', 1);
       done();
+      
     });
   });
 
+  it('Span should have correct text label', function(done) {
```
travist: Thank you for writing tests!

criticalmaas: done!

src/components/checkbox/Checkbox.js
criticalmaas: done!

PR #110, opened 2017-06-09
Share jsonLogic with ng-formio factory

douglaselee:
ng-formio FormioUtils factory refers to jsonLogic in formio.js version but it's not set.
Can cause problems when needed to evaluate things like JSON Next Page.


travist:
Thank you @douglaselee!

douglaselee:
Happy to help @travist !

Issue #126, opened 2017-07-05
Button component custom javascript integration


krixerx:
We would very like to execute javascript script when button is clicked.

Can we add a new attribute to button schema like "script"?

douglaselee:
I think what you're asking for was just added to ngFormio and ngFormBuilder but it hasn't been added to the core renderer in formiojs yet.

douglaselee:
May be implemented by pull request #145.

PR #128, opened 2017-07-05
#126 customScript attribute in button component

krixerx:
#126 customScript attribute in button component

### DIFF
--- a/src/components/button/Button.js
+++ b/src/components/button/Button.js
@@ -42,6 +42,9 @@ export class ButtonComponent extends BaseComponent {
       this.label = this.text(this.component.label);
       this.element.appendChild(this.label);
     }
+    if (this.component.customScript) {
+        this.customScript = this.text(this.component.customScript);
+    }
     if (this.component.action === 'submit') {
       this.on('submitButton', () => {
         this.loading = true;
@@ -73,6 +76,9 @@ export class ButtonComponent extends BaseComponent {
           break;
         case 'reset':
           this.emit('resetForm');
+          break;
+        case 'script':
+          eval(this.component.customScript);
           break;
         case 'oauth':
           console.log('OAuth currently not supported.');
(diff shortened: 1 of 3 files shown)


travist:
I am curious why this is a necessary feature when I believe you can achieve everything you want with using the events like this.

```
Formio.createForm('https://myproject.form.io/myform').then((form) => {
  form.on('customEvent', () => {
    // Execute custom code here....
  });
});
```

krixerx:
The reason is to build complex component.
Edit view + grid. In every grid row there are buttons (edit, delete) and below grid "add new".
It is more convenient to add (edit, delete) javascript code directly into button instead of playing with events.
And event doesn't support parameters. It is difficult to say to event which datagrid row i want to edit on edit panel.

src/components/button/Button.js
```diff
@@ -42,6 +42,9 @@ export class ButtonComponent extends BaseComponent {
       this.label = this.text(this.component.label);
       this.element.appendChild(this.label);
     }
+    if (this.component.customScript) {
+        this.customScript = this.text(this.component.customScript);
```
travist: Why is this needed?  I don't see that this does anything other than create a DOM element that has the contents of the script. I believe this would still work without this script.

src/components/button/Button.js
```diff
@@ -74,6 +77,9 @@ export class ButtonComponent extends BaseComponent {
         case 'reset':
           this.emit('resetForm');
           break;
+        case 'script':
+          eval(this.component.customScript);
```
travist: Needs to be wrapped within a ```try { } catch { }``` statement.

src/components/complexdatagrid/ComplexDataGrid.js
```diff
@@ -0,0 +1,368 @@
+import _each from 'lodash/each';
```
travist: This needs to be a separate pull request.

PR #133, opened 2017-07-15
Fixed the custom layout

Abdelaziz-Khabthani:
The custom layout, where the user can add margins to the component, was ignored.


travist:
Awesome! thank you!

PR #136, opened 2017-07-16
Fixed required validation for multiple-select

Abdelaziz-Khabthani:
Multiple select is a special case, it's the only component when flagged as multiple it will only contain one input.

### DIFF
--- a/src/components/Validator.js
+++ b/src/components/Validator.js
@@ -36,7 +36,7 @@ export var Validator = {
     _each(component.validators, (name) => {
       if (this.validators.hasOwnProperty(name)) {
         let validator = this.validators[name];
-        if (component.component.multiple && _isArray(value)) {
+        if ((component.component.multiple && _isArray(value)) && !(component.component.type === 'select' && validator.key === 'validate.required')) {
           _each(value, (val) => {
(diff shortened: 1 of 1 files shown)


src/components/Validator.js
```diff
@@ -36,7 +36,7 @@ export var Validator = {
     _each(component.validators, (name) => {
       if (this.validators.hasOwnProperty(name)) {
         let validator = this.validators[name];
-        if (component.component.multiple && _isArray(value)) {
+        if ((component.component.multiple && _isArray(value)) && !(component.component.type === 'select' && validator.key === 'validate.required')) {
```
travist: I would prefer we do this in an OOP way.  Where there is a method within ```BaseComponent``` called ```isMultiple``` that looks something like this.

```
isMultiple(value) {
  return this.component.multiple && _isArray(value);
}
```

And then in the ```SelectComponent``` you would do something like...

```
isMultiple(value) {
  return false;
}
```

This could then be changed to...

```
if (component.isMultiple(value)) {
```

Which is much cleaner IMO.

Abdelaziz-Khabthani:
Yes you are right, i will make another pull request !

PR #137, opened 2017-07-17
Fixing multiple select required validation.

Abdelaziz-Khabthani:


travist:
Now looking at this, I think the method needs to be renamed. ```isMultiple``` would be interpreted as "is this component configured for multiple" which would simply be ```this.component.multiple```. I think it should be called ```validateMultiple``` since that is closer to what it is doing. Once you make that change, I will commit it. Thanks!

PR #139, opened 2017-07-18
Bugfix/FOR-611

YuryRybakSoftensity:


src/utils/index.js
```diff
@@ -206,6 +211,88 @@ const FormioUtils = {
...
+   *   Value formatted for Currency component.
+   */
+  formatAsCurrency: function(value) {
+    const parsedValue = this.parseFloat(value);
+
+    if (_isNaN(parsedValue)) {
+      return '';
+    }
+
+    const parts = _round(parsedValue, 2)
```
travist: This bit of code is crazy. I would feel so much better if this had some Unit tests within the ```utils.spec.js``` file to accompany how this is supposed to behave.

YuryRybakSoftensity: Done.

travist:
Looks good. Thanks!

PR #142, opened 2017-07-26
Wait for form to load before loading submission

douglaselee:
I'm using the Chrome browser on Windows 10.
My form contains a resource component followed by some text components.
Using code like this to render form with existing submission...

Formio.setBaseUrl('http://localhost:3001');
var form = new FormioForm(document.getElementById('formio'));
form.src = 'http://localhost:3001/form/59259f9614746b2a8819866c/submission/59774528acdba937887da5e7';

Code which populates resource component and code which sets submissions resource value are both trying to access instance of 'choices' at the same time and things go haywire.  Following component values do not get set with submission data.
Fix is to load submission after loading of form as in Formio.createForm.
Also fix select dropdown duplication caused by different instances of the same resource.
Also emit formio.error event when resource item loading fails for something like unauthorized.

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -188,6 +188,12 @@ export class SelectComponent extends BaseComponent {
         // Search for the choice.
         const choices = this.choices.store.getChoices();
         const foundChoice = choices.find((choice) => {
+          // For resources we may have two different instances of the same resource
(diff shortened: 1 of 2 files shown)


travist:
Looks good. Thanks!

douglaselee:
Happy to help Travis!

PR #143, opened 2017-07-26
Send request to delete when uploaded file is removed from list

douglaselee:
For file components using 'url' storage send delete requests when uploaded files are removed from the list.
ngFormio does this.


travist:
Thank you!

douglaselee:
Happy to help Travis!

PR #145, opened 2017-07-27
Add execution of custom JavaScript on button click

douglaselee:
Same thing was recently added to ngFormio.
May satisfy issue #126.


src/components/button/Button.js
```diff
@@ -71,6 +73,22 @@ export class ButtonComponent extends BaseComponent {
             event: event
           });
           break;
+        case 'custom':
+          var parent = this.parent;
+          while (parent.parent) {
+            parent = parent.parent;
+          }
+          var components = FormioUtils.flattenComponents(parent.components, true);
```
travist: This should probably say... 

```
var components = FormioUtils.flattenComponents(parent.component.components, true);
```

src/components/base/Base.js
```diff
@@ -32,6 +32,12 @@ export class BaseComponent {
     this.id = (component && component.id) ? component.id : Math.random().toString(36).substring(7);
 
     /**
+     * The API key of this component.  FormioUtils.eachComponent needs it to formulate the path.
+     * @type {string}
+     */
+    this.key = (component && component.key) ? component.key : '';
```
travist: See my comment below. If you change it to ```parent.component.components``` you probably won't need this.

src/components/Components.js
```diff
@@ -116,6 +116,7 @@ export class FormioComponents extends BaseComponent {
       _assign(this.options.components, FormioComponents.customComponents);
     }
     let comp = this.options.components.create(component, this.options, data);
+    comp.parent = this;
```
travist: This is handy.

douglaselee:
Thanks Travis.  Should be able to clean it up tomorrow.

douglaselee:
Hi Travis,

Changes are pushed.  Thanks again for the guidance.  This is much better in that now the components exposed to the script are the same as those as in the ngFormio renderer and not the HTML element components specific to the formio.js renderer.

With the ngFormio renderer a script could do something like _components.someComponent.label = 'Catfood';_ and, by virtue of Angular interpolation, the label change would be immediately reflected in the form.  With the formio.js renderer this is not happening.  Same applies to other component properties like prefix and suffix and other types of scripts like the ones for calculated data.  I know this isn't your problem but it may be a problem for me so I need to give it some thought.

travist:
What is the specific use case you are trying to accomplish with altering the labels with a button click? We may be able to figure out a better way to achieve this. What I am thinking is that we could expose a variable called "self" to the controller which is the component instance. We could then create a method to get the "root" component using your while loop with parent.

The form component already has a method called ```getComponent``` which will locate a component within a tree. https://github.com/formio/formio.js/blob/master/src/components/Components.js#L60

You could then combine that to be able to do something like this in your controller.

```
self.getRoot().getComponent('firstName').setLabel('First Name');
```

Obviously we would need to expose a new method called ```setLabel``` within the BaseComponent, but this is easy enough.

Hope this helps,

Travis.

douglaselee:
Hi Travis,

Thanks a lot for giving it so much thought so late on a Friday night!  That's way above and beyond the call of duty.

Don't have a single specific use case in mind but form designers will come up with many.  For example end user has entered Mary into spouseName component and designer would like spouseAge label to read 'What is Mary's age?'.  Or user has selected an EU country in country component and designer would like to prefix a currency component with a Euro symbol.

I think what you're suggesting would work but may be asking too much of non-technical form designers.

As this pull request branch stands now the script sees an object containing the flattened schema components as opposed to the HTML element components.  If there were a link from the schema component back to the HTML element component, call it _element_, the scripts could do something like _components.someComponent.element.setLabel('Catfood');_

On the other hand we could roll the branch back so that the scripts would see an object containing the flattened HTML element components.  (The schema components would still be available through the HTML element component's component property not that they'd be very useful to the scripts.)  Then the scripts could be simplified to do something like _components.someComponent.setLabel('Catfood');_

Even better the BaseComponent class could have a label property like:
    get label() {
        return this.component.label;
    }
    set label(value) {
        this.component.label = value;
        // Do whatever it takes to update inner HTML text of this component's label element
    }

Then scripts could do something thing like _components.someComponent.label = 'Catfood';_  just as they can today with the ngFormio render.  No script changes would be needed to accommodate one renderer or the other.  Of course the scripts would be seeing schema components when running under the ngFormio renderer and HTML element components when running under the formio.js renderer but they wouldn't need to know the difference or care about it.

What do you think about this?

douglaselee:
Strictly speaking, I wouldn't roll back the key duplication in Base.js but would improve FormioUtils.eachComponent to be able to operate on nested arrays of either type of component through _component.key_ or failing that _component.component.key_.

travist:
I agree with your approach with the previous comment where you use getters and settings to achieve the desired and compatibility between the core renderer and Angular 1.

douglaselee:
Thanks Travis.  In a frenzy putting Add Resource button into core renderer.  Basically got it working.  Just needs some styling, dialog-polyfill and more testing.  Will come back to this when that's wrapped up

douglaselee:
Hi Travis,

Latest commits should implement agreed upon changes (if I understand correctly).
Required renaming of BaseComponent _label_ to _labelElement_.
Get/set property rendering for other things like prefix, suffix... can follow in later pull requests.
Important one will be ability for scripts to change disabled state of components on the fly.
May already be possible, I just need to read the code some more.

travist:
Looks good! Thanks!

douglaselee:
Happy to help!  Also happy to see that _disabled_ property was implemented by get/set and that scripts which enable and disable components on the fly work unchanged in either the ngFormio or core renderer.

PR #146, opened 2017-07-28
Support custom column width, offset, push and pull

douglaselee:
Like ngFormio does.

### DIFF
--- a/src/components/columns/Column.js
+++ b/src/components/columns/Column.js
@@ -1,6 +1,11 @@
 import { FormioComponents } from '../Components';
 export class ColumnComponent extends FormioComponents {
   get className() {
-    return 'col col-sm-' + this.component.colWidth;
+      var comp   = this.component;
+      var width  = ' col-sm-'        + (comp.width  ? comp.width  : comp.colWidth);
+      var offset = ' col-sm-offset-' + (comp.offset ? comp.offset : 0);  
+      var push   = ' col-sm-push-'   + (comp.push   ? comp.push   : 0);  
+      var pull   = ' col-sm-pull-'   + (comp.pull   ? comp.pull   : 0);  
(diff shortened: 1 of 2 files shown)


travist:
We can probably get rid of this code now as well. This was attempting to calculate the width based on how many columns. It can go away.

https://github.com/formio/formio.js/blob/master/src/components/columns/Columns.js#L11

douglaselee:
Yes, I saw that but left it in for backward compatibility with forms created before they could be customized.  If you're sure then I can take it out.

travist:
It is an undocumented and hidden feature. It can be removed.

douglaselee:
OK, will do.  What do you recommend for the default width if none is specified for a given column?  I used 0 for the default offset, push and pull.

Speaking of undocumented and hidden features can I ask what is the purpose of the commits to ngFormio and ngFormBuilder like 'Updated renderer to support iframes.' related to iframes and PDFs?  Looks like it could evolve into something very useful.  Sorry, I know this is not the right place to ask but I'm not sure what the right place would be.

douglaselee:
Also, in what sense is the column customization hidden?  I see it at https://portal.form.io.

travist:
@douglaselee send me an email @ travis at form.io. I will setup a GotoMeeting with you and fill you in on all the PDF form stuff we are working on. It would also be great to personally connect with you to thank you for all the amazing work you have been doing in our open source repos.

douglaselee:
@travist thanks, will email you in a moment and include my cell phone number.  If it's easier you can just give me a call.  I don't have GotoMeeting installed yet and am not much to look at anyway.  Either way will be great to connect.

travist:
Looks good. Thanks!

douglaselee:
Happy to help!

PR #148, opened 2017-08-01
Add Resource button and dialog

douglaselee:
Like ngFormio renderer.
Tried actual 'X' close button in top right but it was too big and ugly so used \<b\> tag.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -14,6 +14,7 @@
   ],
   "dependencies": {
     "choices.js": "^2.8.7",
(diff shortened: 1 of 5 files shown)


douglaselee:
Now fully functional and can be merged if passes review.
Tested with Chrome and IE 11 which definitely needs the polyfill.
Tested with self-referencing recursive resources, the dialogs stack atop one another as expected.

travist:
Awesome! Looks good!

douglaselee:
Thanks Travis!

PR #158, opened 2017-08-24
Fix wizard nextPage looking for page instead of next like old version…

randallknutson:
…s do.


randallknutson:
This was a breaking change from ngFormio. Previously we were using "next" to specify the next page. If we are going to change this to "page" it should be backwards compatible.

PR #180, opened 2017-09-19
Use handler.type instead of undefined handler.event property

uhef:
In `destroy` we were accessing an undefined property `handler.event`. Correct property to access is `handler.type`.


travist:
Awesome thanks!

PR #182, opened 2017-09-19
Choices.js was sorting the select lists

silbermm:
We wanted the select lists to be sorted how we input them and not
alphabetically by default. To remedy this, we can pass `shouldSort:
false` to choices.js.


travist:
Works for me!  Thanks!

PR #188, opened 2017-09-25
Fix conditional performance on set submission.

randallknutson:


travist:
Should be fixed with 2.18.0. Please verify.

PR #200, opened 2017-10-02
Updates to eachComponent function

YuryRybakSoftensity:


YuryRybakSoftensity:
@travist Should I add unit-tests for editgrid (as well as datagrid and container)?

src/utils/index.js
```diff
@@ -95,20 +98,20 @@ const FormioUtils = {
       if (!noRecurse) {
         if (hasColumns) {
           component.columns.forEach(function(column) {
-            eachComponent(column.components, fn, includeAll, subPath(), parent ? component : null);
```
travist: Why did you remove ```parent ? component : null```. I am not sure what, but am certain this will break something out there.

YuryRybakSoftensity: You see, there is no assignment to *parent* in this function, that's why once this function is invoking with **parent === undefined** in condition **parent ? component : null** - **null** is always selected. Then recursively we always have **parent === null**.

I've spent some time investigating why this check was added and found that is was added in [this](https://github.com/formio/formio.js/commit/ab9c441cf395281c8817c06c7a796c78a10e8b05#diff-8968c62593320351ced6a0ef776ef599) commit. I understand the reason of [this](https://github.com/formio/formio.js/commit/ab9c441cf395281c8817c06c7a796c78a10e8b05#diff-b15c60e5637eec34588f8508175c2ff5L65) change, but I'm totally confused with [this](https://github.com/formio/formio.js/commit/ab9c441cf395281c8817c06c7a796c78a10e8b05#diff-b15c60e5637eec34588f8508175c2ff5L81) one and I think that it makes no sense. Please correct me if I'm wrong.

travist: This is not needed anymore due to the encryption finding the parent in a different (and more performant) way. Closing this for now unless absolutely necessary.

PR #214, opened 2017-10-11
force the select to open downwards

silbermm:
Some of our forms appear high on the page and if it is a small form with
select boxes, were opening upwards and off the page. In all cases, we
think opening downwards provides a much better experience.

I added the [documented option](https://github.com/jshjohnson/Choices#position) `position:bottom` to force always opening downwards.


travist:
We are turning this into a configuration. It will still default to "auto", but you can force it to be below or above by using the ```dropdown``` property on the JSON of the component.

```
{
  "type": "select",
  ...
  "dropdown": "below"
}
```

PR #216, opened 2017-10-12
Added file input to DOM as a hidden item to fix issue with IE

topisark:
At least some versions of IE were not displaying the file browser dialog. The problem was fixed by adding the file input element into the DOM as a hidden item (apparently display: none is not ok but opacity: 0 is fine). Also made the element positioned absolutely to not create empty space anywhere.


randallknutson:
What versions of IE does this fix?

topisark:
At least IE 11.

randallknutson:
👍

PR #217, opened 2017-10-12
Feature/i18n number

randallknutson:
Makes number and currency localizable


src/components/number/Number.js
```diff
@@ -1,52 +1,91 @@
...
+import _get from 'lodash/get';
+import createNumberMask from 'text-mask-all/addons/dist/createNumberMask';
 import { BaseComponent } from '../base/Base';
+
 export class NumberComponent extends BaseComponent {
   constructor(component, options, data) {
     super(component, options, data);
     this.validators = this.validators.concat(['min', 'max']);
+
+    this.decimalSeparator = (12345.6789).toLocaleString(i18next.language).match(/345(.*)67/)[1];
```
travist: This should be cached for performance. Think a form that may have 100 number fields. You can do this pretty easily by doing the following.

```
this.decimalSeparator = options.decimalSeparator = options.decimalSeparator || (12345.6789).toLocaleString(i18next.language).match(/345(.*)67/)[1];
```

travist: This would also allow you to force the decimalSeparator and thousandsSeparator with options.

PR #219, opened 2017-10-12
Respect tab index

silbermm:
Tab Indexes are being overwritten after creating a new Choice. This will fix that and allow for users to tab in order.


travist:
Awesome, and thank you for writing tests!

PR #226, opened 2017-10-17
Fix nested forms with save to another resource fails.

randallknutson:


src/components/form/Form.js
```diff
@@ -98,12 +98,14 @@ export class FormComponent extends FormioForm {
    * Submit the form before the whole form is triggered.
    */
   beforeSubmit() {
-    // Before we submit, we need to filter out the references.
-    this.data[this.component.key] = this.component.reference ? {_id: this._submission._id} : this._submission;
-
     // Ensure we submit the form.
     if (this.component.submit && !this.submitted) {
-      return this.submit(true);
+      let submission = this.submit(true);
```
travist: ```this.submit``` is asynchronous and returns a promise. You are setting the submission to a promise?

randallknutson: Turns out it wasn't async. It was returning the submission.

randallknutson: Didn't seem right but that's what this.submit(true) was returning.

travist: Something isn't right then... This returns ```beforeSubmit``` or ```executeSubmit```

https://github.com/formio/formio.js/blob/master/src/formio.form.js#L762

Those methods, return promises, so I am not sure how this returns a submission.

src/components/form/Form.js
```diff
@@ -98,12 +98,14 @@ export class FormComponent extends FormioForm {
...
+
+      // Before we submit, we need to filter out the references.
+      this.data[this.component.key] = this.component.reference ? {_id: submission._id, form: submission.form} : submission;
+
+      return submission;
```
travist: Oh... this may be causing it.

You are returning the submission here, but this should return a promise. This is getting propagated up to the parent form which is why this seems to work, but I am sure this is a bug.

src/components/form/Form.js
randallknutson: I'm not sure either but I got an error when trying to treat it as a promise. I inspected the result and it was a submission. Worked great when I did it this way.

src/components/form/Form.js
randallknutson: I'm taking a second look at this to see what is going on.

src/components/form/Form.js
randallknutson: Okay. Not sure why this was working the opposite yesterday but seems to be fixed now.

PR #240, opened 2017-11-02
Feature/FOR-784

YuryRybakSoftensity:


travist:
I did this in a different way. I created a new hook system. I also ported the beforeSubmit to use it as well.

PR #244, opened 2017-11-08
Feature/FOR-786: keyboard shortcuts for buttons, checkboxes, radio buttons

YuryRybakSoftensity:


src/formio.form.js
```diff
@@ -726,6 +818,7 @@ export class FormioForm extends FormioComponents {
    */
   submit(before) {
     if (!before) {
+      const { beforeSubmit } = this.options;
```
travist: Need to remove this line.

travist:
Before we pull this into the renderer, can you please verify that this does maintain reverse compatibility? Basically would like to know if this only is enabled with a new option, otherwise behaves the same?

PR #250, opened 2017-11-15
Feature/FOR-775: Ability to change placement of component label

YuryRybakSoftensity:


travist:
Exceptional work @YuryRybakSoftensity! Thank you for taking extra care to convert things into ES6 where you see them.

PR #251, opened 2017-11-15
Fixed customClass option on column component

Vercjames:


dist/formio.utils.js
```diff
@@ -760,76 +748,9 @@ http://ricostacruz.com/cheatsheets/umdjs.html
         }
       }
       return current; // Last
-
```
travist: Why was all of this removed?  You need to delete your ```node_modules``` folder, ```npm install``` and then rebuild and recommit.

PR #255, opened 2017-11-20
Add method and body parameters for URL option for Select Component

mark-james:
Take 2...
This change provides support for url methods other than Get for the selector dataSrc.
It has two new json attributes, method and body.
If not found method defaults to get.
If the method isn't post the body is set to null.

This is needed as GraphQL implementations are not restful and are often post only.

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -106,8 +106,15 @@ export class SelectComponent extends BaseComponent {
     }
   }
 
-  loadItems(url, search, headers, options) {
+  loadItems(url, search, headers, options, method, body) {
     options = options || {};
+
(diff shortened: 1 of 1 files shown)


travist:
Looks good to me. Thanks!

mark-james:
No problem. Thanks for merging. 

If this could make it's way into an angular-formio release fairly soon, that would be much appreciated!

travist:
Once I update this library, there is no need to update angular-formio since this library is a dependency. Just reinstall ```angular-formio``` and it should get the latest version.

mark-james:
Great, thanks. I didn't realise.

PR #260, opened 2017-11-22
Feature interpolate select url headers

mark-james:
For the selector component I found the need to make the header value dynamic based on environment and/or user. The common use case is auth tokens.

So the json may look something like this -
"value": "{{ data.serviceToken }}",
"key": "Authorization"


mark-james:
Hi @travist ,
I've discovered that interpolation doesn't happen when the select component is part of a datagrid. The datagrid only has access to the data from the grid when the interpolation function is run. This impacts both the existing url and also the headers code in this change.

I was wondering if you knew of the top of your head the likely issue and could point me in the right direction. If practical I can fix it as part of this pull request.
cheers
Mark

travist:
Looks good. Thanks!

mark-james:
Thanks Travis. Do you have any feedback regarding the datagrid issue in the comment above? Any pointer to reduce the time it takes me to find the issue would be handy.

PR #263, opened 2017-11-22
Adding Optional "BreadcrumbSettings" to options

danrye:
Support for making the pagination non-clickable

Adds new class to disable clicking, and doesn't add the even listener.


danrye:
Bug, don't merge, update coming

travist:
Looks good. Thanks @danrye!

PR #269, opened 2017-11-27
Feature/wizard button settings

danrye:
Adding the ability to hide next/previous/cancel buttons on the wizard pages via an optional option of "buttonSettings" and setting next/previous/cancel to false.

### DIFF
--- a/src/formio.wizard.js
+++ b/src/formio.wizard.js
@@ -246,10 +246,9 @@ export class FormioWizard extends FormioForm {
     }
 
     // Check for and initlize breadcrumb settings object
-    this.options.breadcrumbSettings = this.options.breadcrumbSettings || {};
-    if(this.options.breadcrumbSettings.clickable === undefined) {
-      this.options.breadcrumbSettings.clickable = true;
-    }
+    this.options.breadcrumbSettings = defaults(this.options.breadcrumbSettings, {
+      clickable: true
+    });
 
     this.wizardHeader = this.ce('ul', {
       class: 'pagination'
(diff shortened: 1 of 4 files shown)


danrye:
I'd love to document these options somewhere, but I couldn't find where options are documented.  Would /formio.js/app/examples/wizard.md be the right place?

travist:
Hello @danrye.

Thanks for your contribution. There are a few places to do documentation.

 - Wiki - https://github.com/formio/formio.js/wiki
 - Site - by adding examples @ https://github.com/formio/formio.js/tree/master/app
 - Using code with [ESDoc](https://esdoc.org/). Run ```npm run build``` to generate the SDK docs from the code.

Thanks!

Travis.

src/formio.wizard.js
```diff
@@ -206,12 +206,27 @@ export class FormioWizard extends FormioForm {
   }
 
   hasButton(name, nextPage) {
+    // Check for and initlize button settings object
+    this.options.buttonSettings = this.options.buttonSettings || {};
```
travist: You could use [_.defaults](https://lodash.com/docs/4.17.4#defaults), something like this.

```
const _defaults = require('lodash/defaults');
...
...

this.options.buttonSettings = _defaults(this.options.buttonSettings, {
  showPrevious: true,
  showNext: true,
  showCancel: true
});
```

danrye:
@travist I've added documentation on the Site and via ESDoc, if/after this is merged then I will update the wiki as well

travist:
Looks good. Thanks!  Will include this in a new build before the weekend.

PR #271, opened 2017-11-28
Adding trigger on row remove

cabrerabywaters:
If applied this PR will trigger an event every time a row is deleted from the Datagrid array.


travist:
Looks good, thanks! We will include this in a new build before the weekend.

sssxiaocainiao:
What is the trigger method

PR #286, opened 2017-12-14
Add StripeCheckout component

GuillaumeSmaha:
Hi,

This a new component to integrate a payment button using Stripe.
For context, Stripe will get the card and payment information and will return a token which will use on backend to made the payment.
So, the component is based on ButtonComponent and on click event, it will show the Stripe popup. After validation of information, Stripe send a request on their server and in response, the obtained token will be saved in the form data.

The component also manage the case where the button can submit the form after saving the token in the data. (action="submit" for this case)

### DIFF
--- a/app/examples/stripeCheckout.md
+++ b/app/examples/stripeCheckout.md
@@ -1,0 +1,121 @@
+---
+title: Stripe Checkout
+layout: vtabs
+section: examples
+weight: 150
+---
+
+### Stripe Checkout
+Stripe securely collect card information from your customers and create a card payment.
+This component is based on the ButtonComponent and add an overlay to call [Stripe Checkout](https://stripe.com/docs/checkout) library.
+After the customer validate his payment information, a token will be returned by Stripe API and it will be stored in the form data.
+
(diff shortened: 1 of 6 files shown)


travist:
This is awesome! I am wondering if we should start a ```contrib``` folder for all of the contributed components and this would be a good candidate for that. Others that I can think of are things like ReCaptcha, etc. You could then include it using the following.

```js
import { Formio } from 'formiojs/full';
import { StripeCheckoutComponent } from 'formiojs/contrib/stripe/StripeCheckout';
Formio.registerComponent('stripeCheckout', StripeCheckoutComponent);
```

Let me know what you think. This would probably need to go into a ```src/contrib``` folder, and then we would have a build routine run that through Babel to create the root ```contrib``` folder.

GuillaumeSmaha:
Yes, good idea. It will avoid to overload the component directory with basic component.
Gmap component could also be moved to this directory (or not for compatibility)

GuillaumeSmaha:
@travist I moved the component in `contrib` directory

travist:
Thank you for making this change. I am happy to include it now that it is in contrib folder.

PR #287, opened 2017-12-15
Feature/viewAsHtml initial release

YuryRybakSoftensity:


src/components/address/Address.js
```diff
@@ -450,4 +451,9 @@ export class AddressComponent extends TextFieldComponent {
     info.attr.class += ' address-search';
     return info;
   }
+
+  get viewOnlyValue() {
```
travist: I think this should just be called ```view``` since that means the "view" html of a component.

```
get view() {
  ...
}
```

PR #289, opened 2017-12-19
Add contrib directory in build process

GuillaumeSmaha:
Following discussion in PR https://github.com/formio/formio.js/pull/286


travist:
Looks good. I will pull this into the next major release. Going to get out a patch release next.

Issue #288, opened 2017-12-19
Columns are not visible when adding row in Datagrid


Aiso786:
Hello guys, i have the following issue on the DataGrid  component.

I have 2 columns but when i click on the "Add row" button, the columns are not visible.

![image](https://user-images.githubusercontent.com/22521614/34146259-93b7c91a-e499-11e7-8ba1-7128b10ff2af.png)

I think the problem is here:

![image](https://user-images.githubusercontent.com/22521614/34146334-dcf8e69a-e499-11e7-8282-27d805c7a1d5.png)

I think we should have this instead of the highlighted lines:
`showColumn |= col.checkConditions(data);`

What do you think guys ?

## Update
We can't do this:
`showColumn |= col.checkConditions(data);`
Because col doesn't have the 'checkConditions' function.

## Update 2
Need help guys :(

Aiso786:
I made a PR:

#290

Aiso786:
PR has been merged to master, i close the issue.

PR #290, opened 2017-12-19
Fixing the following issue: columns are not visible when clicking on…

Aiso786:
… "Add" button on an empty DataGrid

Resolves the following issue:
https://github.com/formio/formio.js/issues/288

Sorry am new to Github


travist:
Looks good. Thanks!

PR #292, opened 2017-12-21
Conditional enable disable

tank4u:
@travist : We need functionality to enable/disable controls in place of hide/show on conditional check. I have tried my way (might not be best) to fix it in the pull request changes.

Please review and let me know if this works or there is better way to handle it?

Thanks in advance.

### DIFF
--- a/src/components/base/Base.js
+++ b/src/components/base/Base.js
@@ -10,6 +10,7 @@ import _clone from 'lodash/clone';
 import _defaults from 'lodash/defaults';
 import _isEqual from 'lodash/isEqual';
 import _isUndefined from 'lodash/isUndefined';
+import _has from 'lodash/has';
 import i18next from 'i18next';
 import FormioUtils from '../../utils';
 import { Validator } from '../Validator';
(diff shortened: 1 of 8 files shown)


travist:
My goal is to be able to implement this in a way so that it could apply to any property. We should also implement this in a new property called ```conditions``` so that we could maintain reverse compatibility, and then eventually deprecate the ```conditional``` property. I think this should be an array, where you could provide several conditions. For example, lets say you wish to have a conditionally disabled field, but also a conditionally required, you could do the following.

```
{
  "conditions": [
    {
      "property": "disable",
      "condition": {
        "eval": "...."
      }
    },
    {
      "property": "required",
      "condition": {
        "json": {...}
      }
    }
  ]
}
```

Before I commit this, I would rather put forth the effort to implement this capability.

Your thoughts?

tank4u:
I like the idea of handling more than just enable/disable, but there will some properties which might be mutually exclusive like "if field is Disabled then making it Required will stop form to be submitted, as user can't change the value".

We need to lay down some rules for the list of properties supported and order of them. Also the condition configuration to be supported when-eq, json, eval, custom?

I took short-cut :) because the time crunch in the current project, so I would say I will revisit the solution after around 2 weeks if that is fine.

travist:
I agree with you. And not a problem on timeframe. This is on our roadmap, but if you tackle it before then we will be grateful. 

No rush.

tank4u:
so any plans of merging this as committed or should I use my custom formio-js version in my application? 

Also had question on how to clear the value on disable, if configured? I tried with 'setValue to null' but that didn't work, what am I missing?

travist:
We will work on getting this committed, but my only hesitation is that it introduces a new property on the JSON schema. We will work out what the new format is going to be and will probably pull this in once we begin that effort. In the meantime, you will probably need to work off of your own fork.

tank4u:
Sounds good... thanks @travist for your cooperation.

travist:
This is now going to be handled with our new "Field Logic" feature.

https://github.com/formio/formio.js/pull/339

PR #293, opened 2017-12-22
Feature/FOR-787: Date formatting based on the locale configuration

YuryRybakSoftensity:


src/utils/index.js
```diff
@@ -527,6 +527,17 @@ const FormioUtils = {
   },
   isValidDate(date) {
     return _isDate(date) && !_isNaN(date.getDate());
+  },
+  getLocaleDateFormatInfo(locale) {
+    const formatInfo = {};
+
+    const day = 21;
```
travist: What is the significance of the magic date of 11/21/2017. This seems like something that should not be hard coded into the renderer.

randallknutson: I believe this is similar to what we're doing in currency were we have to render a random value to check the correct format. This is just the date he is using the check the format.

YuryRybakSoftensity: That just sample date to understand, that month is the first in locale string.

PR #298, opened 2018-01-02
FOR-928: Radio Component -> Implemented adding 'radio-selected' CSS class to selected option

Alena-Levina-Softensity:


src/components/radio/Radio.js
```diff
@@ -165,6 +170,40 @@ export class RadioComponent extends BaseComponent {
...
+      const value = this.data[this.component.key];
+      const optionSelectedClass = 'radio-selected';
+
+      const inputs = this.inputs;
+      _each(inputs, (input) => {
+        const wrapperClass = this.optionWrapperClass;
+        let wrapper;
+        //find wrapper of this option
+        let ancestor = input.parentElement;
+        while (ancestor) {
```
travist: This seems pretty non-performant since it seems to be traversing the DOM tree every time someone selects a new radio. Could we alter this to instead keep track of these wrappers at the time they are created and then instead of iterating through the inputs, we would just iterate through the containers? That would speed up this code and also make it much cleaner. We also have some methods that may help in the BaseComponent called ```addClass``` and ```removeClass```. You should look into using those.

Alena-Levina-Softensity: Thanks for the note. I've added wrappers to this in createInput and iterated over them on updateValue and also used BaseComponent's addClass and removeClass methods

PR #308, opened 2018-01-09
Adding row property to checkCalculated

edwinanciani:


travist:
I included this in https://github.com/formio/formio.js/pull/309

PR #311, opened 2018-01-09
Bugfix/FOR-933: Search in Select Resource

YuryRybakSoftensity:


travist:
Please keep an eye on your pull request for the choices library. Once that gets pulled in, we will need to point this back to the main repo.

PR #312, opened 2018-01-09
Add stripe component: card input field/Pay button

GuillaumeSmaha:
Hi,

This is a another stripe component. This component provides a card input field and optionally a pay button (Apple Pay, Android Pay, Payment request API). Like the other component, Stripe will return a token.

There are 3 cases:
- With a card input field, the validation to stripe will be done before submission with `beforeSubmit()` method
- With a card input field and a button to Authorize the payment, the call to Stripe will be done when the user clicks on the button
- With the pay button, the authorization will be done when the user validates the popup for payment

@travist I created a parameter `componentsMapping` to allow injection of the value from other fields into the request to Stripe for card validation. But I saw your nice example on http://formio.github.io/formio.js/app/examples/stripeCheckout.html where you inject values from other component.
So, should I change `componentsMapping` to take a value as input instead of the key of another component ?

I made an example here with different cases:
https://htmlpreview.github.io/?https://raw.githubusercontent.com/GuillaumeSmaha/formio.js/feature/add_stripe_component_build/app/examples/stripe.html

cc: @stephenott

### DIFF
--- a/app/examples/stripe.md
+++ b/app/examples/stripe.md
@@ -1,0 +1,203 @@
+---
+title: Stripe
+layout: vtabs
+section: examples
+weight: 150
+contrib: true
+---
+
+### Stripe
+Stripe securely collect card information from your customers and create a card payment.
+This component provides a card input field created by [Stripe](https://stripe.com/docs/stripe-js) library.
+The component also provides a "pay button" using Android Pay, Apple Pay or Payment Request API.
+
+The authorization will be automaticly called when users use the pay button. Instead, for the card input field, the call will be done before submission
+or you can call manually add an button which call the method `authorize()`.
+
+These parameters are directly injected to Stripe functions:
+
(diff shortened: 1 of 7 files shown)


travist:
The way I prefer this to work is to use interpolation to pass in the values you want into the ```componentsMapping```. I see in your example, that you do something like this..

```
"componentsMapping": {
  "name": "firstName"
}
```

This is not very flexible. Example, how would I pass both firstName and lastName as the ```name``` parameter into Stripe? Using this, I could not. However, if I change it to interpolation, I can do something like this.

```
"componentsMapping": {
  "name": "{{ data.firstName }} {{ data.lastName }}"
}
```

I already made this change to your ```StripeCheckoutComponent``` here.  https://github.com/formio/formio.js/blob/master/src/contrib/stripe/checkout/StripeCheckout.js#L68

Using the translate method ```t()```, this will handle interpolation for you.

GuillaumeSmaha:
I followed your advice and updated componentsMapping (renamed to cardData like Stripe parameter name)
I also updated the doc

PR #314, opened 2018-01-09
Datagrid's add another position logic

Vercjames:
The render accommodations for https://github.com/formio/ngFormBuilder/pull/371 allowing for a more versatile placement of the "+ Add Another" button in the Datagrid component.

### DIFF
--- a/src/components/base/Base.js
+++ b/src/components/base/Base.js
@@ -668,9 +668,32 @@ export class BaseComponent {
     let addIcon = this.ce('span', {
       class: 'glyphicon glyphicon-plus'
     });
+
     addButton.appendChild(addIcon);
     addButton.appendChild(this.text(this.component.addAnother || ' Add Another'));
+
     return addButton;
+  }
+
+  /**
+   * Adds a new icon button to  new rows for multiple input elements.
+   * @returns {HTMLElement} - The "+" button html element.
+   */
+  addIconButton(){
+    let addIconButton = this.ce('a', {
+      class: 'btn btn-primary'
+    });
+    this.addEventListener(addIconButton, 'click', (event) => {
+      event.preventDefault();
+      this.addValue();
+    });
+
+    let addIcon = this.ce('span', {
+      class: 'glyphicon glyphicon-plus'
+    });
+
+    addIconButton.appendChild(addIcon);
+    return addIconButton;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/components/datagrid/DataGrid.js
```diff
@@ -121,14 +127,31 @@ export class DataGridComponent extends FormioComponents {
...
+      )){
+        let tr = this.ce('tr');
+        let td = this.ce('td', {
+          colspan: (this.component.components.length + 1)
+        });
+        td.appendChild(this.addButton());
+        tr.appendChild(td);
+        this.tbody.appendChild(tr);
+
+      // if(
```
travist: Do not comment out code. If you don't need it, then just remove it. GIT helps us keep track of things removed and added.

src/components/datagrid/DataGrid.js
```diff
@@ -60,6 +60,12 @@ export class DataGridComponent extends FormioComponents {
     if (!this.shouldDisable) {
       let th = this.ce('th');
       tr.appendChild(th);
+
+      // add the "+" - "add another" button
+      if (this.component.addAnotherPosition === "top" || this.component.addAnotherPosition === "both") {
```
travist: You still need a check here for ```!this.shouldDisable```

src/components/base/Base.js
```diff
@@ -668,11 +668,33 @@ export class BaseComponent {
...
     addButton.appendChild(addIcon);
     addButton.appendChild(this.text(this.component.addAnother || ' Add Another'));
     return addButton;
   }
 
+  /**
+   * Adds a new icon button to  new rows for multiple input elements.
+   * @returns {HTMLElement} - The "+" button html element.
+   */
+  addIconButton(){
```
travist: This is VERY similar to the ```addButton``` method. Why did you create a new method here. You could very easily pass in a parameter to the ```addButton``` method to not add the text piece of it if that variable is set, and it would make the code much cleaner.

src/components/datagrid/DataGrid.js
Vercjames: Why do I still need to check !this.shouldDisable, the if statement is embedded inside the  check on line 60?

travist: didn't see that, so all good.

src/components/base/Base.js
```diff
@@ -668,9 +668,15 @@ export class BaseComponent {
...
       class: 'glyphicon glyphicon-plus'
     });
-    addButton.appendChild(addIcon);
-    addButton.appendChild(this.text(this.component.addAnother || ' Add Another'));
-    return addButton;
+
+    if(justIcon){
+      addButton.appendChild(addIcon);
+      return addButton;
+    }else{
```
travist: Please fix spacing in both if statement and else statement. Our standard is as follows.

```
if (justIcon) {

} else {

}
```

PR #322, opened 2018-01-16
Bootstrap 4 support

travist:


randallknutson:
Is it possible to add some tests to make sure the classes are added properly? Other than that this looks good.

PR #339, opened 2018-01-25
Field Logic

randallknutson:


randallknutson:
Example config form can be found at https://jxijnknquwhvpuv.form.io/advancedconditions

src/components/button/Button.js
```diff
@@ -89,8 +89,6 @@ export class ButtonComponent extends BaseComponent {
           });
           break;
         case 'custom':
-          // Get the FormioForm at the root of this component's tree
-          var form = this.getRoot();
           // Get the form's flattened schema components
           var flattened = FormioUtils.flattenComponents(form.component.components, true);
```
travist: Won't ```form``` be undefined now?

randallknutson: No. It is passed in to new Function below.

travist: Right, but in this line it calls.

```
var flattened = FormioUtils.flattenComponents(form.component.components, true);
```

which refers to a variable called ```form```. Won't that now be undefined since you removed the variable?

randallknutson: Oh, yup. That's a problem

PR #340, opened 2018-01-26
WYSIWYG Update - placeholder, backspace, etc.

Vercjames:
Added placeholder and addressed the error that was causing core WYSIWYG component to fail preventing several items form working.

[Resolved] FOR-976 Placeholders do not work with wysiwyg textarea.
[Resolved] FOR-977 The tab key into the text area captures the buttons of the editor.
[Resolved] FOR-998 WYSIWYG submitted while it is still in “focus” then the data isn’t saved.
[Resolved] FOR-1006 Core Renderer - Unable to backspace using the WYSIWYG component.

### DIFF
--- a/src/components/Validator.js
+++ b/src/components/Validator.js
@@ -71,7 +71,6 @@ export const Validator = {
         if (!FormioUtils.boolValue(setting)) {
           return true;
         }
-
         return !component.isEmpty(value);
       }
     },
--- a/src/components/Validator.js
+++ b/src/components/Validator.js
@@ -72,14 +72,6 @@ export const Validator = {
           return true;
         }
 
-        // WYSIWYG validation
-        if (component.quill) {
-          if (value === "<p><br></p>") {
-            return false
-          }
-          return true
-        }
-
         return !component.isEmpty(value);
       }
     },
--- a/src/components/Validator.js
+++ b/src/components/Validator.js
@@ -71,6 +71,15 @@ export const Validator = {
         if (!FormioUtils.boolValue(setting)) {
           return true;
         }
+
+        // WYSIWYG validation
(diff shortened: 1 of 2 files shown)


src/components/Validator.js
```diff
@@ -71,6 +71,15 @@ export const Validator = {
         if (!FormioUtils.boolValue(setting)) {
           return true;
         }
+
+        // WYSIWYG validation
+        if (component.quill) {
```
travist: This should not be here. You should implement this in the ```isEmpty``` method inside of the ```TextAreaComponent``` class. This will keep out component specific logic out of the validator.

src/components/textarea/TextArea.js
```diff
@@ -99,8 +107,19 @@ export class TextAreaComponent extends TextFieldComponent {
       return super.getValue();
     }
 
-    if (this.quill) {
-      return this.quill.root.innerHTML;
+    // FOR-998 Assigns default value
+    if (!this.data[this.component.key]) {
+      this.data[this.component.key] = "<p></p>";
```
travist: We want this for any non-wysiwyg TextArea's?.... I think this would set the default value of ```<p></p>``` inside of the normal TextArea won't it?

travist:
I made quite a bit of changes. Please verify that it still works to correct the issues you set out to fix.

travist:
Did some tests on my end and it seems good to me. Let me know otherwise.

Vercjames:
I pulled in your the changes to make sure the fixes work in my testing environments. There does seem to be one non-blocking on this item. The "[Key] is not required" message doesn't trigger. 
 
https://www.screencast.com/t/wIQwbq3DG

PR #342, opened 2018-01-26
Fix FOR-400 for false conditionals and null values

douglaselee:
Create a component with simple conditional where,
This component should Display: is false,
When the form component: is some other Text component with no default value,
Has the value: is any non-empty string.

View form and refresh browser.
Component will not display even though Text component referred to is empty (null).

FOR-400 returns false for null values regardless of setting of `component.conditional.show`.
Better to convert null value to empty string so normal logic can precede which takes into account setting of `component.conditional.show`.


travist:
Looks good. Thanks @douglaselee!

douglaselee:
Happy to help @travist!

PR #344, opened 2018-01-28
Update Select.js

socotra69:
change the generated url in order to avoid two '?' in the query.
That can happened if you tried to show a form with a select that filters resources


travist:
Thank you!

PR #348, opened 2018-02-01
Custom var_search operator using dotty for accessing array of objects…

Renaud009:
This feature should be available in jsonLogic core in fact.

This feature in formio would especially be usefull when using the datagrid component that store data in an array of object.
If you want to calculate over some properties of this array without having to re-create a new Array of integers, this solution is the best way to archieve this.

The new operator will permit accessing deep json objects, which explains the use of dotty 'search' which will respond to more various use cases.
See dotty 'search' documentation: https://github.com/deoxxa/dotty

Example: if you want to calculate final total of every datagrid row totals (datagrid key: 'items')
```json
{
...
  calculateValue: {
    _sum: {
      var_search: "data.items.*.total"
    }
  }
}
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "formiojs",
-  "version": "2.24.6",
+  "version": "2.25.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


travist:
I believe that you can just use [map/reduce](http://jsonlogic.com/operations.html#map-reduce-and-filter) to do this.

```
calculateValue: {
  "reduce": [
    {
      "map": [
        {"var":"data.items"},
        {"var":"total"}
      ]
    },
    {
      "+": [
        {"var":"current"}, 
        {"var":"accumulator"}
      ]
    },
    0
  ]
}
```

PR #355, opened 2018-02-07
Event listener no longer fires unintentionally.

Vercjames:
The Event listener inside the Quill object could fail if the WYSIWYG configuration  were changed to exclude certain modules. This would prevent WYSIWYG data from updating field values correctly. 

FOR-990
Video Before & After fix (For Testing)
https://www.screencast.com/t/PBEAfe3Vn41P


travist:
Well done.

PR #357, opened 2018-02-08
Bugfix/FOR-969: 'Delimiter' property for Number component.

YuryRybakSoftensity:


src/components/number/Number.js
```diff
@@ -70,17 +83,18 @@ export class NumberComponent extends BaseComponent {
       mask: createNumberMask({
         prefix: '',
         suffix: '',
-        thousandsSeparatorSymbol: _get(this.component, 'thousandsSeparator', this.thousandsSeparator),
-        decimalSymbol: _get(this.component, 'decimalSymbol', this.decimalSeparator),
-        decimalLimit: _get(this.component, 'decimalLimit', this.decimalLimit),
-        allowNegative: _get(this.component, 'allowNegative', true),
-        allowDecimal: _get(this.component, 'allowDecimal', !(this.component.validate && this.component.validate.integer))
+        thousandsSeparatorSymbol: _.get(this.component, 'thousandsSeparator', this.delimiter),
```
YuryRybakSoftensity: @travist Should I also deprecate 'thousandsSeparator' property for component schema here?

travist: @YuryRybak just leave it for now.

PR #360, opened 2018-02-08
FOR-1035 Fix clear on hide

randallknutson:


src/components/checkbox/Checkbox.js
```diff
@@ -173,6 +173,7 @@ export class CheckBoxComponent extends BaseComponent {
     if (!this.input) {
       return;
     }
+    this.data[this.component.key] = value;
```
travist: Why did you need to add this? This is done within ```updateValue``` which is called at the bottom of this method. That is also used to determine if the value changed so it can trigger an update. I am pretty sure this breaks the "changed" functionality where the change event is no longer fired on checkboxes after doing this.

src/components/container/Container.js
```diff
@@ -34,13 +39,18 @@ export class ContainerComponent extends FormioComponents {
     if (!value || !_.isObject(value)) {
       return;
     }
+    this.data[this.component.key] = value;
```
travist: This will probably break the "changed" trigger from occurring.

src/components/checkbox/Checkbox.js
randallknutson: I can take another look at this but what happens is when the data is cleared on "clear on hide" it no longer exists in this.data. This was causing lots of issues where the component was attempting to access a property that no longer existed. I'll try taking this out in several places and see if it still passes tests.

src/components/container/Container.js
randallknutson: This one is necessary since Components.updateValue never sets the original data object for the component. Sub fields will still trigger changed.

src/components/survey/Survey.js
```diff
@@ -63,6 +63,7 @@ export class SurveyComponent extends BaseComponent {
 
   setValue(value, flags) {
     flags = this.getFlags.apply(this, arguments);
+    this.data[this.component.key] = value;
```
travist: Shouldn't this be within the ```if (!value)``` statement? I suppose you needed to add this in case it is a "reset" of value and it does not get into ```updateValue```?

src/components/selectboxes/SelectBoxes.js
```diff
@@ -54,6 +54,7 @@ export class SelectBoxesComponent extends RadioComponent {
   setValue(value, flags) {
     value = value || {};
     flags = this.getFlags.apply(this, arguments);
+    this.data[this.component.key] = value;
```
travist: Why doesn't ```updateValue``` take care of this? I am worried that this will break the "changed" triggers from working.

src/components/textarea/TextArea.js
```diff
@@ -118,6 +118,7 @@ export class TextAreaComponent extends TextFieldComponent {
       return super.setValue(value, flags);
     }
 
+    this.data[this.component.key] = value;
```
travist: ```updateValue``` should handle this, does it not?

PR #364, opened 2018-02-13
Bugfix/FOR-1042: Components asString.

YuryRybakSoftensity:


src/components/base/Base.js
```diff
@@ -395,28 +395,29 @@ export class BaseComponent {
...
+      return value.join(', ');
+    }
 
-  updateViewOnlyValue() {
-    this.empty(this.valueElement);
-    this.setupValueElement(this.valueElement);
+    return value.toString();
   }
 
-  empty(element) {
```
travist: Please don't remove this. It is a helpful helper method that I am using in the form builder branch.

src/components/base/Base.js
```diff
@@ -1832,8 +1833,7 @@ export class BaseComponent {
    * Prints out the value of this component as a string value.
    */
   asString(value) {
-    value = value || this.getValue();
```
travist: We probably need two versions of getting a view of a component. 1.) to return an HTML view of the component and 2.) to get a "string" only representation of a component. With this change, you basically combine 1 and 2 together, but I don't think that is right. There are cases where I may want one vs. the other (like when saving the value of the component to a database).

YuryRybakSoftensity: So I need simply return implementation of `asString` method and use `getView` instead of it in angular-formio?

PR #372, opened 2018-02-19
Adjust uploadFile for new layout of providers.storage

douglaselee:
Adjust downloadFile for new layout of providers.storage too.


douglaselee:
Happy to help @travist!

PR #376, opened 2018-02-23
Fix validations automatically triggering if you have only a container…

randallknutson:
… on a form at the root level.


randallknutson:
Wasn't able to fix the datagrid as well right now as it caused side effects and didn't seem to have the same problem.

PR #377, opened 2018-02-27
fix(FOR-1076): On the Data Grid component, Min and Max length

edwinanciani:
Validation is not rendering.

Datagrid.js changed.

Added the same validations from ngFormio.
Max length example.
https://monosnap.com/file/edVfLwWjHXtdXBJ8E5emLTbH4IIgBF

Min length example
https://monosnap.com/file/u0OrooMCnb7hcrJ5VqjVOL8EHXk8PF


src/components/datagrid/DataGrid.js
```diff
@@ -86,10 +86,20 @@ export class DataGridComponent extends FormioComponents {
...
   buildRows(data) {
     this.tbody.innerHTML = '';
     this.rows = [];
     this.components = [];
+
+    // Check minLength validation
+    if (this.component.validate && this.component.validate.hasOwnProperty('minLength') && this.data[this.component.key].length < this.component.validate.minLength) {
+      let toAdd = this.component.validate.minLength - this.data[this.component.key].length;
+      for (let i = 0; i < toAdd; i++) {
+        this.data[this.component.key].push({});
```
travist: Why are you doing this? This code looks really suspicious.

edwinanciani: I did this because we need to have a minLength property. So basically what it those is if you have minLength set will create those fields for you. It's what ngFormio does.

travist:
Please resolve the merge conflict and I will pull this in.

PR #381, opened 2018-03-06
Bugfix/object object

Vercjames:
Select and Resource Components when loaded in via a submission grid have there values nested. As a result the current logic evaluated the "value" as object object. The code submitted below provides logic to check to see if data is an object and extract the nested data and return a string. 

Broken View
https://www.screencast.com/t/12WynaGzA6V

Fixed View
https://www.screencast.com/t/z1DhMNv3xph

### DIFF
--- a/src/components/base/Base.js
+++ b/src/components/base/Base.js
@@ -405,7 +405,6 @@ export class BaseComponent {
   }
 
   getView(value) {
-
     if (!value) {
       return '';
     }
--- a/src/components/base/Base.js
+++ b/src/components/base/Base.js
@@ -405,11 +405,23 @@ export class BaseComponent {
   }
 
   getView(value) {
+
     if (!value) {
       return '';
     }
     if (Array.isArray(value)) {
       return value.join(', ');
+    }
+
+    // FOR-1094
+    if (typeof value === 'object') {
+      let nestedObject = [];
+
+      for (let propt in value.data) {
+        nestedObject.push(value.data[propt]);
+      }
+
+      return nestedObject.join(' ');
     }
 
     return value.toString();
--- a/src/components/base/Base.js
+++ b/src/components/base/Base.js
@@ -415,13 +415,13 @@ export class BaseComponent {
 
     // FOR-1094
     if (typeof value === 'object') {
-      let nestedObject = [];
+      let convertObj = [];
 
       for (let propt in value.data) {
-        nestedObject.push(value.data[propt]);
+        convertObj.push(value.data[propt]);
       }
 
-      return nestedObject.join(' ');
+      return convertObj.join(' ');
     }
 
     return value.toString();


travist:
We should be using the ```component.template``` to render these items and then use interpolation which would then show the element as it is shown within the select dropdown. Also, this logic should be done in the Select component and not Base since that component is the only one that has the ```component.template```. Resource derives from Select, so it would be covered there as well.

src/components/select/Select.js
```diff
@@ -498,6 +498,22 @@ export class SelectComponent extends BaseComponent {
...
+    if (data && !this.useTemplate) {
+      return this.t(data.label || data);
+    }
+    if (typeof data === 'string') {
+      return this.t(data);
+    }
+    if (this.component.template) {
+      return this.interpolate(this.component.template, {item: data})
+    }
+    return data.label || this.component.placeholder;
```
travist: I cannot think of a good use case where you would want to return the "label" of the component as its "data" view. If the data is empty, then just return empty string. Actually, why can you not just do this?...

```js
getView(data) {
  return this.itemTemplate(data);
}
```

Vercjames: I tried that but the "replace" method isn't working

src/components/select/Select.js
```diff
@@ -97,6 +97,10 @@ export class SelectComponent extends BaseComponent {
     if (typeof data === 'string') {
       return this.t(data);
     }
+    if (!data.data) {
+      return '' || this.component.placeholder
```
travist: Isn't this the same thing as just saying... ```return this.component.placeholder```?

And on that note, why in the world would we ever return the component placeholder as the "value" of the component. If the value is empty, we should see that it is empty in all displays. Having it instead show ```Select your Country``` as the value would be pretty strange.

travist: Also, what is ```data.data``` here?  Looking at the code below this, I am not sure how this is protecting anything since ```data.data``` is not referenced anywhere here. Maybe the code should be

```
getView(data) {
  return this.itemTemplate({data});
}
```
Not sure, but this is something that a good breakpoint would easily show you what is going on.

src/components/select/Select.js
```diff
@@ -97,9 +97,12 @@ export class SelectComponent extends BaseComponent {
...
       return this.t(data);
     }
+
     const template = this.component.template ? this.interpolate(this.component.template, {item: data}) : data.label;
-    const label = template.replace(/<\/?[^>]+(>|$)/g, '');
-    return template.replace(label, this.t(label));
+    if (template) {
+      const label = template.replace(/<\/?[^>]+(>|$)/g, '');
+      return template.replace(label, this.t(label));
+    }
```
travist: So, does this mean that if there isn't any template it will return ```undefined```. This would probably show ```undefined``` in the datagrid. Is this expected?

Vercjames: Datagrid is blank and no value is recorded in the <td></td>
https://www.screencast.com/t/bFgmNdRv

travist: Ok, but what about Select components that do not define templates (like manually configured), but also have data as an object. This will return empty which isn't right. At that point, I would prefer to do something like this just so that they get a representation of what data is stored.

```
else {
  return JSON.stringify(data);
}
```

travist: Also, I don't see anywhere that is checking to see if the data is ```null``` to return an empty string.  It needs this at the beginning.

```
if (!data) {
  return '';
}
```

PR #383, opened 2018-03-08
Fix(FOR-1086): Ability to define the number of decimal places after 

edwinanciani:
the comma for number and currency


src/components/currency/Currency.js
```diff
@@ -7,24 +7,20 @@ import {NumberComponent} from '../number/Number';
...
     // Get the prefix and suffix from the localized string.
-    const regex = `(.*)?100(${this.decimalSeparator === '.'
-      ? '\\.'
-      : this.decimalSeparator}0{${this.decimalLimit}})?(.*)?`;
-    const parts = (100).toLocaleString(options.language, this.getFormatOptions()).match(new RegExp(regex));
+    const regex = `(.*)?100(${this.decimalSeparator}0{${this.decimalLimit}})?(.*)?`;
+    const parts = (100).toLocaleString(this.getBrowserLanguage(), this.getFormatOptions()).replace('.', this.decimalSeparator).match(new RegExp(regex));
     this.prefix = parts[1] || '';
-    this.suffix = parts[3] || '';
+      this.suffix = parts[3] || '';
```
randallknutson: Why are we adding spacing here?

src/components/currency/Currency.js
```diff
@@ -7,24 +7,20 @@ import {NumberComponent} from '../number/Number';
 export class CurrencyComponent extends NumberComponent {
   constructor(component, options, data) {
     super(component, options, data);
-
     this.currency = this.component.currency || 'USD';
-    this.decimalLimit = this.component.decimalLimit || 2;
-
+    this.decimalLimit = options.decimalLimit || this.component.decimalLimit || 2;
```
randallknutson: Is there a reason we are adding an options.decimalLimit override? I don't think this was a requirement.

src/components/currency/Currency.js
```diff
@@ -7,24 +7,20 @@ import {NumberComponent} from '../number/Number';
...
     // Get the prefix and suffix from the localized string.
-    const regex = `(.*)?100(${this.decimalSeparator === '.'
-      ? '\\.'
-      : this.decimalSeparator}0{${this.decimalLimit}})?(.*)?`;
-    const parts = (100).toLocaleString(options.language, this.getFormatOptions()).match(new RegExp(regex));
+    const regex = `(.*)?100(${this.decimalSeparator}0{${this.decimalLimit}})?(.*)?`;
```
randallknutson: What was the purpose for these changes? The existing code came from standard libraries of how to locate the decimal and thousand separators. Is this going to cause any side effects?

src/components/currency/Currency.js
```diff
@@ -7,24 +7,20 @@ import {NumberComponent} from '../number/Number';
...
-    this.suffix = parts[3] || '';
+      this.suffix = parts[3] || '';
   }
 
   getFormatOptions() {
     return {
       style: 'currency',
       currency: this.currency,
-      useGrouping: true,
+      useGrouping: false,
```
randallknutson: Why are we changing this? Don't we want this to be true?

src/components/currency/Currency.js
```diff
@@ -54,8 +49,8 @@ export class CurrencyComponent extends NumberComponent {
       mask: createNumberMask({
         prefix: this.prefix,
         suffix: this.suffix,
-        thousandsSeparatorSymbol: _.get(this.component, 'thousandsSeparator', this.delimiter),
-        decimalSymbol: _.get(this.component, 'decimalSymbol', this.decimalSeparator),
+        thousandsSeparatorSymbol: _.get(this.component, 'thousandsSeparator', this.overrideLanguage.delimiter || this.delimiter),
```
randallknutson: This doesn't seem right. Instead of creating an override language and then checking it here, we should set scope.thousandsSeparator and scope.decimalSeparator when they are added to scope.

src/components/number/Number.js
```diff
@@ -2,23 +2,28 @@ import maskInput from 'vanilla-text-mask';
 import _ from 'lodash';
 import {createNumberMask} from 'text-mask-addons';
 import {BaseComponent} from '../base/Base';
+import currencyData from '../../currencyData';
 
 export class NumberComponent extends BaseComponent {
   constructor(component, options, data) {
     super(component, options, data);
+    this.currencies = currencyData;
```
randallknutson: Overrides should not be included in this pull request. They should be passed in with options to the form. So options.languageOverrides or something like that.

src/components/currency/Currency.js
edwinanciani: Fixing it right now.

src/components/currency/Currency.js
edwinanciani: I think there is a Requirement.

src/components/currency/Currency.js
edwinanciani: But that Previous regex gave me a wrong suffix when the decimalSeparator was ",". I changed it to this and works correctly.

src/components/currency/Currency.js
edwinanciani: That was testing purpose. I will fixed it right now.

src/components/number/Number.js
edwinanciani: I'm not disagree. But do you mean options creating the form it self? because is what I did before. I passed in the language list that you wanted to change, then compare the language with the current browser language and change the delimiter and decimal separator. Then I had a conversation with travis and I did that currency list, probably I did not understood travis.

randallknutson: Should be something like this: https://github.com/formio/ngFormio/pull/569/files#diff-34a7bb8e8a80000cf8069d3add22ef5cR61

PR #387, opened 2018-03-19
Bugfix/FOR-1112: Number default value.

YuryRybakSoftensity:


travist:
Because this seems to break reverse compatibility, we will only be able to include it in the 3.x branch with the form builder which isn't expected for another month. If you need to get this rolled out to current branch to fix other things, you will need to do this an maintain reverse compatibility (i.e. do not remove settings like "integer")

PR #388, opened 2018-03-19
Submit button was not disabled even if form was readonly

Aiso786:
I also added the added the ability to prevent a component from being disabled.

image

Then in my custom code, i can do this:
```javascript
if (component.properties && component.properties.alwaysEnabled === 'true') {
   component.alwaysEnabled = true;
}
```


travist:
looks good! Thanks!

PR #397, opened 2018-03-28
Feature/phraseapp

Vercjames:
Changed default submit error message from static text to a dynamic so it can be translated.

Resolved issue where customErrorMessage was always undefined and thus would never trigger the translate method.

### DIFF
--- a/src/components/Validator.js
+++ b/src/components/Validator.js
@@ -55,7 +55,6 @@ export const Validator = {
       });
      }
 
-
     return result;
   },
   validators: {
--- a/src/components/Validator.js
+++ b/src/components/Validator.js
@@ -47,12 +47,14 @@ export const Validator = {
       }
     });
 
-    const customErrorMessage = _.get(component, 'component.validate.customMessage');
-    if (result && customErrorMessage) {
-      result = component.t(customErrorMessage, {
-        data: component.data
+    const customMessage = _.get(component, 'component.validate.customMessage');
+     const validateCustom = _.get(component, 'component.validate.custom');
+     if (result && (customMessage || validateCustom)) { 
+      result = component.t(result, { 
+        data: component.data 
       });
-    }
+     }
+
 
     return result;
   },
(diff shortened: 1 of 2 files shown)


src/components/Validator.js
```diff
@@ -47,9 +47,9 @@ export const Validator = {
       }
     });
 
-    const customErrorMessage = _.get(component, 'component.validate.customMessage');
```
travist: Doesn't this break reverse compatibility?

Vercjames: I've added back in the 'component.validate.customMessage' to handle reverse compatibility. 

component.validate.custom is the entire string object for example: 
  "valid = (input === 3) ? true : 'Must be 3'" 

where 'result' is the actual string value. therefore 'result' needs to be translated.

travist: But you are not executing the value to get the result... you are basically translating ```valid = (input === 3) ? true : 'Must be 3'``` instead of ```Must be 3```. I don't see how this is working the way you expect.

src/components/Validator.js
```diff
@@ -47,12 +47,13 @@ export const Validator = {
...
-    if (result && customErrorMessage) {
-      result = component.t(customErrorMessage, {
-        data: component.data
+    const customMessage = _.get(component, 'component.validate.customMessage');
+    
const validateCustom = _.get(component, 'component.validate.custom');
+    
if (result && (customMessage || validateCustom)) {

+      result = component.t(result, {
```
travist: This now seems to be ignoring the ```customErrorMessage```. So what I am seeing here is if I have the following within custom....

```valid = input === 'test' ? true : 'You must enter "test"'```

and then have the following set for customErrorMessage...

```
You need to input "test" for this to work.
```

This shows that my customErrorMessage is ignored since it is just going to give me the translation of the result of the validation i.e 'You must enter "test"'.  This breaks all implementations that rely on the customErrorMessage.

src/components/Validator.js
travist: That's not right. Otherwise what is the point of ```customErrorMessage``` if it is never used or being ignored? This breaks reverse compatibility. Please fix.

PR #399, opened 2018-03-30
Bugfix/FOR-1076: DataGrid Min-Max.

YuryRybakSoftensity:


src/components/datagrid/DataGrid.js
```diff
@@ -51,7 +51,7 @@ export class DataGridComponent extends FormioComponents {
     ) {
       const toAdd = this.component.validate.minLength - this.dataValue.length;
       for (let i = 0; i < toAdd; i++) {
-        this.dataValue = this.dataValue.push({});
```
travist: I know this seems equivalent to what was there before, but because we are now using Getters and Settings for the dataValues, this may not be the same. If you see the getter for dataValue it looks like this.

```
get dataValue() {
    if (!this.component.key) {
      return this.emptyValue;
    }
    if (!this.hasValue) {
      this.dataValue = this.emptyValue;
    }
    return _.get(this.data, this.component.key, this.emptyValue);
  }
```

As long as there is an existing value in ```this.data[this.component.key]``` you are fine, but in the case that there isn't, this would just return the emptyValue which just an array for datagrids. This would turn this logic into

```
[].push({})
```

which does not do anything. This is why I explicitly set the value back with the setter so that this looks more like this...

```
_.set(this.data, this.component.key, _.get(this.data, this.component.key, []).push({}));
```

travist: Looking at this further, we may be ok because of this logic in the getter.

```
if (!this.hasValue) {
   this.dataValue = this.emptyValue;
}
```

but please verify.

PR #401, opened 2018-04-05
WIP - Formio Grid

edwinanciani:
Creating a Submission Grid with[ Handson Table.](https://docs.handsontable.com)


src/formio.grid.builder.js
```diff
@@ -0,0 +1,358 @@
+import _ from 'lodash';
```
travist: Why is this called ```GridBuilder```?

travist:
We have quite a bit that needs to be fixed here.

- Use handsontable and not handsontable-pro
- Use 2.x branch of handsontable
- We do not need a gridFactory remove it
- We do not need a createGrid file
- Copy contents from formio.grid.builder into formio.grid and then delete formio.grid.builder.
- Rename GridBuilder to FormioGrid
- Do not extend from FormioComponents in Grid class

travist:
Closing due to inactivity.

PR #407, opened 2018-04-09
Bugfix/FOR-731: HTML5 select with custom values

YuryRybakSoftensity:


src/components/select/Select.js
```diff
@@ -427,7 +422,7 @@ export class SelectComponent extends BaseComponent {
 
     if (this.component.widget === 'html5') {
       this.triggerUpdate();
-      this.addEventListener(input, 'focus', () => this.activate());
```
travist: We need to verify that this still works with the "lazyLoad" parameter turned on... where the list will load dynamically once it is active.

Issue #337, opened 2018-01-24
File upload not working


cabrerabywaters:
I'm trying to use the File upload component, but no matter what I use (Dropbox, S3, URL, Base64) I always get the same message when I try an upload "File Service not provided."
I did all the configuration for Dropbox and S3 at the project level, but nothing seems to work.

Any idea?

travist:
I assume you are setting the form IS IN directly on the form

You need to set the form url of the form in order to upload so it can create a signed upload url. You can do this using...

```
Formio.createForm(....).then((form) => {
   form.url = '......';
});

**Comment by cabrerabywaters** [comment]
Thank you @travist 
That was it!

**Comment by cabrerabywaters** [comment]
OK, so at least now I can upload a file as Base64 and it gets stored in the submission. 
Now if I save that and I go again to "Edit" (load the previous submission) I can see the files in the array but the File component does not show them.

Is there another trick for that?

**Comment by StephenOTT** [comment]
@travist your comment should be a Help text in the Upload component:
https://github.com/formio/formio.js/issues/337#issuecomment-360161652

as this is a SUPER easily missed and a easy blocker to run into!

**Comment by StephenOTT** [comment]
@travist also what is used at the `form.url` value when you inject a form object instead of a url to a form?

**Comment by StephenOTT** [comment]
For Base64 we made some changes to support it in the File.js

On like 137 of File.js we changed:

```
get fileService() {
    return this.options.fileService || this.options.formio;
  }
```

to:
```
get fileService() {
  if (this.options.fileService) {
    return this.options.fileService;
  }
  if (this.options.formio) {
    return this.options.formio;
  }
  return new Formio();
}
```

and added at the top of File.js

```
import Formio from ‘../../formio.js’;
```

all as per recommendation from @travist 


But now we are seeing duplicate base64 values being stored.  Is this expected?


```json
{
    "somefile": [
        {
            "storage": "base64",
            "name": "screenshot-b08ad188-422e-40c2-9384-6a0655623684.png",
            "url": "data:image/png;base64,iV....",
            "size": 67380,
            "type": "image/png",
            "data": "iV....",
            "originalName": "Screen Shot 2018-01-26 at 2.50.43 PM.png"
        }
    ],
    "textArea": "",
    "submit": false
}
```

travist:
Agreed this needs to be resolved. We will add it to the list of items to focus on next.

StephenOTT:
@travist any progress on this?

[...]

PR #414, opened 2018-04-18
Pull request to fix an issue with File component

asalhani:
I'm submtting this pull request as a suggested fix for an issue I discoved in File component. The issue was when you select **base64** as a storage in builder mode, in the submission, I couldn't upload the file and I kept getting error message:

> File Service not provided.

[File.zip](https://github.com/formio/formio.js/files/1923567/File.zip)

To fix this issue, I came an [opened  defect](https://github.com/formio/formio.js/issues/337), and I applied the suggested fix by @travist and @StephenOTT and it worked. The only remaining issue, which wait @travist fix as he promised is to fix base64 data duplicate in url and data fields which severally affect form rendering time especially if you have multiple files.

Suggested fixed code:
`

>   get fileService() {
>     if (this.options.fileService) {
>       return this.options.fileService;
>     }
>     if (this.options.formio) {
>       return this.options.formio;
>     }
>     return new Formio();
>   }

`

Now, in the submission mode for File component, the current implementation for _getFile()_ method is not valid anymore since Chrome has disabled the ability to open base64 images. So I modified the implementation to open file in a workaround using the _download_ property in HTML 5.

```
  getFile(fileInfo, event) {
    const fileService = this.fileService;
    if (!fileService) {
      return alert('File Service not provided');
    }
    fileService.downloadFile(fileInfo).then((file) => {
      if (file) {
        // commented becouse this way is not supported any more by chrome due to security practices.
        //window.open(file.url, '_blank');

        // this is a workaround to render base64 files in Chrome. Still not working on IE/Edge
        var hiddenElement = document.createElement('a');
        hiddenElement.href = 'data:' + file.type + ';base64,' + encodeURI(file.data);
        hiddenElement.target = '_blank';
        hiddenElement.download = file.originalName;
        hiddenElement.click();
      }
    })
      .catch((response) => {
        // Is alert the best way to do this?
        // User is expecting an immediate notification due to attempting to download a file.
        alert(response);
      });
    event.preventDefault();
  }
```

Please let me know if you need any further clarification.


StephenOTT:
@travist  thoughts on this fix?

travist:
I agree with these changes, but the one that concerns me is the change in how files are downloaded. I am worried because it removes the attempt at ```window.open```. Can we implement this as a fallback? Something like this?

```js
try {
  window.open(file.url, '_blank');
}
catch (err) {
  var hiddenElement = document.createElement('a');
  hiddenElement.href = 'data:' + file.type + ';base64,' + encodeURI(file.data);
  hiddenElement.target = '_blank';
  hiddenElement.download = file.originalName;
  hiddenElement.click();
}
```

This would work in Chrome if it throws an exception on ```window.open``` which I assume it does now due to the security changes.

asalhani:
@travist so should i open a new pull request with your suggested code or you can handle it within this one?

travist:
I would like if you would test the above change and if it works for you update this pull request with that change.

asalhani:
@travist sure no problem. I know it's out of this topic, but I really appreciate your kind input on a question I raised before. Thanks in advance.

https://github.com/formio/angular-formio/issues/166

randallknutson:
Hi asalhani, thanks for the commit. Could you clean up the diff for this pull request so it only shows the changes you actually made? Currently it shows a lot of stuff including the dist folder changes which makes it impossible to see what is being done. Thanks!

PR #415, opened 2018-04-19
Stop talking about FFiles.

colons:


randallknutson:
This was manually merged because of conflicts with dist directory. If you could not include those in the future it makes it easier to merge.

PR #430, opened 2018-05-01
Fix regressions with day component

randallknutson:


src/components/day/Day.js
```diff
@@ -51,6 +51,10 @@ export class DayComponent extends BaseComponent {
     return '';
   }
 
+  isEmpty(value) {
+    return value === '' || value === 'Invalid date';
```
travist: I would prefer the following.

```
return super.isEmpty(value) || (value === 'Invalid Date');
```

randallknutson: Yup

PR #434, opened 2018-05-02
fix issue when downloading a file of type base64

asalhani:
This is an enhanced pull request to fix an issue I reported previousley.
Kindly igonre the  [previous pull request](https://github.com/formio/formio.js/pull/414) and apply this one.


asalhani:
Enhance rendering of Signature component in ViewOnly mode such as the original captured signature will be rendered instead of [Yes/No].

src/components/file/File.js
```diff
@@ -509,7 +515,15 @@ export class FileComponent extends BaseComponent {
     }
     fileService.downloadFile(fileInfo).then((file) => {
       if (file) {
-        window.open(file.url, '_blank');
+        // commented becouse this way is not supported any more by chrome due to security practices.
+        //window.open(file.url, '_blank');
```
travist: We need this to remain, and only execute the code below if ```window.open``` fails. Please see if this works for your use case.

```js
try {
  window.open(file.url, '_blank');
}
catch (err) {
  var hiddenElement = document.createElement('a');
  hiddenElement.href = 'data:' + file.type + ';base64,' + encodeURI(file.data);
  hiddenElement.target = '_blank';
  hiddenElement.download = file.originalName;
  hiddenElement.click();
}
```

src/components/signature/Signature.js
```diff
@@ -79,9 +79,10 @@ export class SignatureComponent extends BaseComponent {
   }
 
   build() {
-    if (this.viewOnly) {
-      return this.viewOnlyBuild();
-    }
+    // no need to build a view only (which will display Yes/No) since we need to render the captured signuature instead.
+    // if (this.viewOnly) {
```
travist: Why did you comment this out?

travist: I see you added some code below that handles viewOnly.  You should just delete this code and not comment it out. We have GIT to keep track of changes.

src/components/file/File.js
asalhani: unfortunately it didn't work because Chrome consider this as a security issue rather than an exception occurred. So the code didn't continue to the fallback branch.

asalhani: @travist I just updated the pull request in away that the new code will run only if file upload mode is in base64. Otherwise, window.open() will work as is.

src/components/signature/Signature.js
asalhani: comment has been removed..

travist:
Looks good. Thanks!

PR #436, opened 2018-05-02
Fixed issues with the decimal limits of currency component being less…

travist:
… than 2.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -728,12 +728,18 @@ const FormioUtils = {
     lang,
   }) {
     // Get the prefix and suffix from the localized string.
-    const regex = `(.*)?100${decimalSeparator === '.' ? '\\.' : decimalSeparator}0{${decimalLimit}}(.*)?`;
+    let regex = '(.*)?100';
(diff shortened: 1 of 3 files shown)


src/utils/index.js
```diff
@@ -728,12 +728,18 @@ const FormioUtils = {
...
+      regex += '0{' + decimalLimit + '}';
+    }
+    regex += '(.*)?';
     const parts = (100).toLocaleString(lang, {
       style: 'currency',
       currency,
       useGrouping: true,
-      maximumFractionDigits: decimalLimit
+      maximumFractionDigits: decimalLimit,
+      minimumFractionDigits: decimalLimit
```
randallknutson: Are we sure we want to add minimumFractionDigits? Won't that require the full decimalLimit which is technically a "limit"?

randallknutson: Looks like some tests aren't passing. As long as they pass I'm fine with this.

travist: Yeah getting tests to pass and will commit this. Thanks.

PR #437, opened 2018-05-03
Update signature in view only mode

asalhani:


asalhani:
**This pull request contains the following:**
- [ File component: allow downloading file if it's base64] 
- [ Signature component: In submission mode, display the captured signature rather than (yes/no)]

PR #444, opened 2018-05-07
Restructuring the application.

travist:


src/index.js
```diff
@@ -1,9 +1,8 @@
...
-export {default as FormioPDF} from './formio.pdf';
+export {default as Formio} from './Formio';
+export {default as Form} from './Form';
+export {default as Wizard} from './Wizard';
+export {default as PDF} from './PDF';
 export {default as createForm} from './createForm';
 export {default as formFactory} from './formFactory';
-export {Formio};
-export default Formio;
+export {default as FormBuilder} from './Builder';
```
randallknutson: We should keep this name consistent. That way the imports would be
import {FormBuilder} from 'formiojs';
or
import FormBuilder form 'formiojs/FormBuilder';

or just use Builder for both.

travist: Ok

PR #457, opened 2018-05-14
Enhance downloading base64 file such as it works on all major browser (Chrome, IE, Edge)

asalhani:
This is to enhance my previous fix regarding downloading base64 file. The previous solution was working only on Chrome. However, I found [this library](https://github.com/rndme/download) which already handles downloading files (raw, base64, html,..). Moreover, the library supports most major browsers. I tested the code in Chrome, IE 12 and Firefox.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "browser-cookies": "^1.2.0",
     "choices.js": "3.0.3",
     "dialog-polyfill": "^0.4.9",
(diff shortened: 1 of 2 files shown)


travist:
I like this change. I would say if this library works well for all cases, then maybe we could also replace window.open with this library.

travist:
I will pull this in and make this the default for all file downloads. Great find!

PR #470, opened 2018-05-17
ie10 issue: "Unable to get property '1' of undefined or null reference"

zruchala:
ie10 rounds up number to 12345.68 thus the decimalSeparator is not set and the browser console is spoiled by message: '"Unable to get property '1' of undefined or null reference" '.


travist:
Looks good. Thanks!

PR #484, opened 2018-05-25
FOR-1190: Some improvements to EditGrid.

YuryRybakSoftensity:


src/components/editgrid/EditGrid.js
```diff
@@ -148,13 +140,13 @@ export default class EditGridComponent extends NestedComponent {
               }),
               this.ce('div', {class: 'editgrid-actions'},
                 [
-                  this.ce('div', {
+                  this.ce('button', {
```
randallknutson: Why are we changing these to a button? If we do then we need to use event.preventDefault and stop propogation.

PR #501, opened 2018-06-04
Change class references in static methods.

ElizavetaDeliatnik:


travist:
👏 👏 👏 

Well done. Had no idea how rampant this pattern was in the renderer.

PR #539, opened 2018-06-29
FOR-1239, FOR-1240, FOR-1309: fix nested forms validation

ElizavetaDeliatnik:


src/components/nested/NestedComponent.js
```diff
@@ -42,6 +42,18 @@ export default class NestedComponent extends BaseComponent {
     return this.components;
   }
 
+  getAllComponents() {
+    return this.getComponents().reduce((components, component) => {
```
travist: This will cause serious performance issues since you are iterating through ALL components at every level of the form. Imagine there is a form with 3 levels deep of nested components. This will iterate over the 3rd level deep components several times since at every level you are iterating over ALL components.

src/components/nested/NestedComponent.js
```diff
@@ -42,6 +42,18 @@ export default class NestedComponent extends BaseComponent {
...
+      let result = component;
+
+      if (component.type === 'form') {
+        result = component.subForm.getAllComponents();
```
travist: This will only grab the components at the first level of the nested form. What if the form has multiple levels deep of nested components and you wish to grab the components within them? Will this still work?

travist: Also, I don't like have ```component.type === 'form'``` anywhere. This is what we have derived classes for. You should create a method within the form component that does this.

ElizavetaDeliatnik: @travist It's a recursion there. I've checked nested forms up to three-level depth and everything works correctly.

I have changed a `component.type === 'form'` part as you suggested.

PR #547, opened 2018-07-09
Configure dragula to check if a component can be dragged using the 'no-drag' class

dierbro:
Configure dragula to check if a component can be dragged using the 'no-drag' class


pjmuller:
@travist , @randallknutson : can you guys merge this in?

PR #563, opened 2018-07-17
Fixed error appearing for Data Grid with invisible columns

Alena-Levina-Softensity:


src/components/datagrid/DataGrid.js
```diff
@@ -234,9 +234,11 @@ export default class DataGridComponent extends NestedComponent {
     }), options, row);
     comp.rowIndex = rowIndex;
     this.hook('addComponent', container, comp, this);
-    container.appendChild(comp.getElement());
     this.rows[rowIndex][column.key] = comp;
-    return container;
+    if (isVisible) {
+      container.appendChild(comp.getElement());
+      return container;
```
travist: Shouldn't you return the container even if the component is not visible?

Alena-Levina-Softensity: Do you think there is reason to pass empty container for invisible component?
In this case if we do nothing additional it will render empty cell for this component (though header will be missing - so it breaks the display of data grid)
Or do you mean something different?

travist: Oh, I didn't realize that returning nothing was an option.

PR #565, opened 2018-07-19
feature(components): Add property dataGridLabel to components and add setting to form builder

edwinanciani:
### Change
This commit will add a checkbox in the form builder settings section and will add a property in the component object called dataGridLabel. 

### Reason
When a Panel component is added into a Datagrid component the title or property key is shown in the Datagrid header.

### Screenshots
[image - 1](http://take.ms/1CNm9)

### DIFF
--- a/src/components/base/Base.js
+++ b/src/components/base/Base.js
@@ -86,6 +86,11 @@ export default class BaseComponent {
       tableView: true,
 
       /**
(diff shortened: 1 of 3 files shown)


travist:
Seems to break tests.

travist:
I fixed the tests and made this control conditional if the component is actually within a data grid.

PR #576, opened 2018-08-01
Data Grid: Fixed grid excess re-rendering on setValue

Alena-Levina-Softensity:


travist:
I think I may have fixed this issue with this less obtrusive change.

https://github.com/formio/formio.js/pull/579/commits/f444438c5585af0f54a7c0bbd10d9afbbf8741a5

Will you verify if this fixes this problem for you?

travist:
I included my fix above in version 3.1.0. I will also update ng-formio 3.x to next version as well. You may try that and see if it fixes this problem.

PR #583, opened 2018-08-07
feature(FOR-1534): Add Camera Plugin in File Upload Component

edwinanciani:
### Change
This commit will replace File Upload string section for 2 buttons that trigger Camera and PhotoLibrary plugins in a Webview, IF mobileDevices property is true in builder.

### Reason
In a WebView App the user couldn't activate the camera trying to upload a file.


travist:
I am going to pull this in manually. You added a lot of unnecessary formatting changes like your IDE settings are different than what we have configured in our .editorconfig (such as 4 spaces vs 2).

PR #597, opened 2018-08-14
FOR-1542: Update `build` method of `Container` component.

iobuhov:
`build` method now includes `createLabel` call.


iobuhov:
Please, when you will perform testing, apply label position `left (Right-aligned)` and `Right (left-aligned)`. It seems that position is broken.

PR #607, opened 2018-08-16
Protect against malformed language strings that may contain ;q=1

randallknutson:


travist:
It was requested from NSW team to put the logic of setting the timezones in ```try { } catch``` blocks so that even if the the library crashes they can still submit.

PR #608, opened 2018-08-16
FOR-1445: Fix addCurrentChoices method of SelectComponent.

iobuhov:
Problem was not in comparison function, but in source of options.


travist:
Looks good @iobuhov thanks!

PR #618, opened 2018-08-21
Fix in TextArea edit mode - An error occured within custom function f…

Drizin:
An error occured within custom function for wysiwyg TypeError: r.wysiwygDefault is not a function


travist:
Thank you!

PR #626, opened 2018-08-27
FOR-1644: Fixed required mark missing when Hide Label is checked

Alena-Levina-Softensity:


Alena-Levina-Softensity:
@travist I've found a bug with this PR, let me fix it before merging

Alena-Levina-Softensity:
@travist this should be good now

PR #630, opened 2018-08-28
FOR-1635: Allow webcam to take pictures.

edwinanciani:
### Reason
Allow webcam to scan and upload a doc to the form/s3 without saving it locally

### Change
Add webcam button when display as image is checked.


randallknutson:
There are 658 files changed including the dist folder. Could you clean up the commit so we can see what actually changed?

PR #635, opened 2018-08-30
Add 'display in Edit Grid' setting to the button component.

ElizavetaDeliatnik:


src/components/button/Button.js
```diff
@@ -381,5 +381,16 @@ export default class ButtonComponent extends BaseComponent {
   focus() {
     this.buttonElement.focus();
   }
-}
 
+  getView() {
+    if (this.component.displayInEditGrid) {
+      this.build();
```
travist: Why are you having to "build" the component here. ```getView``` will usually get called after the component has already been built.

src/components/button/Button.js
```diff
@@ -381,5 +381,16 @@ export default class ButtonComponent extends BaseComponent {
...
+       setTimeout(() => {
+        const element = document.getElementById(this.id);
+        element.parentNode.replaceChild(this.element, element);
+      });
+       return this.element.outerHTML;
```
travist: This really seems strange to me. I think I may need you to show me tomorrow what this is doing.

travist:
@ElizavetaDeliatnik need a response to my inquiries... What are you going to do with this

PR #640, opened 2018-09-05
Builder: Error when having an hidden component inside DataGrid

Aiso786:
My DataGrid has an "Hidden" component inside a column.
When using the builder, i had the following error:

`Cannot read property 'noDrop' of undefined
    at DataGridComponent.WebformBuilder._this.options.hooks.addComponent (WebformBuilder.js?8806:171)`

This pull request solves the issue.
A column containing an hidden component is now visible while building the form and hidden when the form is rendered.

**Added new option on the Tabs design component:**
image

When set, each tab\'s content will be rendered on the first load. By default, only the first tab has its content rendered.

## UPDATE
I reverted the changes on the Tabs Component, this PR is now focused on resolving the issue on the DataGrid in Builder mode.


travist:
I totally get the first part of the changes you made to allow the hidden components to be visible when in "builder" mode. That is an easy contribution, so it may be best to separate that out by itself into a separate pull request.

The other addition I am not so sure about. What is the purpose to render all of the contents of the tabs on the first tab? I do not see why this is needed.

Aiso786:
@travis it's done.

travis:
@travist ^^

travist:
Looks good thanks.

PR #641, opened 2018-09-05
Fix validation error styles

ElizavetaDeliatnik:


src/sass/formio.form.scss
```diff
@@ -676,3 +676,7 @@ body.formio-dialog-open {
 .is-disabled .choices__list--multiple .choices__item .choices__button {
   display: none;
 }
+
+.has-error.alert-danger {
+  background-color: #f2dede;
```
travist: This is hard-coding the color to a pretty opinionated color and will break when people run this with bootswatch or some other style system. It also "feels" like we are hacking something here. Why do the styles not apply when a component .has-error and also is in alert-danger. This seems like it would already have the red color applied to it.

PR #644, opened 2018-09-06
Tabs: fixing issue where only the first tab had its content rendered.

Aiso786:
When we use the Tabs component, only the components of the first tab are rendered. The other tabs are empty.
So if i dont navigate to the other tabs and these tabs contain mandatory fields for example, i can still submit the form. I wont have any validation errors.

### DIFF
--- a/src/components/tabs/Tabs.js
+++ b/src/components/tabs/Tabs.js
@@ -108,14 +108,7 @@ export default class TabsComponent extends NestedComponent {
     }
 
     this.currentTab = index;
-
-    // Get the current tab.
-    const tab = this.component.components[this.currentTab];
-    this.empty(this.tabs[this.currentTab]);
-    _.remove(this.components, (comp) => comp.component.tab === this.currentTab);
-    const components = this.hook('addComponents', tab.components, this);
-    _.each(components, (component) => this.addComponent(component, this.tabs[this.currentTab]));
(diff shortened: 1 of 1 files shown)


travist:
I don't see how this actually changes anything. You moved some logic into its own method and the only time it will get called is under the same conditions as it was before. I am not sure how this solves your problem. It may be helpful to produce a jsfiddle before and after this change is made so that I can see the problem.

Otherwise, this will need to wait until next release.

Aiso786:
Ok i'll do that. But i think you missed this part:

Image

Aiso786:
@travist Here is the JsFiddle:
http://jsfiddle.net/ubL1gqed/

We have a Tabs component containing 2 tabs. The second tab contains a required text field.
If you don't open the second tab, you'll be able to submit the form because the required text field is not rendered.

travist:
I think there is a more performant way to do this using the beforeSubmit hook. This way, we are not adding all the tab components during the build process. This may cause problems if people implement tabs to get around having a lot of fields rendered per page. The wizard component (https://github.com/formio/formio.js/blob/master/src/Wizard.js#L95) and form component (https://github.com/formio/formio.js/blob/master/src/components/form/Form.js#L216) do this.

Aiso786:
I close the PR. Don't know yet how to do it otherwise. @travist

PR #652, opened 2018-09-10
submitDone

kaleguy:
The submit button stays disabled and with spinner unless the submitDone event is emitted.


travist:
Thanks!

PR #657, opened 2018-09-13
FOR-1417: Update caching logic.

iobuhov:


src/Formio.js
```diff
@@ -848,8 +843,20 @@ export default class Formio {
...
+        if (method === 'GET') {
+          delete Formio.cache[cacheKey];
+        }
+
         return Promise.reject(err);
       });
+
+    // Cache the response.
+    if (method === 'GET') {
+      Formio.cache[cacheKey] = result;
```
travist: Since the ```result``` is now a promise, you can probably change the point where it returns the result to the caller to not pass it through ```Promise.resolve``` anymore but rather just say.

```
// Get the cached promise to save multiple loads.
if (!opts.ignoreCache && method === 'GET' && Formio.cache.hasOwnProperty(cacheKey)) {
  return Formio.cache[cacheKey];
}
``

PR #658, opened 2018-09-13
FOR-672: Add uiquify step to copyComponent method.

iobuhov:


src/WebformBuilder.js
```diff
@@ -441,6 +441,7 @@ export default class WebformBuilder extends Webform {
     }
     this.addClass(this.element, 'builder-paste-mode');
     const copy = _.cloneDeep(component.schema);
+    BuilderUtils.uniquify(this._form, copy);
```
travist: This should probably occur on the ```pasteComponent``` method since I can see a problem where you could add another component between the point where you copied it and pasted it and then you would have another collision.

randallknutson: Also, you can paste multiple times which would cause a collision.

PR #666, opened 2018-09-17
FOR-1507: Fix button component.

iobuhov:
Add validation checking for case when 'action' is
set to 'event'.

### DIFF
--- a/src/components/button/Button.js
+++ b/src/components/button/Button.js
@@ -191,6 +191,14 @@ export default class ButtonComponent extends BaseComponent {
         this.loading = false;
       });
     }
+
+    if (this.component.action === 'event') {
+      this.on('change', (value) => {
+        const isValid = this.root.isValid(value.data, true);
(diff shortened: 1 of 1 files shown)


src/components/button/Button.js
```diff
@@ -191,6 +191,14 @@ export default class ButtonComponent extends BaseComponent {
         this.loading = false;
       });
     }
+
+    if (this.component.action === 'event') {
+      this.on('change', (value) => {
+        const isValid = this.root.isValid(value.data, true);
```
travist: The validity is passed as ```value.isValid``` so I would prefer to use that over an additional check on root. Should probably say.

```
const isValid = value && value.hasOwnProperty('isValid') ? value.isValid : this.root.isValid(value.data, true);
```

iobuhov: Ok, I'll push commit with update soon.

PR #669, opened 2018-09-18
Move field logic event trigger to build so context is set correctly a…

randallknutson:
…nd interpolate event so it will work with rowIndex.


src/components/button/Button.js
```diff
@@ -206,13 +206,12 @@ export default class ButtonComponent extends BaseComponent {
           });
           break;
         case 'event':
-          this.emit(this.component.event, this.data);
-          this.events.emit(this.component.event, this.data);
```
travist: We need to keep the ```this.events.emit(this.component.event, this.data);``` so that an event is triggered without the namespace which I think is needed for some of the framework wrappers.

PR #672, opened 2018-09-19
FOR-1705: Fixed HTML Element and Content components losing content when any Logic is applied

Alena-Levina-Softensity:


Alena-Levina-Softensity:
@travist @randallknutson I've also fixed same issue for Content component same way - by adding regular wrapper element. But in case of Content it changed the place where customClass is applied (previously it was applied to html content of component and now it is applied to wrapper). I'm not sure it is OK, could you please provide your feedback?

Alena-Levina-Softensity:
Also I was wondering, shouldn't we add 'CSS Class' (className) setting to Content (same as HTML Element has) which would add CSS class to the div with actual content?

Alena-Levina-Softensity:
Discussed with Randall and Travis and decided following:
1. Fix for Content component is OK
2. Decided to add CSS Class property to Content Element

PR #697, opened 2018-10-01
FOR-1558: Add submission setter test.

iobuhov:
Add test case for forms that contain containers or layout
components.


iobuhov:
Simple test to approve that [1558](https://formio.atlassian.net/browse/FOR-1558) already fixed.

PR #701, opened 2018-10-01
Feature/FOR-1567: DateTime dynamic change.

YuryRybakSoftensity:


src/widgets/CalendarWidget.js
```diff
@@ -92,9 +92,9 @@ export default class CalendarWidget extends InputWidget {
       this.setInputMask(this.calendar._input, convertFormatToMask(this.settings.format));
 
       // Make sure we commit the value after a blur event occurs.
-      this.addEventListener(this.calendar._input, 'blur', () =>
+      ['blir', 'keyup'].map((event) => this.addEventListener(this.calendar._input, event, () =>
```
travist: Shouldn't this be ```blur```?

travist:
I had to revert this pull request due to the issue it caused here.

Alt text

PR #706, opened 2018-10-02
FOR-1303: Add min/max validators for Day component.

iobuhov:


travist:
This should also be dynamic like what we have in the Angular 1 renderer. This allows you to type into a textfield something like ```moment().add(1, 'day')``` which will make the maxDate always 1 day in the future. The way we did this in Angular 1 was to make the input a textfield where they can either provide a date (using the calendar dropdown), and also allow them to type something.

PR #710, opened 2018-10-04
Update table component

iobuhov:


iobuhov:
Travis, If you can please look at my comment about `get schema` method.

travist:
Please be more descriptive with your commit messages. For example.

```FOR-1234: Fixed issue with the Table component where it would not allow you to configure the rows or columns to be less than 3x3.```

PR #752, opened 2018-10-26
Bugfix/FOR-1745: clearOnHide context.

YuryRybakSoftensity:


src/components/datagrid/DataGrid.js
```diff
@@ -318,6 +318,7 @@ export default class DataGridComponent extends NestedComponent {
   setValue(value, flags) {
     flags = this.getFlags.apply(this, arguments);
     if (!value) {
+      this.dataValue = this.emptyValue;
```
travist: This says ```emptyValue``` but the EditGrid says ```defaultValue``` which one is correct?

PR #760, opened 2018-10-30
handle exception

JoeyLi-1:
As mentioned here: https://github.com/formio/angular-formio/issues/217

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -1135,7 +1135,7 @@ export default class Webform extends NestedComponent {
    */
   submit(before, options) {
     if (!before) {
-      return this.beforeSubmit(options).then(() => this.executeSubmit(options));
+      return this.beforeSubmit(options).then(() => this.executeSubmit(options)).catch(console.log);
     }
     else {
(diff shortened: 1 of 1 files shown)


src/Webform.js
```diff
@@ -1135,7 +1135,7 @@ export default class Webform extends NestedComponent {
    */
   submit(before, options) {
     if (!before) {
-      return this.beforeSubmit(options).then(() => this.executeSubmit(options));
+      return this.beforeSubmit(options).then(() => this.executeSubmit(options)).catch(console.log);
```
travist: Calling ```catch(console.log)``` is pretty dangerous since we don't know if the ```this``` pointer would ever be used within the console object. If so, then your call here would crash since ```this``` would be pointing to the window object. 

This should probably say ```catch(err => console.log(err))``` or actually, we would probably prefer to use ```console.warn``` so change it to this.

```js
catch(err => console.warn(err));
```

JoeyLi-1:
@travist You are right. Fixed.

Issue #753, opened 2018-10-26
Wrong visibility resolution in Column component


kamsi:
I believe this is incorrect:

https://github.com/formio/formio.js/blob/v3.6.11/src/components/columns/Column.js#L11

becaue in case of component not having any children it will set `allChildrenHidden` to true and return `false`.

And since it will return `false`, the component will be hidden by this line:
https://github.com/formio/formio.js/blob/v3.6.11/src/components/nested/NestedComponent.js#L463

I believe it should not be hidden in that case. Or am I wrong and it is by design that if a column component has no children it should be hidden?

wag110894:
Hello Kamil,
 
This is William with the Form.io support team. Thanks for reaching out!   

This is actually by design. We would consider making  it into a configuration option if you want to request that feature.  

Please let me know if you would like a quote to add this to the platform.  We offer discounted rates for customer driven platform enhancements.

Thanks, 
William Gay 
Form.io Developer Support

kamsi:
@wag110894 thanks for prompt reply.

Yes, I would like to be able to control that behavior. In our project we are currently switching from 2.x formio (wrapped in ng-formio) to 3.x (through vueJS) and the exact same form renders differently because of this.

wag110894:
Hello Kamil,

We have added this feature to our backlog.For reference, the ticket number is FOR-1828. It may be awhile before you see this implemented. If you need it sooner, your options are to make the changes to the renderer and submit a pull request and we will accept it for our next release, or you can pay us to expedite this for you. It would probably take 4 hours of development to make this configurable and we offer a discounted rate for expedited issues. 
 
Please let me know if you are interested and I can send you a formal quote.
 
Thanks, 
William Gay 
Form.io Developer Support

kamsi:
@wag110894

The PR is here: https://github.com/formio/formio.js/pull/763

PR #763, opened 2018-10-30
#753 add the 'hideOnChildrenHidden' component param

kamsi:
Addresses: https://github.com/formio/formio.js/issues/753

### DIFF
--- a/src/components/columns/Column.js
+++ b/src/components/columns/Column.js
@@ -4,13 +4,14 @@ export default class ColumnComponent extends NestedComponent {
   constructor(component, options, data) {
     super(component, options, data);
     this.noEdit = true;
+    this.hideOnChildrenHidden = this.component.hideOnChildrenHidden != null ? this.component.hideOnChildrenHidden : true;
   }
 
   conditionallyVisible(data) {
     // Check children components for visibility.
     const allChildrenHidden = _.every(this.getComponents(), ['visible', false]);
 
-    if (allChildrenHidden) {
+    if (allChildrenHidden && this.hideOnChildrenHidden) {
       return false;
     }
 


src/components/columns/Column.js
```diff
@@ -4,13 +4,14 @@ export default class ColumnComponent extends NestedComponent {
   constructor(component, options, data) {
     super(component, options, data);
     this.noEdit = true;
+    this.hideOnChildrenHidden = this.component.hideOnChildrenHidden != null ? this.component.hideOnChildrenHidden : true;
```
travist: We cannot change the default behavior. You are defaulting this to ```true``` when it should be ```false```. Because of this, I think the variable should be called ```showOnChildrenHidden``` and default it to ```false```. 

There is also a better way of doing this where you would just add this new property to the ```schema``` definition. Then you can put this in the check.

```js
if (allChildrenHidden && !this.component.showOnChildrenHidden) {

}
```

kamsi:
@travist

My default is fine. It works exactly as it used to when you do not setup it: defaults to true and so the previous behavior (which I consider weird) is preserved and the component will hide. Please read it again carefully.

I also disagree with the rename to showOnChildrenHidden. What would it mean? That no matter anything else the component will show if it does not have children? I believe my name better reflects what is being done here: you will hide the component if hideOnCh... is true, but if it is not yoi will continue the regular resolution logic.

I could probably not write to component variable and resolve it straight from component... not sure that is such a gain though.

To sum it up

1. Code works fine as it is
2. Rename to shoOn... I consider semantically inapropriate because it will be misleading.

Yet, if you still think it should be showOn... tell me so and I will do it.

travist:
I would normally like ```hideOnChildrenHidden```, but the issue I have has more to do with "default" behavior if that property is not present. What you have here suggests that if it is not present, it is TRUE since that is the current behavior. ***I typically don't like introducing variables that are the inverse of the default.***  ```showOnChildrenHidden``` does have meaning and is the inverse of the variable above. It also makes sense that it is FALSE if the variable is not present meaning that FALSE would work the same as if it were provided and false. This keeps the variable inline with "reverse compatibility".

I would say let's have another developer chime in and give their opinion on this. @randallknutson would be a good person to give his opinions.

kamsi:
@travist sure, I will settle for whatever verdict @randallknutson gives.

I get your point on inverse. To be perfectly honest I believe that current default is not sane (I described that in the issue: https://github.com/formio/formio.js/issues/753 ). At least not for the empty case. Should we really hide component just because it does not have children?
My dreams come true would be actually a different default for the no children case.
So sth like:

  `
if (allChildrenHidden && this.getComponents().length) {
  return false;
}
`

But this would indeed break the current behavior for this edge case and hence, would probably have to be in major.

randallknutson:
Ideally we would have made an option to hide a column if all components within it were hidden however we are now in a position where the default is to hide the column if the all components are hidden. 

For now the best we can do is to set the property in a backwards compatible way like Travis said. Moving forward this is something we should clean up in a new schema major version.

travist:
@kamsi @randallknutson if I remember correctly, we recently added the functionality of hiding the columns if all children are hidden. I know it may sound blasphemous, but in this particular case, I am not going to put up much of a fight to keep reverse compatibility since it wasn't that many versions ago that we actually changed "default" behavior to hide the column when children are hidden. It should also be noted that we have always anchored to the way the Angular 1x renderer works when deciding default behavior for this renderer. The Angular 1.x renderer does not hide the column either.

We could say that move introduced a "bug" and are now creating a configuration so that it goes back to the way it was, but does allow you to make it hide to appease the customers we added this functionality for. Give us some time to reach out to these customers and get their take on how disruptive this would be for them, and then we should make it behave the way we think it should behave, regardless of keeping reverse compatibility in place.

travist:
We will take this PR, and increment the minor version. The reasoning is that we recently broke reverse compatibility when we introduced the hide column on children hidden logic, so that could be considered an introduced bug. This will make it configurable and default it to show the column.

kamsi:
@travist I think there is still some confusion here. My PR will make it so that the default behavior will hide that column when no children!
If you want to show the column by default when children are hidden, then hideOnCh... needs to be false.

travist:
I made changes to your pull request to apply the correct default. Please verify on your end, but it should be good.

kamsi:
@travist all good! Thank you.

PR #765, opened 2018-10-30
FOR-1731: Fix bug with Content component in Wizard builder.

iobuhov:
Prior the bugfix while editing content component in wizard builder,
it looses styled content when switching/adding/deleting pages.
Now all content stays the same.


travist:
Well done @iobuhov.

PR #779, opened 2018-11-05
Feature/FOR-1803: Hide PDF submit.

YuryRybakSoftensity:


src/PDF.js
```diff
@@ -143,7 +145,11 @@ export default class PDF extends Webform {
       this.iframe
     ]);
 
-    if (!this.options.readOnly) {
+    if (
+      !this.options.builder &&
+      !this.options.readOnly &&
+      !(String(_.get(this.form, 'properties.formio.hidePdfSubmit', false)) === 'true')
```
travist: Now that I am looking at this, I am not sure I like the idea of introducing yet another configuration when the form schema should be enough information if they want to hide or show the submit button on the PDF. Why can we not just allow them to remove the submit button in the builder? If they want to re-introduce it, then they can switch the display back to form and re-add it there.

travist:
@YuryRybak what is the status of this pull request. I feel it has gone stale.

YuryRybakSoftensity:
@travist Done.

PR #783, opened 2018-11-06
Fix issues with inifinite refeshes with the datagid

randallknutson:
This reverts parts of two changes that caused issues:

https://github.com/formio/formio.js/commit/f53f7c4fc4e9926aa5d56dd73b4b97d5aeab1922
https://github.com/formio/formio.js/commit/2fb62ee95170c7445f6f78a4bf64251ca956fcf2

### DIFF
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -71,7 +71,11 @@ export default class CheckBoxComponent extends BaseComponent {
   }
 
   get emptyValue() {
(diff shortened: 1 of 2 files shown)


travist:
Can you verify that you did not break https://formio.atlassian.net/browse/FOR-1647 by reverting?

travist:
Can you also verify that you did not break https://formio.atlassian.net/browse/FOR-1745 by reverting that change as well?

PR #795, opened 2018-11-15
FOR-1787: Day -> Added ability to have empty parts, added trailing zeros

Alena-Levina-Softensity:


randallknutson:
@travist This was discussed yesterday while you were out. Essentially we changed this to work more like how the angular1 (ngFormio) renderer works since that seems to be how people expect it to work. 

Changes are:
1. Not using moment or date functions to validate the day. Instead using a manual calculation of max days per month
2. If a field is hidden or not entered, the default value is zero instead of the current day/month or year.
3. Add leading zeros to all numbers so that the format is always 00/00/0000. 
4. Skip min day/max day validation if they don't have all three fields visible. They can still add custom validation but we can't reliably calculate min/max validation without a full date.

travist:
I am alright with this as long as it is consistent with ngFormio behavior.

src/i18n.js
```diff
@@ -17,6 +17,7 @@ export default {
         invalid_url: '{{field}} must be a valid url.', // eslint-disable-line camelcase
         invalid_regex: '{{field}} does not match the pattern {{regex}}.', // eslint-disable-line camelcase
         invalid_date: '{{field}} is not a valid date.', // eslint-disable-line camelcase
+        invalid_day: '{{field}} is not a valid day.', // eslint-disable-line camelcase
```
travist: Why do we need to have a separate error? I think ```is not a valid date.``` is good enough.

PR #826, opened 2018-11-29
Fix(FOR-1730) Drag n Drop Files in form

edwinanciani:
## Reason 
When a user drags a file into the form (and he misses the drop area for the file uploader). We get the default browser behavior where the file is shown


travist:
This will probably break the builder. Have you tested it against the builder?

travist:
Closing due to inactivity.

PR #830, opened 2018-11-30
Enable Custom Builder Components Pre-Defined Key

DavidPaparo:
The Business need this fix pertains to is to allow the Custom Formbuilder to pre-define Key names. Although the component could have its key pre-defined, it would always be overwritten by the "Ensure this component has a key" logic due to the fact component.keyModified is undefined.

The proposed fix will now check if the new component has a predefined key, and if so it will leave the key alone. Should no key be predefined, it will simply continue its regular logic of defining one for the component.


travist:
I am pretty sure that this change will break auto-key creation when they are typing in the label. It looks like it would just get set to the first character of the label. Can you ensure that this does not happen?

DavidPaparo:
You were dead on the money with that assumption. The key would be generated via the label and would no longer modify based on your changes. Be that this is not something accomplishable with the current set of component properties, I opted to add a new property (lockKey). This will allow developpers to "lock the key" when defining custom components.

Note this can be brought into the edit screen, but I see very little use for it. It is purely to satisfy the situation where custom component keys should not be modified.

travist:
I'm am better with this, but I think we still have a problem. I did a quick search throughout the codebase and do see that there is reference to a ```keyModified``` variable and now this pull requests changes that name. I am guessing this will break the other places that ```keyModified``` is used.

DavidPaparo:
Understood. In order to prevent breakages otherwise, I could simply add an "And" statement instead of replacing the keyModified entirely (you can see this in the current branch). This way the system will continue as expected unless someone has set the lockKey property to true. I Have personal concerns about potential code bloat using this method, but I will leave the final call in your hands.

travist:
Closing due to inactivity. If this is still needed for 4.x the please submit that there.

PR #849, opened 2018-12-10
FOR-1336: Fix bug with double validation

iobuhov:
Now submit onchange event don't trigger second validation.


test/forms/validationCalls.js
```diff
@@ -0,0 +1,115 @@
+import sinon from 'sinon';
```
travist: Thanks for writing tests!

Issue #847, opened 2018-12-08
SelectComponent Translation Direction Support - RTL


imadbakir:
Hello, 

SelectComponent has (hardcoded?) dir attribute equals to "ltr", 
which breaks the styles in arabic language even if I set dir of my document to "rtl", the "ltr" on the select will "undo" that, thus broken RTL.

airarrazaval:
@randallknutson @travist Choices.js includes RTL support by setting the `.choices` element direction attribute to `dir="rtl"`.  When initializing the choices instance, if an html element is passed, the choices container direction is set to the passed element's direction, defaulting to `ltr` if none set (see [choices.js](https://github.com/jshjohnson/Choices/blob/826384b9d5a3e7fc1575fb0a0545f294705b8dd9/src/scripts/choices.js#L126)).

Currently Select, Resource (based on Select) and Tags Components uses choices.js and initialize its instance in the `addInput()` function.

### Getting the current language direction

Using the current i18next instance, any component can retrieve the current form language direction using the following:

```js
this.i18next.dir();
```

### Setting the direction

To address this issue we have some options:

#### Option 1: `addInput()`

This can be achieved modifying the Select Component and Tags Component `addInput()` function.

https://github.com/formio/formio.js/blob/e5e2c5f7afa0461b0b8c270e19b5750cf78666b2/src/components/select/Select.js#L470-L488


The following line needs to be included:

```js
addInput(input, container) {
  container.setAttribute('dir', this.i18next.dir());
  .
  .
  .
}
```

imadbakir:
@airarrazaval wow! Thanks for the in depth detailed answer that def was helpful, I will choose the last one to implement for now as you suggested

airarrazaval:
Further looking into the language setter, this already calls `this.redraw()` so its only a matter to add the dir attribute once in Select and Tags component.  I have created pull request https://github.com/formio/formio.js/pull/850.

imadbakir:
@airarrazaval sounds good!
I will apply the workaround until merge with master. Thank you!

travist:
Thanks @airarrazaval and @imadbakir. We will work on getting a new renderer released this week with your changes. Thanks!

darabi:
@travist: I came here while searching for 'translation'. Seems like this issue was resolved, would you mind closing it?

PR #850, opened 2018-12-10
added-choices-rtl-support

airarrazaval:
Added `dir` attribute to component's input element to provide Choices.js RTL support (see issue https://github.com/formio/formio.js/issues/847)


travist:
Well done!

PR #852, opened 2018-12-10
Fix issue translation for html element HTML.js

AnasHijazi:
Add Translation to html component

### DIFF
--- a/src/components/html/HTML.js
+++ b/src/components/html/HTML.js
@@ -11,7 +11,7 @@ export default class HTMLComponent extends BaseComponent {
       content: '',
       input: false,
       persistent: false,
-	  translateHtml: false,
+	  translateHtml: 'false',
     }, ...extend);
   }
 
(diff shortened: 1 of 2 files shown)


src/components/html/HTML.js
```diff
@@ -31,7 +31,16 @@ export default class HTMLComponent extends BaseComponent {
 
   setHTML() {
     this.htmlElement.innerHTML = this.interpolate(this.component.content);
-  }
+    const walker = document.createTreeWalker(this.htmlElement, NodeFilter.SHOW_TEXT, null, false);
```
travist: We should probably make this a configuration since I know this would probably kill the performance of some forms with very large content regions. We should call the config for this ```translateContent```

AnasHijazi: @travist  how do this ?

travist: Add a new configuration setting within the HTML edit form seen here... 

https://github.com/formio/formio.js/blob/master/src/components/html/editForm/HTML.edit.display.js

Then you can reference that configuration within here using 

```
if (this.component.translateHtml) {
```

You will also want to add that property with the default to the schema at https://github.com/formio/formio.js/blob/master/src/components/html/HTML.js#L6

It should be defaulted to ```false```

AnasHijazi: Hi travist,
 i added config as select component has true and false options. please check below : 
https://github.com/AnasHijazi/formio.js/invitations

src/components/html/editForm/HTML.edit.display.js
```diff
@@ -59,4 +59,19 @@ export default [
     key: 'refreshOnChange',
     input: true
   },
+  {
+  type: 'select',
```
travist: Spacing is incorrect.
Also, can we just make this a Checkbox so that the setting is a boolean instead of a string?

AnasHijazi: @travist  check now

AnasHijazi: @travist  it's ok ?

travist:
I just pulled this in, and it is pretty clear that you never even tested this code. It is full of syntax errors. Please fix that as well as write tests before we will pull this into our codebase.

PR #858, opened 2018-12-11
Fix #814

### LINKED ISSUE #814: HTML view - Resources as objects
Hi,

With **readOnly** and **viewAsHtml** as true, resources are showed as a plain JSON object instead of showing it value like angular does, since everything looks fine on formio Server.

Note that **Allow Multiple Resources** is enabled.

---

Server:
screenshot from 2018-11-23 12-29-49

SDK:
screenshot from 2018-11-23 12-30-17

Thanks.

Ggleny:
Fix Select Resource multiple

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -94,7 +94,31 @@ export default class SelectComponent extends BaseComponent {
       return this.t(data);
     }
 
-    const template = this.component.template ? this.interpolate(this.component.template, { item: data }) : data.label;
+    //const template = this.component.template ? this.interpolate(this.component.template, { item: data }) : data.label;
+    var template, me=this;
+    var itemTemplate = this.component.template;
+    if(itemTemplate ){
+      if(Array.isArray(data) ) {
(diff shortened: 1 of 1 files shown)


travist:
Thanks for the contribution @Ggleny. We will need this code to be cleaned up  since it does not seem to follow any coding guidelines. If you take a look at our .eslint and editorconfig files, you will see the guidelines that we have put forth for this project.

https://github.com/formio/formio.js/blob/master/.editorconfig
https://github.com/formio/formio.js/blob/master/.eslintrc.json

airarrazaval:
I have added pull request https://github.com/formio/formio.js/pull/867 that addresses this in a better way.  In `asString()` function there is no control over select components with multiple option (typeof value is Array).  thus a simple check and forEach loop is enough.

travist:
I agree @airarrazaval that your approach seems less intrusive. Thank you so much @Ggleny for your contribution, but we will close this one in favor of the one from @airarrazaval.

PR #861, opened 2018-12-13
Fix issue where if radio value is true it won't have radio-selected c…

randallknutson:
…lass set on it.


src/components/radio/Radio.js
```diff
@@ -202,7 +202,7 @@ export default class RadioComponent extends BaseComponent {
 
       _.each(this.wrappers, (wrapper, index) => {
         const input = this.inputs[index];
-        if (input.value === value) {
+        if (input.value === value.toString()) {
```
travist: To be more robust, should we make this...

```if (input.value.toString() === value.toString()) {```

PR #866, opened 2018-12-15
Currency Component: Support other currencies - Add suffix / prefix options

abarrak:
This improves the currency component by exposing `currency`, `prefix`, `suffix` attributes via the builder.

Additionally, [browser-supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) ISO-4217 currencies are maid available as select list.

currencies 
suffix-prefix


travist:
Sweet!  Thank you @abarrak!!!

src/components/currency/editForm/Currency.edit.display.js
```diff
@@ -0,0 +1,18 @@
+export default [
+  {
+    type: 'textfield',
+    input: true,
+    weight: 310,
+    key: 'prefix',
```
travist: We already have a ```prefix``` and ```suffix``` for the text field components. We should either use different names for these or hide the ones from the display tab.

abarrak: @travist could you clarify more?

As far as I can see ‘Currency’ inherits from ‘Number’, and has differed suffix/prefix defined [here](
https://github.com/formio/formio.js/pull/866/files#diff-46b3380c5b96fc2e752f50913ef0c35cR40) ?

travist: Yeah OK, I think this is good.

PR #870, opened 2018-12-17
Replace browser-cookies with hardtack

alik0211:


randallknutson:
This looks really good and we're interested in accepting it. Can you explain what the advantages of switching to hardtack are?

alik0211:
1) hardtack easier browser-cookies on 50%
https://wzrd.in/standalone/browser-cookies@1.2.0 gzipped - **899 bytes**
https://cdn.jsdelivr.net/npm/hardtack@2.1.0/dist/hardtack.min.js gzipped - **431 bytes**

2) hardtack works flexibly with options. In browser-cookies all options are hard-coded.
https://github.com/alik0211/hardtack/blob/master/src/hardtack.js#L9-L25
https://github.com/voltace/browser-cookies/blob/master/src/browser-cookies.js#L9-L14

3) browser-cookies are not updated for more than 9 months. hardtack I will keep a very long time.

travist:
I really like your library. Give us some time to discuss this internally some more.

travist:
I think we have decided that we are going to hold off on this pull request until your repository has more people using it. We do not like the idea of basing our cookie library on a library that is not well known and does not have the amount of users following it and ensuring it works for all scenarios. We will just hold this here for now, but that is why we are not pulling this into the builds.

alik0211:
I have some good news. [Andrey Sitnik](https://github.com/ai) advertised hardtack in his social networks:
1. https://twitter.com/alik0211/status/1066314099846451201/retweets
2. https://twitter.com/sitnikcode/status/1103295192654905346

Thanks to this, hardtack has [accumulated a user base](https://npm-stat.com/charts.html?package=hardtack&from=2019-02-22). I think we can return to the discussion of this PR. I will be glad to any feedback.

cc @travist

travist:
It does look more promising but we still feel that considering how risky it is to swap out "cookie" handling libraries, we would like to see more adoption from a community. I would say we will start to consider this if your library can achieve more legitimate stars and forks that this library.

travist:
Closing for now. We are happy with our current cookie library.

PR #873, opened 2018-12-18
Fix bug with day adjustment when switching months

iobuhov:
- In `monthInput.onchange` expected that dayInput.value
  and maxDate will be `Numbers` but got `String`.

- `Dat.getInputValue` fixed.

- Day validation fixed.

### DIFF
--- a/src/components/day/Day.spec.js
+++ b/src/components/day/Day.spec.js
@@ -4,7 +4,8 @@ import Harness from '../../../test/harness';
 import DayComponent from './Day';
 
 import {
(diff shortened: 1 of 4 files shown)


src/components/day/Day.spec.js
```diff
@@ -96,4 +97,51 @@ describe('Day Component', () => {
       done();
     });
   });
+
+  it('Should keep day value when switching months', (done) => {
```
travist: 👏

Issue #835, opened 2018-12-04
"valueProperty" of select component in form builder 


kran:
I found the "valueProperty" setting of select component in examples, but can't find it in builder.
how to set it in form builder?

randallknutson:
This field is currently missing from the form builder. I'll open a ticket to add it.

AnasHijazi:
@randallknutson  you open it ?

abarrak:
I think the issue caused by the definition duplication of `valueProperty` for: [`Resource` as `select` control](https://github.com/formio/formio.js/blob/master/src/components/select/editForm/Select.edit.data.js#L128-L162), and for [others `Url`/`Json` as `textfield` control](https://github.com/formio/formio.js/blob/master/src/components/select/editForm/Select.edit.data.js#L175-L192).

Specifically, the `conditional` of both occurs and prevent appearance of them.

One solution is to <b>differentiate the key</b>.
It solves the problem, but I think it requires refactor on `valueProperty` handling for one of the dependents.

<img width="461" alt="1" src="https://user-images.githubusercontent.com/20784990/50347975-8e053180-0548-11e9-83fd-a1b2a1f4d0c4.png">
<img width="475" alt="2" src="https://user-images.githubusercontent.com/20784990/50347976-8e9dc800-0548-11e9-88e2-4d9b82067238.png">

jhen1422:
This issue has been resolved

PR #884, opened 2018-12-21
Add option to skip merging in unifyComponents

iobuhov:
Add new `skipMerge` option, that mark component as "unmergeable" and when `unifyComponents` face with such component, it get ignored.


abarrak:
Great @iobuhov. 
Does it identify the origin of `valueProperty` builder bug, and solve #835  issue?

iobuhov:
@abarrak Yes, Select component has more then one field that configure `valueProperty`, but because of merging this field was hidden. Now when you select `Raw JSON` or `URL` as `Data Source`, textfield should appear below where you can set `valueProperty`.

PR #890, opened 2018-12-26
Removing CSS white color by default

edwinanciani:


travist:
Looks good, but I have a feeling this was added for a reason which we will find out later.

PR #891, opened 2018-12-27
 SELECT: Added option in Data tab to specify custom options that will be used in the Choices JS library

Aiso786:
This is useful if you want to override some default options without having to change the code. For example, the select's search result is limited to 4 items.

You can do this to override this default value:

image


travist:
Thanks for the contribution!

PR #896, opened 2018-12-28
cut reference for values stored in cache

kamilogerto2:
## Problem

_There are some problems with references when formio library getting values from cache. Let's assume scenario - in first view you use form in read and view only mode but after click on edit button you render the same form but without read only flags. What is unexpected that submit button is hidden. Why? Because in the first view - formio library hide submit button due to read only mode. It will be proper if not the fact that formio library works on reference so it change original value. When we go to the second view - hidden flag is set because we get this changed definition from cache._

## Solution

_In proposed solution We assume that cache should work as immutable object so when we get or put values we need to use cloneDeep to remove linking reference. Without that - in app we will be using still original value due to reference._

Of course I am open to another solution which should resolve the problem.

### DIFF
--- a/src/Formio.js
+++ b/src/Formio.js
@@ -820,6 +820,11 @@ export default class Formio {
           return result;
         }
 
+        // Cache the response.
+        if (method === 'GET') {
+          Formio.cache[cacheKey] = _cloneDeep(result);
+        }
+
         let resultCopy = {};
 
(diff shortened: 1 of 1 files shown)


travist:
This looks good to me. We will do some tests on our end and pull this into the next release if it passes tests.

Thanks!

Issue #887, opened 2018-12-25
(file component) upload url modified


kran:
`/api?r=upload` becames `/api?r=upload?baseUrl=xxx&project=xxx`
notice the second `?`.

kran:
@travist 
in form builder
1. add or edit an file component
2. select tab "file"
3. select "url" option for "storage"
4. input "url" field: `/api?r=upload/image'
5. render the schema, and upload a file to test, the url request will be appended by `?baseUrl=....`

randallknutson:
Sounds like we aren't checking for a ? in the url option. Is this something you could fix and add a pull request for or would you like us to add it to our backlog to fix?

Looks like the code is here: https://github.com/formio/formio.js/blob/72351633a07112c2b0ddcbc5dad97bbce7cb1833/src/providers/storage/url.js#L48-L54

kran:
@randallknutson yes, I can fix it. I'll checkout the code base to see how formio works.

PR #897, opened 2018-12-29
fix: check if ? exists in url

kran:
#887


travist:
Looks good! Thanks... This will be included in the next build.

PR #900, opened 2019-01-03
NUMBER - Bug Fix: couldn't set empty value

Aiso786:
If i clear a Number field's value, the new value isn't sent to the server.
This pull request resolves this issue.

### DIFF
--- a/src/components/number/Number.js
+++ b/src/components/number/Number.js
@@ -112,7 +112,7 @@ export default class NumberComponent extends BaseComponent {
     const val = this.inputs[index].value;
 
     if (!val) {
-      return undefined;
+      return '';
     }
 
     return this.parseNumber(val);


travist:
This was fixed before...
 - https://github.com/formio/formio.js/commit/9af675b1f1b59f7e71afda26b7847913523f95f0

Is there a way for you to better describe why returning "undefined" is a problem?

Aiso786:
When we return "undefined", the related property is missing in the "submission" object . I think the property should still figure in the object with a null or undefined value.

However, if you guys don't think like that, it's better to reject this pull request.

And i think returning `''` is not a good idea.

travist:
Yeah, that is what I am thinking as well. A number should never return an empty string. I am fine with either ```null``` or ```undefined``` so I am thinking that this issue that you are talking about may be caused elsewhere.

Aiso786:
Alright, i close this PR and will look for another way to fix the issue.
Thanks @travist

PR #909, opened 2019-01-08
FOR-1970: Fixed metadata missing in nested child form

Alena-Levina-Softensity:


src/Webform.js
```diff
@@ -1139,11 +1139,9 @@ export default class Webform extends NestedComponent {
         });
       }
 
-      const submission = _.cloneDeep(this.submission || {});
-
       // Add in metadata about client submitting the form
-      submission.metadata = submission.metadata || {};
-      _.defaults(submission.metadata, {
+      this.submission.metadata = this.submission.metadata || {};
```
travist: ```this.submission``` is a getter method so this is equivalent to saying.

```
this.getSubmission().metadata = this.getSubmission().metadata || {};
_.defaults(this.getSubmission().metadata, {....});
```

Because of this, I don't see how this is working since the metadata isn't getting set on an object that persists. What was wrong with the previous version?

Alena-Levina-Softensity: `this.getValue()` for Webform returns `this._submission` which as I see is an object that persists. 

The previous version made a copy of `this.submission` and then added `metadata` to it, which broke reference to it for parent form. As a result form could only have metadata if it is root form, not when it's a child nested form for some other form

travist: Ok, sounds good.

PR #917, opened 2019-01-10
Update Checkbox.unit.js

wag110894:


src/components/checkbox/Checkbox.unit.js
```diff
@@ -13,4 +13,52 @@ describe('Checkbox Unit Tests', () => {
     const checkbox = new CheckBoxComponent();
     assert.equal(checkbox.defaultValue, false);
   });
+  it('Change the label of the checkbox',() => {
+  const checkbox = new CheckBoxComponent({
+checkbox: [
```
travist: This is not actually testing anything other than that the CheckBoxComponent will not crash if you provide it bogus JSON as an input into the constructor. You need to actually create a legit checkbox component by providing actual checkbox JSON, and then make some assertions... such as the following.

```js
const checkbox = new CheckBoxComponent({
  type: 'checkbox',
  label: 'Label A',
  key: 'checkbox'
});
assert.equal(checkbox.labelElement.textContent, 'Label A');
```

wag110894: @travist Trying to use textContent.
TypeError: Cannot read property 'textContent' of undefined.
What can I use instead?

PR #926, opened 2019-01-11
Add breadcrumb settings for Panel

iobuhov:


travist:
Excellent work @iobuhov

PR #928, opened 2019-01-11
FOR-1077: Added reCAPTCHA component

Alena-Levina-Softensity:


src/Webform.js
```diff
@@ -1294,6 +1295,16 @@ export default class Webform extends NestedComponent {
       return console.warn('You should add a URL to this button.');
     }
   }
+
+  triggerRecaptcha() {
+    const recaptchaComponent = this.root.components.find((component) => {
```
travist: This will only search in the "root" scope of the form. You should use ```getComponent``` instead which is a method exposed to the nested components.

src/components/recaptcha/editForm/ReCaptcha.edit.display.js
```diff
@@ -0,0 +1,103 @@
...
+    label: 'Type of event',
+    tooltip: 'Specify type of event that this reCAPTCHA would react to',
+    type: 'radio',
+    values: [
+      {
+        label: 'Form Load',
+        value: 'formLoad'
+      },
+      {
+        label: 'Button Click',
```
travist: If this is attached to the Submit button, this component will need to block the submission until it finishes the recaptcha check and then allow the submission to be completed.

src/Webform.js
Alena-Levina-Softensity: @travist Thanks! Fixed it by using ```everyComponent``` since I need more complex search than just by key (which is what ```getComponent``` does as I see)

src/components/recaptcha/editForm/ReCaptcha.edit.display.js
Alena-Levina-Softensity: @travist fixed

PR #945, opened 2019-01-17
Bugfix for number component large numbers handling.

iobuhov:
This PR is actually a proposal to alter the behaviour of
NumberComponent. This update changes how `setValue` works, now if you
try to set number greater then `MAX_SAFE_NUM` it would be ignored and
insted `MAX_SAFE_NUM` value would be assigned to component value (e.g.
4299299299 -> 1000000000). The same logic will be applied to
`MIN_SAFE_NUM`.

Pros:
- we will have stable and predictable numbers.
- only correct numbers can be assigned to number component.

Cons:
- user might be confused in case when his value would be silently
   converted to `MIN_SAFE_NUM` (-1000000000)
   or `MAX_SAFE_NUM` (1000000000)

Another approach is to add validators to NumberComponent.

Pros:
  - don't affect user experience

Cons:
  - still possible to set incorrect value
  - still possible to loose precision when parsing value


travist:
Closing due to inactivity.

PR #1000, opened 2019-02-05
fix(FOR-1742): Initial focus wasn't working, Focus Event wasn't triggering

edwinanciani:


travist:
This seems like a hack. Is there not a better way (maybe using events) to know if these inputs are ready to receive focus events? This also would probably break with older browsers or devices not able to process the forms as quickly. Please see if you can come up with a better way of doing this.

PR #1004, opened 2019-02-07
Validation discussion

iobuhov:
I would like to have some comments on my change on `Webform.onChange` handler as well as some question about button `set disable`.

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -1078,7 +1078,12 @@ export default class Webform extends NestedComponent {
     super.onChange(flags, true);
     const value = _.clone(this._submission);
     value.changed = changed;
-    value.isValid = this.checkData(value.data, flags, changed ? changed.instance : null);
+    if (!this.initialized) {
+      value.isValid = this.isValid(value.data, true);
+    }
+    else {
+      value.isValid = this.checkData(value.data, flags, changed ? changed.instance : null);
+    }
     this.showElement(true);
     this.loading = false;
     if (this.submitted) {


iobuhov:
About this setter:
```
  set disabled(disabled) {
    // Do not allow a component to be disabled if it should be always...
    if ((!disabled && this.shouldDisable) || (disabled && !this.shouldDisable)) {
      return;
    }
    super.disabled = disabled;
    this.setDisabled(this.buttonElement, disabled);
  }
```

I'm consorned about this expression:
```
    if ((!disabled && this.shouldDisable) || (disabled && !this.shouldDisable)) {
      return;
    }
```
I know that this should resolve button `disabled state` inconsistency, but in usual usage, this expression prevent button disabling because `shouldDisable` getter is inherited from `Base` class and by default always resolves to `false`, that means that when `disabled` argument has value true we never can reach `super.disabled`. If I understand correctly, both `disabled` flag and `shouldDisable` property should reflect state changes and both should return the same value (false or true). But right now `shouldDisable` property constantly returns false.
My solution either remove this expression or override `shouldDisable` getter in `Button` class. Any advice would be helpful.

travist:
```shouldDisable``` does not always resolve to false. There is logic there to protect a component that should never disable to not get disabled when "disabled = true" is called. This logic helps protect that, so I am not sure I agree with your assessment since looking at the code, it appears to be necessary. I would prefer to just override the ```shouldDisable``` method for the button class that is consistent with the validity of the form.

travist:
I had to revert this pull request. The tests are failing.

PR #1005, opened 2019-02-07
Feature/for 2039 select continious scrolling

VladTsygankov:


travist:
Very well done @VladTsygankov

src/components/select/editForm/Select.edit.data.js
```diff
@@ -271,12 +271,30 @@ export default [
...
+    type: 'textfield',
+    input: true,
+    key: 'sort',
+    label: 'Sort Query',
+    weight: 18,
+    description: 'The sort query for results',
+    tooltip: 'User this to provide additional sorting using query parameters',
+    conditional: {
+      json: {
+        or: [
```
YuryRybakSoftensity: Please consider usage of 'in' operator here http://jsonlogic.com/operations.html#in

travist: This can be resolved in another PR. going to accept it as is.

travist:
There is an issue with this change. If you are loading an API that does not support pagination, it will continually load the same page over and over and the list gets bigger and bigger..

travist:
I spent some time to investigate. I also found quite a few other issues with infinite scroll. Please review these changes.

https://github.com/formio/formio.js/commit/cecabd3047013eac1ef48be8da5fd61b6f1c86a9

PR #1024, opened 2019-02-15
Allow custom components to control the Fuse search behavior

noordawod:
This allows custom components to control the Fuse search behavior, and not only the search threshold, and it's backward-compatible.


travist:
Looks good, thanks!

PR #1029, opened 2019-02-19
Convert templates to handlebars templates

iobuhov:


iobuhov:
Reopen PR, it seems that because my first PR was against `master` and then switched to `4.x` comparison not recalculated and there were conflicts. Now it's ok.

PR #1031, opened 2019-02-20
Added option to use an array of refreshOn attributes for custom form components

RahulRavi1604:
Recently I faced an issue where I need to create a select component that must refresh based on 2 other inputs. So I modified the base Component to overcome this. Feel free to correct me if I am wrong.


HarishGangadharan:
Even i too faced the same issues.Is it a feasible solution to move on?

travist:
Thanks!

PR #1042, opened 2019-02-25
Allow combining "multiple" and "Add Another" options.

noordawod:
You can see a similar behavior for regular TEXTFIELD box, but unfortunately this behavior isn't found for SELECT field. This patch will add support for adding an "Add Another" button which can be combined with MULTIPLE to have almost the same effect.


noordawod:
Hold on with this for a while, I am submitting more commits to make this even better.

Issue #1058, opened 2019-03-06
URL parsing error in Select Component after 3.14.1


airarrazaval:
##Description

Since formiojs@3.14.1, Select component fails to load items  when using URL as `dataSrc`.  When no query is defined, but filter is defined, no `?` character is included to separate the url from URL query parameters, and instead a `&` character is used.

This bug was introduced in https://github.com/formio/formio.js/commit/cecabd3047013eac1ef48be8da5fd61b6f1c86a9#diff-d63440e917cdc7944ac6e07a5be846c0 and is located at the `loadItems` function in Select component:

https://github.com/formio/formio.js/blob/cecabd3047013eac1ef48be8da5fd61b6f1c86a9/src/components/select/Select.js#L399-L407

## Expected Behaviour

For example, if the following component is defined:

```json
{
  "type": "select",
  "dataSrc": "url",
  "data": {
    "url": "https://example.com"
  },
  "filter": "data.prop__in=1,5"
}
```

The expected parsed URL when loading items should be:

- `https://example.com?data.prop__in=1,5`

## Current Behaviour

Instead, you get the following URL:

- `https://example.com&data.prop__in=1,5`

## Solution

Replace 

https://github.com/formio/formio.js/blob/cecabd3047013eac1ef48be8da5fd61b6f1c86a9/src/components/select/Select.js#L406

With

```js
  url += (!url.includes('?') ? '?' : '&') + this.interpolate(this.component.filter);
```

airarrazaval:
@travist @randallknutson pull request 1059 created

PR #1059, opened 2019-03-06
fixed select component url parsing

airarrazaval:
Hotfix for https://github.com/formio/formio.js/issues/1058


travist:
Looks good. Thanks!

PR #1074, opened 2019-03-13
Number - fixed Hide Input

VladTsygankov:


src/components/number/Number.js
```diff
@@ -102,7 +102,7 @@ export default class NumberComponent extends Input {
 
   get inputInfo() {
     const info = super.inputInfo;
-    info.attr.type = 'text';
+    info.attr.type = this.component.mask ? 'password' : 'text';
```
travist: The mask is different from a "password" field. The mask is like what you see with a phone number where you see ```(___) ___ - ____``` but when you type you see the numbers. By turning this into a password field, you would not see the input.

randallknutson:
Merged separately.

PR #1094, opened 2019-03-19
Fix empty required select is valid onChange

AlexeyDavyd:


travist:
I am having to revert this commit because of this.

Alt text

Issue #1085, opened 2019-03-15
[DRAFT PATCH] Submit button should be disabled when file uploads are in progress


franke-mc:
As large file uploads take time, form submissions probably should be prevented when file uploads are still in progress.

Here is a partial solution that prevents accidental mouse clicks on a submit button in this case:
```
diff --git a/src/components/file/File.js b/src/components/file/File.js
index ee1400bb..86b180ed 100644
--- a/src/components/file/File.js
+++ b/src/components/file/File.js
@@ -741,6 +741,42 @@ export default class FileComponent extends BaseComponent {
         this.uploadStatusList.appendChild(uploadStatus);
 
         if (fileUpload.status !== 'error') {
+          // Track uploads in progress.
+          if (fileService.uploadsInProgress === undefined) {
+            const cssClass = 'uploads-in-progress';
+            var submitButton = this.root.element
+                .querySelector('.formio-component-submit')
+                .querySelector('button');
+            var array = new Array();
+            fileService.uploadsInProgress = {
+              array: array,
+              add: function(item) {
+                if (cssClass && (array.length === 0)) {
+                  submitButton.classList.add(cssClass);
+                }
+                array.push(item);
+              },
+              remove: function(item) {
+                array.splice(array.indexOf(item), 1);
+                if (cssClass && (array.length === 0)) {
+                  submitButton.classList.remove(cssClass);
+                }
+              },
+              report: function() {
+                const n = array.length;
+                var msg = `Uploads in progress: ${n}`;
+                if (n>0) {
+                  const keys = array.map(x => x.component.key);
+                  msg += ` (${keys.join(', ')})`;
+                }
+                console.log(msg);
+              }
+            };
+          }
+          const uploadsInProgress = fileService.uploadsInProgress;
+          uploadsInProgress.add(this);
+          //uploadsInProgress.report();
+          //
           if (this.component.privateDownload) {
             file.private = true;
           }
@@ -762,6 +798,10 @@ export default class FileComponent extends BaseComponent {
               fileInfo.originalName = file.name;
               this.dataValue.push(fileInfo);
               this.refreshDOM();
+              // Track uploads in progress.
+              uploadsInProgress.remove(this);
+              //uploadsInProgress.report();
+              //
               this.triggerChange();
             })
             .catch(response => {
```
Corresponding CSS for disabling the accidental click on submit:
```
+div.formio-component-submit button.uploads-in-progress {
+    pointer-events: none !important;
+    cursor: not-allowed !important;
[...]

Issue #951, opened 2019-01-18
disableOnInvalid is not working for button component


sriram275:
I set checkbox for disableOnInvalid , So I am expecting the submit button should be disabled if the form is invalid( means if there is any required field not filled). But it is enabling always and once after I click it is showing me error instead have to disable the click if form is invalid.

travist:
We are aware of this issue and will have a developer look into it asap. Thanks.

travist:
Should be fixed in latest build.

PR #1096, opened 2019-03-19
#1085 Disable submit button while file uploads in progress.

franke-mc:
Pull request for issue #1085 .

> As large file uploads take time, form submissions probably should be prevented when file uploads are still in progress.
> 
> Here is a partial solution that prevents accidental mouse clicks on a submit button in this case,
> [and the] corresponding CSS for disabling the accidental click on submit.

> As noted e.g. [here](https://bitsofco.de/theres-no-reason-to-use-pointer-events-for-html-elements/#disablinginteractiveelements), to prevent activating the submit button via keyboard, it needs to be properly disabled, so instead of using a CSS class and styling this should actually set disabled="disabled" on the button (and find a way to return to the original state once all uploads have finished). Probably best to integrate with other conditions causing the submit button to become disabled. Maybe combine with fix for issue #951 .

### DIFF
--- a/src/components/file/File.js
+++ b/src/components/file/File.js
@@ -783,6 +783,42 @@ export default class FileComponent extends BaseComponent {
         this.uploadStatusList.appendChild(uploadStatus);
 
         if (fileUpload.status !== 'error') {
+          // Track uploads in progress.
+          if (fileService.uploadsInProgress === undefined) {
+            const cssClass = 'uploads-in-progress';
+            var submitButton = this.root.element
+                .querySelector('.formio-component-submit')
+                .querySelector('button');
+            var array = new Array();
+            fileService.uploadsInProgress = {
+              array: array,
+              add: function(item) {
+                if (cssClass && (array.length === 0)) {
+                  submitButton.classList.add(cssClass);
+                }
+                array.push(item);
+              },
+              remove: function(item) {
+                array.splice(array.indexOf(item), 1);
+                if (cssClass && (array.length === 0)) {
+                  submitButton.classList.remove(cssClass);
+                }
+              },
+              report: function() {
+                const n = array.length;
+                var msg = `Uploads in progress: ${n}`;
+                if (n>0) {
+                  const keys = array.map(x => x.component.key);
+                  msg += ` (${keys.join(', ')})`;
+                }
+                console.log(msg);
+              }
+            };
+          }
+          const uploadsInProgress = fileService.uploadsInProgress;
+          uploadsInProgress.add(this);
+          //uploadsInProgress.report();
+          //
           if (this.component.privateDownload) {
             file.private = true;
           }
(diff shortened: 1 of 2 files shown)


travist:
Thanks for your contribution! I like this a lot. Give us some time to test this internally and we will pull it through if that goes well.

src/components/file/File.js
```diff
@@ -783,6 +783,42 @@ export default class FileComponent extends BaseComponent {
...
+                if (n>0) {
+                  const keys = array.map(x => x.component.key);
+                  msg += ` (${keys.join(', ')})`;
+                }
+                console.log(msg);
+              }
+            };
+          }
+          const uploadsInProgress = fileService.uploadsInProgress;
+          uploadsInProgress.add(this);
```
travist: what if ```uploadsInProgress``` is null here?

src/components/file/File.js
```diff
@@ -808,6 +844,10 @@ export default class FileComponent extends BaseComponent {
               }
               files.push(fileInfo);
               this.setValue(this.dataValue);
+              // Track uploads in progress.
+              uploadsInProgress.remove(this);
```
travist: what if ```uploadsInProgress``` is null here?

src/components/file/File.js
franke-mc: Shouldn't happen, at least I can't see how.

In the patch above, it is created before (first) use, and then never cleared.
if (fileService.uploadsInProgress === undefined) {
            fileService.uploadsInProgress = {...}
}

As it works as intended, I can't see how it could be null.

I'd be more worried about multiple forms being displayed simultaneously on the same page interfering with one another; that's something I haven't tried yet.

Maybe the initialization of uploadsInProgress should happen together with fileService, but I'm not familiar enough with those portions of source code. Having it all in one place seemed simpler for now, for me at least.

src/components/file/File.js
franke-mc: Shouldn't happen. (Same reason as above.)

src/components/file/File.js
travist: ok, I see that now.

src/components/file/File.js
```diff
@@ -783,6 +783,42 @@ export default class FileComponent extends BaseComponent {
         this.uploadStatusList.appendChild(uploadStatus);
 
         if (fileUpload.status !== 'error') {
+          // Track uploads in progress.
+          if (fileService.uploadsInProgress === undefined) {
+            const cssClass = 'uploads-in-progress';
+            var submitButton = this.root.element
+                .querySelector('.formio-component-submit')
```
noman115pk: @franke-mc @travist This button does not exist in Wizard forms and giving null for next querySelector so file uploading is not working in wizard forms.

franke-mc: Sorry for that. Could you please try https://github.com/formio/formio.js/pull/1309 ?

PR #1111, opened 2019-03-25
Implemented Edit JSON button in builder

Alena-Levina-Softensity:


src/WebformBuilder.js
```diff
@@ -118,7 +128,13 @@ export default class WebformBuilder extends Webform {
         // Add the edit buttons to the component.
         comp.prepend(this.ce('div', {
           class: 'component-btn-group'
-        }, [removeButton, copyButton, pasteButton, editButton]));
+        }, [
+          removeButton,
+          copyButton,
+          pasteButton,
+          this.options.enableEditJsonButton ? editJsonButton : null,
```
travist: Can we call this option ```jsonEdit```?  I don't typically like ```Json``` within the name of a config even though it is technically considered valid CamelCase.... I always confuse it with ```JSON``` like ```editEditJSONButton``` or something like that.  ```jsonEdit``` looks cleaner.

Alena-Levina-Softensity: @travist I usually tend to prefer more descriptive names, though now I think that 'Button' part might be excess since it might be implemented with different UI. I also wanted to add to the name somehow that it's setting for the builder, but at the moment didn't find a way for it. 

I like the shortness of `jsonEdit`, but I'm a bit concerned that by the meaning it can be related to anything (not to the builder at all). What are your thoughts on that? What do you think about something like `enableBuilderJsonEdit`? 

Also I'm not sure I got the idea about difference between `json`, `Json` and `JSON` for you, could you please explain it in a bit more detail?

travist: It isn't a big issue for me, but I tend to not like labels where an Acronym is a non-first word in a camel case name. The reason for this is I always need to refer to the documentation to determine what is correct. For example, I always get ```innerHTML``` wrong since the "correct" camel case way of saying it is ```innerHtml``` which actually looks weird in my opinion. This is why if I have an acronym, I prefer it at the beginning since it is less ambiguous. If we have a new property called ```enableBuilderJsonEdit```, I will forever be referring to source code / docs to figure out how to spell it right and in what order... I will constantly be asking myself "was it ```enableBuilderJSONEdit```, ```enableBuilderJsonEdit```, ```enableJsonBuilderEdit```????" , whereas ```jsonEdit``` I will remember without referring to anything since JSON as the first word in camel case will ALWAYS be ```json```. Maybe we take ```json``` out of it.   What about ```componentEditor```?

randallknutson: jsonEdit is worded more the way that "Yoda" would say it. To me editJson or editJSON would be most descriptive of what the button does. Because it is already editJson I'd recommend just leaving it as is for now instead of possibly changing it to editJSON.

travist: I guess I am alright with ```editJSON```... but that brings me back to my original issue. I don't like acronyms as non-first words within camel-cased options. It's ambiguous. I will forever need to refer to the source code to know if it is either ```editJSON``` or ```editJson```.

Alena-Levina-Softensity: @randallknutson @travist putting discussion of the choice between ```editJson``` or ```jsonEdit``` aside for a moment, I'm not sure I understood what are your arguments against ```enableBuilderJsonEdit``` , could you please clarify?

travist: It's too long, has an ambiguous ```Json``` vs ```JSON```, and is "Yoda" speak. It also sounds like a bunch of words just put together instead of providing a clear indication of what it does. I know our suggestions have that same problem, so maybe we just stop the bikeshedding and commit it. @randallknutson thoughts?

randallknutson: Since these are options for builder already, having that as part of the option is extraneous. enableEditJSON(Button)? would be good...

Or... 

```
enableButtons = {
  remove: true,
  copy: true,
  paste: true,
  editJSON: true,
  remove: true,
}
```

Alena-Levina-Softensity: I really like the 'Or' version! @travist ?

travist: :+1:

travist:
@Alena-Levina-Softensity are you going to make that change? I will get a new build going once you do.

Alena-Levina-Softensity:
@travist yes, I just didn't get a moment to work on it due to other prioritized tasks

PR #1130, opened 2019-04-03
Fix offline draft sumbissions

AlexeyDavyd:


src/components/button/Button.js
```diff
@@ -215,7 +215,7 @@ export default class ButtonComponent extends BaseComponent {
           event.preventDefault();
           event.stopPropagation();
           this.emit('submitButton', {
-            state: this.component.state || 'submitted'
+            state: this.component.action === 'submit' ? 'submitted' : 'draft'
```
randallknutson: This change is incorrect. The default state should be 'submitted' unless they specifically set it to draft. Your testing form may not have been configured properly. Set it to "Save in State" and State of "draft" to test.

src/Formio.js
```diff
@@ -232,6 +232,15 @@ export default class Formio {
   load(type, query, opts) {
     const _id = `${type}Id`;
     const _url = `${type}Url`;
+
+    let url = '';
+    if (this[_url].indexOf(this.formId) !== this[_url].lastIndexOf(this.formId) && type === 'submission') {
```
randallknutson: Can you explain why this change is necessary?

randallknutson:
These changes still aren't necessary. You can use the "state" field in the button settings to set the state.

PR #1136, opened 2019-04-04
Update Checkbox to add checkbox-checked class like Radio adds radio-selected class

cymen:
I'm making use of the `radio-selected` class to do some styling and I'd like to do the same thing for checkboxes however I noticed that no class is set on the component/wrapper div (in the checkbox case, the div with class formio-component-checkbox) when the checkbox is checked.

I did these things:
- [x] added test to Radio for it setting `radio-selected`
- [x] added test to Checkbox for setting `checkbox-checked`
- [x] updated Checkbox to set `checkbox-checked` when the checkbox is checked
- [x] per linting rules, added some whitespace to Checkbox to get linting to pass
- [x] add tests to Checkbox and Radio for clicking them on and off again (that classes change as expected)

### DIFF
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -250,19 +250,21 @@ export default class CheckBoxComponent extends BaseComponent {
       this.input.value = 0;
       this.input.checked = 0;
     }
+
     if (this.input.checked) {
       this.input.setAttribute('checked', true);
     }
     else {
       this.input.removeAttribute('checked');
     }
+
     return value;
   }
 
   setValue(value, flags) {
     flags = this.getFlags.apply(this, arguments);
     if (this.setCheckedState(value) !== undefined) {
-      return this.updateValue(flags);
+      return this.updateValue(flags, value);
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/components/checkbox/Checkbox.js
```diff
@@ -250,19 +250,21 @@ export default class CheckBoxComponent extends BaseComponent {
...
     }
+
     return value;
   }
 
   setValue(value, flags) {
     flags = this.getFlags.apply(this, arguments);
     if (this.setCheckedState(value) !== undefined) {
-      return this.updateValue(flags);
+      return this.updateValue(flags, value);
```
cymen: Not sure if `value` needs to be passed (following same pattern as `Radio`).

src/components/checkbox/Checkbox.js
```diff
@@ -250,19 +250,21 @@ export default class CheckBoxComponent extends BaseComponent {
       this.input.value = 0;
       this.input.checked = 0;
     }
+
```
cymen: Linting

src/components/checkbox/Checkbox.js
```diff
@@ -250,19 +250,21 @@ export default class CheckBoxComponent extends BaseComponent {
...
     if (this.input.checked) {
       this.input.setAttribute('checked', true);
     }
     else {
       this.input.removeAttribute('checked');
     }
+
```
cymen: Linting

travist:
This is great! And thank you for writing tests!. 

Give us some time to review this internally and we will pull this in as soon as possible. Thanks for your contributions.

PR #1153, opened 2019-04-09
Fixed missing value when adding rows for textare component with multi…

VladTsygankov:
…ple value


src/components/textarea/TextArea.js
```diff
@@ -303,7 +303,8 @@ export default class TextAreaComponent extends TextFieldComponent {
         return;
       }
       else {
-        return super.setValue(this.setConvertedValue(value), flags);
+        value = this.component.multiple ? this.getValue() : this.setConvertedValue(value);
```
travist: I would think this would still need to set the converted value if multiple... the way you check this is change the editor to save as JSON and then make it multiple. I bet that still breaks here...

PR #1155, opened 2019-04-09
Feature added: before next hook

AlexeyDavyd:


src/Wizard.js
```diff
@@ -120,6 +120,13 @@ export default class Wizard extends Webform {
     }));
   }
 
+  beforeNext() {
+    if (this.options.hooks && this.options.hooks.hasOwnProperty('beforeNext')) {
+      return this.hook('beforeNext', this.currentPage, this.submission);
```
randallknutson: This will need to wait for a callback from the hook (you can use the last parameter of the hook for a callback) and then call super.beforeNext(). We can't short circuit the calls.

src/Wizard.js
```diff
@@ -120,6 +120,19 @@ export default class Wizard extends Webform {
     }));
   }
 
+  beforeNext() {
+    return new Promise((resolve, reject) => {
+      this.hook('beforeNext', this.currentPage, this.submission, (err) => {
+        console.log('leeel');
```
travist: why?

PR #1157, opened 2019-04-09
Fixed wrong character counter in textarea component with editor

VladTsygankov:


src/components/textfield/TextField.js
```diff
@@ -135,7 +135,8 @@ export default class TextFieldComponent extends BaseComponent {
       this.setCounter('words', this.wordCount, this.dataValue.trim().split(/\s+/).length, this.maxWordCount);
     }
     if (this.charCount) {
-      this.setCounter('characters', this.charCount, this.dataValue.length, this.maxCharCount);
+      const value = this.component.wysiwyg ? this.removeTagsFromString(this.dataValue) : this.dataValue;
```
travist: This will crash if you configure just a TextField with character counts because the TextField will show undefined for this method.

PR #1164, opened 2019-04-11
Added validation for min & max cheched for selectboxes

AlexeyDavyd:


src/components/selectboxes/SelectBoxes.js
```diff
@@ -110,6 +112,41 @@ export default class SelectBoxesComponent extends RadioComponent {
     this.updateValue(flags);
   }
 
+  onChange(flags, fromRoot) {
```
travist: This code isn't needed... the validator should get triggered on change already and all the logic exists in the validators and should not be duplicated here.

PR #1166, opened 2019-04-11
update Text field component to link Label ("for" attribute) with input's id

phucduong86:


src/components/textfield/TextField.js
```diff
@@ -43,7 +43,8 @@ export default class TextFieldComponent extends BaseComponent {
   elementInfo() {
     const info = super.elementInfo();
     info.type = 'input';
-
+    info.attr.id = "".concat(this.component.key);
```
randallknutson: Is there a reason you are doing a concat with an empty string? If you want ensure it is a string you could use this.component.key.toString() although it should already be a string.

phucduong86: No it works without the concat. 
I saw this done somewhere else in the code base and just thought there was a reason for it.

phucduong86:
Thanks Randall I just updated the line.

travist:
Thanks for the contribution @phucduong86

PR #1171, opened 2019-04-12
added id value for Number and Select component

phucduong86:
Similar change to the TextField Component - improve Accessibility by linking the label "for" attribute with the input's "id".


travist:
These were added to the 3.20.0-beta.1 release. Thanks for your contribution!

PR #1172, opened 2019-04-14
added id to elementInfo

moshefi:
Added id attribute to the input element info to enable the <label for=''> relation.
Is is vital for accessibility


travist:
I will just make the change manually since we typically don't put the build files in a pull request.

travist:
Please verify if this will work for you.

https://github.com/formio/formio.js/pull/1173

Going to pull this in the next build.

moshefi:
Works perfect !

moshefi:
I merged with upstream and I see now that it was already handled.

 

From: moshe.fishman@gmail.com <moshe.fishman@gmail.com> 
Sent: Monday, April 15, 2019 9:45 AM
To: 'formio/formio.js'‏ <reply@reply.github.com>
Subject: RE: [formio/formio.js] added id to elementInfo (#1172)

 

Works perfect !

PR #1191, opened 2019-04-19
include additional texts under translation

kapilratnani:
Thanks for this awesome! extremely customisable library. In our use-case we want to present users with form-builder too and hence would want some additional texts to be included under translation.


kapilratnani:
@travist Thanks

PR #1193, opened 2019-04-19
Fixing Okta SSO get AccessToken error

austindbertola:
the OktaAuth TokenManager get() will return a promise. Without this, the subsequent call will always have the accessToken as being undefined. It is also unnecessary to create a new OktaAuth object at this point. Just use the one that is provided.

### DIFF
--- a/src/Formio.js
+++ b/src/Formio.js
@@ -1211,29 +1211,31 @@ export default class Formio {
     return new Promise((resolve, reject) => {
       const Okta = options.OktaAuth;
       delete options.OktaAuth;
-      var authClient = new Okta(options);
-      var accessToken = authClient.tokenManager.get('accessToken');
-      if (accessToken) {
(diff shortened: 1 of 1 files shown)


travist:
Looks good. Thanks!

Alena-Levina:
@austindbertola I had to revert your removal of OktaAuth constructor call in this PR: https://github.com/formio/formio.js/pull/2182

As I understood the changes you did in your PR where done in order to make formio.js work with https://github.com/okta/okta-auth-js v.2 (while our previous implementation was working with v.1) -  this explains why you changed `authClient.tokenManager.get` to be promise based. 

But with your `var authClient = Okta;` instead of `var authClient = new Okta(options);` I get an error that `tokenManager` is not defined. I was using latest v2.11.0 release of okta-auth-js and wasn't able to get it working without OktaAuth constructor call. As a reference please see that okta-auth-js is making constructor calls in all their examples: https://github.com/okta/okta-auth-js#configuration-reference

Issue #1198, opened 2019-04-23
Progress Bar in Wizard


noman115pk:
Do we have any component like progress bar for wizard which can show user about the completion of steps of the form? Or something like CSS class to show page is visited and completed.
![image](https://user-images.githubusercontent.com/17640825/56576556-18869c00-65e2-11e9-8cbd-ac640816632b.png)

Thanks for the awesome library.

randallknutson:
We don't currently have something like that as part of our default library. You could pretty quickly add something like that by extending the Wizard component and add some CSS classes for any pages below the current page. We would love a pull request adding this functionality through CSS classes to our library.

noman115pk:
@randallknutson PR created.

Josephalexantony-aot:
It will be really nice if a progress bar component is added in form.io. Currently there is nothing like that. and if someone knows how to implement it and connect to anther component, requesting your help.

heather-formio:
This can be accomplished using custom CSS. Also, out of the box using the USWDS template with Form.io, a wizard step indicator with this kind of styling can be used. Also check out how city of San Francisco accomplished a vertical and responsive wizard step indicator here: https://github.com/SFDigitalServices/formio-sfds / https://formio-sfds.herokuapp.com/#form-address

PR #1204, opened 2019-04-24
Added Styling for Wizard page numbers

noman115pk:
#1198 
Minimalist styling of page numbers added for form wizard.


src/sass/formio.form.scss
```diff
@@ -1035,3 +1035,9 @@ body.formio-dialog-open {
 .formio-tagpad-form-title {
   font-size: 1.7em;
 }
+
+.page-item.seen .page-link {
+  color: #fff;
+  background-color: #17a2b8;
```
travist: This is pretty opinionated. If you wish to have custom styles then you can include your own CSS classes that handle these types of styles. The other change you made I am fine with, but we probably would not be able to accept this since it may conflict with other bootstrap themes.

noman115pk: @travist Updated the CSS class name from bootstrap defaults to formio customs.

travist:
Can now be done with 4.x branch.

PR #1226, opened 2019-05-03
Fixed validation of next page components

AlexeyDavyd:


src/components/_classes/nested/NestedComponent.js
```diff
@@ -517,8 +521,12 @@ export default class NestedComponent extends Field {
       this.setCustomValidity('');
       return true;
     }
+    console.log(this.isLastPage());
```
randallknutson: Please remove the console.log

AlexeyDavyd: Have updated a PR

PR #1246, opened 2019-05-10
Add automatic contrib component registration

aiwebb:


aiwebb:
Note that this will override the Stripe component's assignment of `components.stripe` with an object that houses both the Stripe and StripeCheckout components under `components.stripe.stripe` and `components.stripe.checkout` respectively.

The Stripe components will remain at `components.stripe` and `components.stripeCheckout` if they are pulled in individually instead of via the Contrib module / formio.contrib.js; I'm wary of removing those registrations since it will likely break for anyone currently pulling them in that way, but we should probably avoid doing that for new Contrib modules and just handle them in the Contrib module's dictionary.

We may also want to consider updating `registerComponent` to prefix any incoming definitions with e.g. `user` or `custom` to avoid collisions between user components and core+contrib components.

PR #1254, opened 2019-05-14
Unbreak search in resources

franke-mc:
Currently, search in resources appears to be broken because of some left-over references to `.searchFields` which should be `.searchField` (without s).  These changes appear to fix it.
(Although I should say that the purpose of the line in Select.js is not clear to me.)

Some problems still remain with these changes applied, but they appear to be necessary to start debugging.

### DIFF
--- a/src/components/resource/editForm/Resource.edit.display.js
+++ b/src/components/resource/editForm/Resource.edit.display.js
@@ -41,13 +41,25 @@ export default [
     weight: 51.6
   },
   {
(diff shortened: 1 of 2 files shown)


travist:
Thanks @franke-mc!  We will take a look at this very soon. Thanks for your contribution!

franke-mc:
This needs more work in order to allow for both Choices.js-based search and client-server-communication based search. Closing for now.

PR #1264, opened 2019-05-15
Cleanup Choices.searchFields & add config for search in Resource components

franke-mc:
Continuation of #1254.

regarding Select.js: It turns out that for Choices.searchFields to understand SelectComponent.searchField, it needs a 'value.' prefix.

regarding: Resource Config UI: This patch makes existing configuration available for both variants of search (local client-side search in Choices, and web-request-based search queries). The latter might still be buggy, e.g.
- In multiple-selection components, previously added items sometimes disappear. (And sometimes these gone-missing items show up again when the search "finds" them.)
- Once a search query has been done, even removing the entire search input via backspace key does not bring back the full list of resources to select from.

(Those bugs were observed in a version derived from 3.19.4.)

### DIFF
--- a/src/components/resource/editForm/Resource.edit.display.js
+++ b/src/components/resource/editForm/Resource.edit.display.js
@@ -41,15 +41,6 @@ export default [
     weight: 51.6
   },
   {
-    type: 'tags',
-    input: true,
-    key: 'searchFields',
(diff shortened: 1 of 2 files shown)


travist:
Looks good! Thank you and also thanks for keeping reverse compatibility.

PR #1279, opened 2019-05-21
Implemented select component autofill

AlexeyDavyd:


src/components/base/Base.js
```diff
@@ -1398,13 +1398,31 @@ export default class BaseComponent extends Component {
...
+
   /**
    * Creates a new input element.
    * @param {HTMLElement} container - The container which should hold this new input element.
    * @returns {HTMLElement} - Either the input or the group that contains the input.
    */
   createInput(container) {
     const input = this.ce(this.info.type, this.info.attr);
+    if (this.info.type === 'select') {
+      this.addAutofillHoneyInput(container, input);
```
travist: Why not extend ```createInput``` in the SelectComponent class and add this so you don't have to do this?

PR #1288, opened 2019-05-22
Fix various issues around radio-style checkboxes and checkbox change …

aiwebb:
…logic

### DIFF
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -312,9 +312,7 @@ export default class CheckBoxComponent extends BaseComponent {
       return;
     }
 
-    if (this.setCheckedState(value) !== undefined) {
-      return this.updateValue(flags, value);
-    }
+    return this.updateValue(flags, value);
   }
 
   getView(value) {
@@ -340,8 +338,13 @@ export default class CheckBoxComponent extends BaseComponent {
         value = this.getRadioGroupValue();
       }
     }
+    else if (flags && flags.modified && this.input.checked && value === undefined) {
+      value = true;
+    }
 
     const changed = super.updateValue(flags, value);
+
+    this.setCheckedState(value);
 
     return changed;
   }
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -31,11 +31,36 @@ export default class CheckBoxComponent extends BaseComponent {
   }
 
   get defaultValue() {
-    return this.component.name ? '' : (this.component.defaultValue || false).toString() === 'true';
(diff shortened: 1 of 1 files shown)


travist:
I am a little concerned about this change. From what I gather, the PDF radio issue was broken by this commit

https://github.com/formio/formio.js/commit/33d6eb9195e2523b673b1af8be5413faf535eb85#diff-32d85b53aba5403fc7a7351272e484f4R48

This seems like a lot of code that adds complexity and also performance regressions to fix a commit aiming to make it so that nested forms do not conflict. I am guessing here, but it seems that the ```this.root.id``` may be the same for a parent form with many nested forms. If this is the case, then it maybe the solution to just change this to ```this.parent.id```. I assume you have tried this?  Basically just replace anywhere you see ```this.root.id``` to say ```this.parent.id```? This would make it so that the separated radios must have the same parent, but this is pretty much always going to be the case, but would also solve the issue where a parent form with nested forms cause name collisions with the radio buttons. They may have the same ```this.root.id```, but would certainly have different ```this.parent.id```.

Let me know your thoughts. I am hesitant to commit this as is.

aiwebb:
I don't _think_ 33d6eb9 is the culprit here - the issues I've seen as part of FOR-2212 are independent of nesting. The behavior I've been seeing appears to have been caused by f82f65f, which changed the way checkbox values were updated and CSS classes applied (not that the problems were exclusively from that commit; rather, it seems to have uncovered some deeper state management problems with earlier code).

The issues addressed here are independent of nesting. The core problems were:

1. checking against `component.name` to trigger radio-checkbox behaviors, which is invalid (non-radio checkboxes can sometimes have a `component.name` value, which should be ignored); it now checks against the actual `component.inputType` via the helper accessor `isRadioCheckbox`

2. a `getValue()` implementation that was non-idempotent; it sometimes _modified_ checkbox state as a side effect

3. a `setCheckedState()` that was not being invoked directly on value change, but on value read (the `getValue()` issue above); this has been moved to occur inside `setValue()` instead

4. addition/removal of CSS classes being desynced from actual checked-state changes; resolved by moving them to `setCheckedState()` instead of `updateValue()`

For reference, I've been testing against [this form JSON](https://jsonblob.com/api/jsonBlob/895502eb-7d0c-11e9-afa0-cb18e4e53147) for issues with overall checkbox state management (no issues found with Radio or SelectBox) and a simple two-checkbox PDF.

I haven't done any performance analysis against these changes, but it *should* clear up some of the infinite loops that were occurring due to the previous get/set conflation - if it's substantially worse, I've probably missed something. Can you let me know what you're seeing / what the best way is to go about testing performance differences?

travist:
That sounds good. I was thinking that you were fixing what was broken by that previous commit so it seemed a little extreme. The tests are failing so those will need to be resolved before we can commit this.

Issue #1249, opened 2019-05-13
Tabs content validation errors are not displayed on submit


Sipaha:
https://github.com/formio/formio.js/blob/master/src/components/tabs/Tabs.js

In function checkValidity on line 236 component fill validityTabs array to return it later in getAllComponents (line 209), but component.destroy() (line 238) cleans all saved components and we are losing all errors.

This issue can be reproduced here http://jsfiddle.net/v38du9y1/1145/
You can press Submit button and see that error about mandatory field is missing.

It will be great if you fix it.
Thanks

Sipaha:
My attempts to fix the component with minor changes was failed and I changed the component to render all children on create without destroing it on tab switch. Should I make pull request with this fix or it is a bad solution for Formio.js?

randallknutson:
Feel free to make a pull request if you have changes that work. We love community provided fixes.

What you are describing sounds like something that is fixed in the 4.x branch which is currently in the beta process of release. We create all the components in the tabs regardless of if they are visible or not. I'd recommend testing out 4.0.0-beta.33 and see if it works better for you.

ProteanHub:
@Sipaha We would be interested in what you did to fix the tab component. We are not quite brave enough to try v4 just yet, but need the tabs component to work.

Sipaha:
> I'd recommend testing out 4.0.0-beta.33 and see if it works better for you.
@randallknutson will be 4.0.0 released soon? Does it have backward compatibility with form configs from 3.*?
Thanks

randallknutson:
4.0.0 is nearing release. The main holdback at the moment is full pdf support and squashing the remaining bugs which is getting a lot of attention at the moment. It will be backwards compatible with form configs.

ProteanHub:
Thanks @Sipaha the bug fix is working well!

Sipaha:
> 4.0.0 is nearing release.
>It will be backwards compatible with form configs.

@randallknutson great to hear! Thank you and your team for the awesome library!
@ProteanHub you're welcome.

ASmeha:
Thank you @Sipaha for the fix!

bohan:
Thanks @Sipaha 
Created a [gist](https://gist.github.com/bohan/8e232851f13853ba5932f1a1247cfb3a) for the modified Tabs.js by @Sipaha
Just in case anyone wants it quickly

PR #1289, opened 2019-05-23
Bugfix: Tabs content validation errors are not displayed on submit

Sipaha:
Issue: https://github.com/formio/formio.js/issues/1249


src/components/tabs/Tabs.js
```diff
@@ -55,44 +53,6 @@ export default class TabsComponent extends NestedComponent {
     return schema;
   }
 
-  build(state, showLabel) {
-    if (this.options.flatten) {
```
travist: Why was this logic removed? We need this for PDF generation.

Sipaha: Sorry. This fix was initially for our project where this logic doesn't need. Fixed it

src/components/tabs/Tabs.js
```diff
@@ -188,56 +189,6 @@ export default class TabsComponent extends NestedComponent {
...
       .addClass(this.tabLinks[index].tabLink, 'active')
       .addClass(this.tabs[index], 'active');
-
-    this.triggerChange();
-  }
-
-  /**
-   * Return all the components within all the tabs.
-   */
-  getAllComponents() {
```
travist: Why did you remove all this code? This is needed to validate the items within each tab.

Sipaha: This code is not required when we render all components from all tabs on the build stage. Validity checking from NestedComponent, in this case, works well.

travist: Well, we need this code back for us to accept this pull request. Other people use this renderer and we cannot just rip out code because it does not fit your use case.

Sipaha: This code is the main problem which this pull request will fix and component behaviour won't change for any users (not only for us). Do you have any auto tests or testers who can verify that nothing changed from a user side? Or you need a deep analysis of the solved problem to accept this request? Removing the build method was a mistake, but this code (getAllComponents and checkValidity) just not working and leads to a lot of bugs. Any attempts to correct this logic will be failed in my opinion.

travist: Ok, thanks for the response. We will take a look as soon as we can if your changes work for validation across tabs.

travist:
Please rebase and resubmit if you still need this.

PR #1299, opened 2019-05-26
semantic-ui wizard

aminmc:
Attempt at adding semantic ui for wizard. 

Wizard steps should look:
Screenshot 2019-05-26 at 08 47 25

and buttons should look:
Screenshot 2019-05-26 at 08 47 38

### DIFF
--- a/src/templates/semantic/wizard/builder.hbs
+++ b/src/templates/semantic/wizard/builder.hbs
@@ -1,1 +1,1 @@
-<div style="color:#6c757d; text-align:center; padding: 0.5rem">{{ t(component.title) }}</div>
+<div class="formio-wizard-builder-component-title">{{ t(component.title) }}</div>
--- a/src/templates/semantic/wizard/builder.hbs
+++ b/src/templates/semantic/wizard/builder.hbs
@@ -1,0 +1,1 @@
+<div style="color:#6c757d; text-align:center; padding: 0.5rem">{{ t(component.title) }}</div>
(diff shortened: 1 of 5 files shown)


travist:
Agree!

aminmc:
@travist @randallknutson would it be possible to get feedback on this PR today? Very keen to get this tested :)

travist:
@randallknutson is the lead dev over 4.x. I will ping him about this item.

src/templates/semantic/wizard/builder.hbs
```diff
@@ -0,0 +1 @@
+<div style="color:#6c757d; text-align:center; padding: 0.5rem">{{ t(component.title) }}</div>
```
randallknutson: We generally don't use inline styles since users may need to override them. Is it possible to add these styles in the CSS based on the generated class names?

src/templates/semantic/wizard/form.hbs
```diff
@@ -0,0 +1,38 @@
+<div style="position: relative;">
```
randallknutson: Same here with the styles.

src/templates/semantic/wizard/builder.hbs
aminmc: @randallknutson I based this on existing inline css for semantic ui. Do you have a preference for where where the css class definitions should go?

src/templates/semantic/wizard/form.hbs
aminmc: @randallknutson this seems to be what's currently in bootstrap and bootstrap 3 of the current 4.x codebase.

src/templates/semantic/wizard/builder.hbs
aminmc: @randallknutson Moved the inline css into cssClasses.js but not sure if this is the right place..

aminmc:
@randallknutson Any updates on this? 

thanks

PR #1301, opened 2019-05-27
4.x-beta.36-fixes

YuryRybakSoftensity:


src/components/select/editForm/Select.edit.data.js
```diff
@@ -183,7 +187,7 @@ export default [
     type: 'textfield',
     input: true,
     label: 'Value Property',
-    key: 'valueProperty',
+    key: ' ',
```
randallknutson: Revert this.

src/components/select/editForm/Select.edit.data.js
```diff
@@ -143,26 +144,29 @@ export default [
...
+            key: `data.${path}`
+          });
+        }
       });
 
       return newItems;
     },
     data: {
-      url: '/form/{{ data.resource }}'
+      url: '/form/{{ data.data.resource }}'
```
randallknutson: This doesn't look correct. If the resource is on the wrong path, we should figure out why it is in the wrong place instead of changing the reference to it.

PR #1303, opened 2019-05-27
FOR-2079: Fixed cursor jumping in WYSIWYGs when typing in ng-formio, added test

Alena-Levina-Softensity:


src/components/textarea/TextArea.js
```diff
@@ -423,7 +426,7 @@ export default class TextAreaComponent extends TextFieldComponent {
     // Set the value when the editor is ready.
     this.dataValue = value;
 
-    this.setWysiwygValue(value, flags);
+    this.setWysiwygValue(value, skipSetting, flags);
```
travist: Why do we need to pass in a variable? Why not just change the line 

```
if (!skipSetting) {
```

to say

```
if (!_.isEqual(value, this.getValue())) {
```

Issue #1290, opened 2019-05-23
Upload file (base 64) not working in wizard


marcopollacci:
Hi, i try to upload file by file input.
Everything it's ok in "normal" form, but when i try to wizard this is what appens:

File.js:853 Uncaught TypeError: Cannot read property 'querySelector' of null
    at eval (File.js:853)
    at Array.forEach (<anonymous>)
    at FileComponent.upload (File.js:807)
    at HTMLInputElement.onChange (File.js:305)
(anonymous) @ File.js:853
upload @ File.js:807
onChange @ File.js:305

`var submitButton = _this14.root.element.querySelector('.formio-component-submit').querySelector('button');`

![image](https://user-images.githubusercontent.com/16632565/58261596-e733f400-7d78-11e9-8cd9-cd32e653edb1.png)

marcopollacci:
I found that in wizard form, not exist "formio-component-submit" class but only "btn-wizard-nav-submit".

Another problem, it's in line 807... file it's not defined

![image](https://user-images.githubusercontent.com/16632565/58262570-ccfb1580-7d7a-11e9-80e3-5cb5e7d75ae0.png)

wag110894:
Hello, 
Thanks for reaching out! We are aware of this regression. We have tagged you on the ticket for this issue and will let you know once it is resolved. For reference, the ticket number is FOR-2304.

Thanks.

marcopollacci:
For now, i just use a little workaround but i'm looking forward to fix ;)

Daedrias:
Hi @marcopollacci , we encountered the same issue on our side. Could you please share your workaround, please?

marcopollacci:
> Hi @marcopollacci , we encountered the same issue on our side. Could you please share your workaround?

Oh..actually it's not really a great workaround...but i was in hurry and it works... :P 
I've just added a try catch option

`try{
 var submitButton = _this14.root.element.querySelector('.formio-component-submit').querySelector('button')
 }catch{
 var submitButton = _this14.root.element.querySelector('.btn-wizard-nav-submit')
 }`

and if your upload field it's in your last page of wizard form, it works :) 

But, bad news iif you have upload field in your different page than last, 'cause actually seems that only way to make thing work (after added try catch ) its to add an hidden submit form. (really dirty and strange things)

noman115pk:
@marcopollacci I have upload field in different pages than last, so i had to add a null check and bypass the #1096 where button gets disabled.

[...]

PR #1309, opened 2019-05-28
Unbreak file uploads in wizards.

franke-mc:
In wizards, the submit button may be absent.

Thanks to @noman115pk for pointing this out.

### DIFF
--- a/src/components/file/File.js
+++ b/src/components/file/File.js
@@ -798,21 +798,22 @@ export default class FileComponent extends BaseComponent {
           // Track uploads in progress.
           if (fileService.uploadsInProgress === undefined) {
             const cssClass = 'uploads-in-progress';
-            var submitButton = this.root.element
-                .querySelector('.formio-component-submit')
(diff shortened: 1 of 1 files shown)


franke-mc:
@wag110894 This may be the quick null-check fix for FOR-2304 mentioned in #1290.

In a second step, probably wizard nav buttons probably should be disabled, too, while file uploads are still in progress. I guess that out of those mentioned in src/Wizard.js, all except 'cancel' should be disabled?
>       { name: 'cancel',    method: 'cancel',   class: 'btn btn-default btn-secondary' },
>       { name: 'previous',  method: 'prevPage', class: 'btn btn-primary' },
>       { name: 'next',      method: 'nextPage', class: 'btn btn-primary' },
>       { name: 'submit',    method: 'submit',   class: 'btn btn-primary' }

travist:
Looks good @franke-mc. I am working on a new release today and will include this one.

PR #1332, opened 2019-05-31
Fixed recaptcha rendering

AlexeyDavyd:


src/components/recaptcha/ReCaptcha.js
```diff
@@ -23,6 +23,10 @@ export default class ReCaptchaComponent extends Component {
     };
   }
 
+  render() {
+    super.render(this, true);
```
randallknutson: Why are we passing "this" to render? I'm not sure that is right.

PR #1338, opened 2019-06-01
Bugfix: updated ul to use div for lists for semantic ui wizard

aminmc:
Screenshot 2019-06-01 at 15 29 50

The buttons still have the ul bullet points so updated to use div with class. 

The only issue is the margin between the steps, components and buttons.  Trying to avoid inline css so I'm not sure what the right approach should be here to create a gap between each. @randallknutson happy to get some feedback on the approach.

### DIFF
--- a/src/templates/semantic/wizard/form.hbs
+++ b/src/templates/semantic/wizard/form.hbs
@@ -13,26 +13,26 @@
   <div class="wizard-page" ref="{{wizardKey}}">
     {{components}}
   </div>
-  <ul class="ui horizontal list">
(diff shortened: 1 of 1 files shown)


aminmc:
@randallknutson Just wondering if you could advise on what the best way is to deal with:

Screenshot 2019-06-03 at 17 54 54

randallknutson:
Looks good!

aminmc:
@randallknutson Thanks for merging. Any thoughts on the wizard builder menu issue?

PR #1353, opened 2019-06-05
Add fix for nested forms with 'Save as reference' from 4.x

franke-mc:
Fixes [angular-formio issue #299](https://github.com/formio/angular-formio/issues/299).
(See also discussion in [angular-formio issue #296](https://github.com/formio/angular-formio/issues/296).)

### DIFF
--- a/src/components/form/Form.js
+++ b/src/components/form/Form.js
@@ -142,6 +142,7 @@ export default class FormComponent extends BaseComponent {
       });
       this.subForm.url = this.formSrc;
       this.subForm.nosubmit = this.nosubmit;
+      this.subForm.nosubmit = false;
       this.restoreValue();
       this.subFormReadyResolve(this.subForm);
       return this.subForm;


src/components/form/Form.js
```diff
@@ -142,6 +142,7 @@ export default class FormComponent extends BaseComponent {
       });
       this.subForm.url = this.formSrc;
       this.subForm.nosubmit = this.nosubmit;
+      this.subForm.nosubmit = false;
```
randallknutson: I'm curious why adding this line improves things. It removes the functionality from the line above which is designed to push the nosubmit setting from the parent form to the sub form.

franke-mc: Note that the same is true in 4.x.

See [here](https://github.com/formio/angular-formio/issues/296#issuecomment-499149013) for the explanation: without this line, when submitForm is called (from Select's beforeSubmit), nosubmit is still true and submitForm is aborted early, but the caller does not detect the failure and proceeds with _id: undefined and form: undefined.
I haven't understood why the other problem (subform data values being kind-of empty) is healed by this one-liner as well, but that's what I observe, and right now I don't care. (In our application, we only use 'reference': false.)

For a plausibility check, look at the [twin line](https://github.com/formio/formio.js/blob/master/src/components/form/Form.js#L394) inside `setValue()` which is there exactly to undo the `.nosubmit=true` assignment happening inside `setUrl()`, supporting my assumption that `.nosubmit=true` is always wrong for subforms. But I agree, understanding the exact intended semantics of nosubmit would be nice.

travist:
This should be fixed with https://github.com/formio/formio.js/commit/f1f0a87fc8e9de52aa7a0e586b002f5bdca13544

PR #1370, opened 2019-06-12
tooltip update: passing title into render 

aminmc:
Passing this.title into the render method based on comment from @randallknutson. My only concern is that this.title is being set inside the tooltips array.


randallknutson:
Sorry for the confusion. This is what I meant. https://github.com/formio/formio.js/commit/f474987cd72f93123af4842a76c491d1141e0b57

aminmc:
Ah ok! 😀

aminmc:
Shall I close this PR?

PR #1372, opened 2019-06-12
FOR-2314: Re-fixed builder buttons missing after redraw + fixed Wizard building being broken

Alena-Levina-Softensity:


src/WizardBuilder.js
```diff
@@ -84,7 +84,7 @@ export default class WizardBuilder extends WebformBuilder {
 
   gotoPage(page) {
     this.currentPage = page;
-    this.redraw();
+    this.redraw(true);
```
Alena-Levina-Softensity: @travist I had to add this since ```WizardBuilder``` is the only class which instance does need to be redrawn in builder mode and that sounds a bit like it should have its own implementation of redraw method, but to avoid duplicating code I just added an option to ```Base``` class redraw method

PR #1375, opened 2019-06-12
Removing the dependency on native-promise-only

travist:


travist:
This breaks angular builds.

PR #1377, opened 2019-06-13
Fix desync of PDFBuilder component IDs vs PDF component IDs after set…

aiwebb:
…Form-initiated rebuilds


aiwebb:
PDF.setForm regenerates components from form schema, granting new IDs to each component. These are posted to the iframe, but were not being returned back to the PDFBuilder. PDFBuilder now updates its own component list according to the results of PDF's setForm logic.

PR #1395, opened 2019-06-17
Add file component to PDF builder components list

aiwebb:


aiwebb:
See associated PR at formio/formio-pdf#11

PR #1438, opened 2019-07-02
Added ability to provide HTML attributes for components inputs

Alena-Levina-Softensity:


src/components/base/editForm/Base.edit.layout.js
```diff
@@ -1,6 +1,25 @@
 export default [
   {
-    type: 'textfield',
-
+    input: true,
```
travist: I may prefer to use a DataMap here so that the attributes can be stored as the following. We should also decide on the namespace. I actually just prefer to use ```attributes```.

```js
"attributes": {
  "one": "one-data",
  "two": "two-data"
}
```

Alena-Levina-Softensity: Done

PR #1467, opened 2019-07-10
added tree support for semantic ui

aminmc:
Added tree component theme form semantic UI.  Looks like this:

Screenshot 2019-07-10 at 11 20 37


aminmc:
Hello...apologies for adding two change requests into a single PR. Hopefully both make  sense.

src/Formio.js
```diff
@@ -650,7 +650,7 @@ export default class Formio {
       .then(() => Formio.pluginGet('staticRequest', requestArgs)
         .then((result) => {
           if (isNil(result)) {
-            return Formio.request(url, method, requestArgs.data, requestArgs.opts.header, requestArgs.opts);
+            return Formio.request(requestArgs.url, requestArgs.method, requestArgs.data, requestArgs.opts.header, requestArgs.opts);
```
randallknutson: Did you mean to include this change? We can quickly merge the tree changes in but this is a major change to a foundational library and we may need to take a closer look at it.

src/Formio.js
```diff
@@ -670,7 +670,7 @@ export default class Formio {
       .then(() => Formio.pluginGet('request', requestArgs)
         .then((result) => {
           if (isNil(result)) {
-            return Formio.request(url, method, requestArgs.data, requestArgs.opts.header, requestArgs.opts);
+            return Formio.request(requestArgs.url, requestArgs.method, requestArgs.data, requestArgs.opts.header, requestArgs.opts);
```
randallknutson: Did you mean to include this change? We can quickly merge the tree changes in but this is a major change to a foundational library and we may need to take a closer look at it.

aminmc: @randallknutson I was trying to create a separate PR for this change but accidentally added this to the semantic ui PR. Happy for you to merge the semantic one and I can create a separate PR for the other change.

aminmc: requestArg can be intercepted using the plugin architecture. Previously opts.header, data and opts where being passed into the Formio.request. By passing the url it makes it possible to modify the url if needed (we have a use case for this).

randallknutson: I'd appreciate separating these two items so we can evaluate them separately. The tree for semantic is straightforward and we'll merge quickly but this change is going to require some discussion and testing to make sure we don't break anything.

aminmc:
Understood. Happy to start a separate discussion about the other part of the PR

PR #1480, opened 2019-07-16
[3.x] Change textfield|area isEmpty() to test trimmed value

aiwebb:


travist:
Tests are failing with what seems to be a legit issue. When you fix this in 3.x, please fix this in 4.x as well.

travist:
This may fix it.... https://github.com/formio/formio.js/commit/86a22b46d4a86f5a1033a7e06b1afa23fdbc2ba7

PR #1482, opened 2019-07-17
FOR-2437: Fixed textarea auto expanding issues with tabs component.

roflanKisel:


src/components/tabs/Tabs.js
```diff
@@ -189,6 +181,16 @@ export default class TabsComponent extends NestedComponent {
       .addClass(this.tabLinks[index].tabLink, 'active')
       .addClass(this.tabs[index], 'active');
 
+    const components = this.hook('addComponents', tab.components, this);
```
travist: Why was this needed? This seems pretty disruptive since I am not sure what this would do AFTER ```this.restoreValue()``` since I believe restoreValue should occur after the components have been added and not before. I think this will require a lot of testing before we can pull this in.

roflanKisel: Moved it after adding components and tested it.

PR #1484, opened 2019-07-17
[4.x] Add toString() step to TextField before trimming

aiwebb:


travist:
Here is how I fixed it in 4.x. 

https://github.com/formio/formio.js/commit/86a22b46d4a86f5a1033a7e06b1afa23fdbc2ba7

Does this not work for you?

aiwebb:
That does resolve the multiple-value issue. Commit d893086 addresses the object-as-value (tests in 3.x were failing in cases where `value === {}`; the open PR for that was updated with both fixes, while this secondary PR was opened for 4.x since the original was already merged).

PR #1485, opened 2019-07-18
1447 - component 'Panel' rebuilding did not update content.

jtamm-red:
We have to check this.element. If this.element has already created, then must to update. If not, then will create.


src/components/panel/Panel.js
```diff
@@ -105,10 +105,18 @@ export default class PanelComponent extends NestedComponent {
     let panelClass = 'mb-2 card border ';
     panelClass += `panel panel-${this.component.theme} `;
     panelClass += this.component.customClass;
-    this.element = this.ce('div', {
```
travist: This should be put inside of the ```this.empty``` method. And from what I understand, ```this.empty()``` does get called before a "rebuild" so maybe the issue is that a Panel is not implementing ```this.empty()``` correctly.

jtamm-red: this.empty() worked well (cleaned children elements), but new element attacting did not work.
If data is changed, then called Base.redraw().
https://github.com/formio/formio.js/blob/master/src/components/base/Base.js#L1476
`this.build(this.clear());`
Actions:
1. called component.empty() - clean children elements
2. called component.build() - create element only.

https://github.com/formio/formio.js/blob/master/src/components/base/Base.js#L758-L762

Base.createElement() - create or update existed element.

We have to check an element existing. If element is already created, then we have to update element.

travist:
Please test if this still exists in latest master, and then rebase and resubmit the pull request if it is still needed.

Issue #1383, opened 2019-06-13
[BUG] Deletion of last character doesn't trigger change


sebastianhaberey:
## Environment

- Formio.js version: latest (3.22.4)

## Steps to Reproduce

1. enter some characters in field of type textfield / number (maybe others?)
2. delete characters one by one using backspace

## Expected behavior

Each deletion triggers change.

## Observed behavior

Deletion of last character doesn't trigger change.

## Example

see https://codepen.io/sebastianhaberey/pen/WqvBGq?editors=0011 (watch console for events)

Alena-Levina-Softensity:
@sebastianhaberey I took a look at your codepen and at line 37 you have following code: 
```submission.data.textfield && console.log(submission.data.textfield);```
and first part of it (```submission.data.textfield &&```) is the reason why you don't see console output for empty textfield value. You can remove this part ot put a breakpoint at this place and you'll see that change event actually gets fired.

Please let me know you have any more questions on this

sebastianhaberey:
Hi Alena,

you are right! My example wasn't very good. I didn't expect the textfield to become undefined once the last character was deleted. I've made an updated codepen that's a bit closer to what we have in our application and shows the bug:

https://codepen.io/sebastianhaberey/pen/MMoLyg

sebastianhaberey:
Here is a pull request that fixes the issue. The proposed solution is for the number component to return null if no value is set, so the entry is not removed entirely from the internal data structure.

Note that this pull request **breaks the 'client-only' feature**, which is why we disabled the corresponding unit test. We are convinced that this feature is mostly broken, since it doesn't work for nested components and it doesn't work for elements that have a default value set (the unit test has been modified to verify the last point).

heather-formio:
This looks to be working as expected in the current version. Closing this as it is stale, please feel free to reopen at any time. Thank you!

PR #1487, opened 2019-07-19
#1383 number component now returns null if not set...

sebastianhaberey:
...instead of being removed from submission data

- Number.getValueAt(): now returns null if not set
- Base.dataValue() (setter): don't unset the value on null (only on undefined)
- unit tests: Number component now appears in data grid if value is null
- disabled test for 'client-only' feature (feature has multiple issues)


travist:
I am a little leery about this pull request. These methods have been through several back-and-forth iterations and this feels like yet another one. I also don't like making commits that comment how things are broken. I would feel much better if tests were written that illustrate the bugs and then those bugs are actually resolved.

sebastianhaberey:
Yes I agree with you there. I guess the reason why I made this pull request is to show where I believe the problem is, and a possible way to fix it. So I can totally understand if you don't want to merge it as-is! But I would want to keep #1383 open as long as it isn't fixed - is there any reason why it was closed?

travist:
Closing for now. Please test in new master branch and re-open if it is still needed.

PR #1495, opened 2019-07-23
added tree support

aminmc:
Added tree semantic theme.


aminmc:
@randallknutson created new PR for semantic tree support.

PR #1508, opened 2019-07-25
Fixed issues with search and infinite scroll in select component.

roflanKisel:


src/components/select/Select.js
```diff
@@ -54,11 +54,17 @@ export default class SelectComponent extends BaseComponent {
     // Keep track of the select options.
     this.selectOptions = [];
 
-    // Keep track of the last batch of items loaded.
```
travist: So this will now work only if they are linked to our Resource? I don't think this is correct... this should work regardless if they are linked to our resources or not.

src/components/select/Select.js
```diff
@@ -738,7 +738,7 @@ export default class SelectComponent extends BaseComponent {
 
     // Add value options.
     if (this.addValueOptions()) {
-      this.restoreValue();
+      this.choices.setChoiceByValue(this.dataValue);
```
travist: This will now break if this.choices is not defined (like when using the HTML5 widget)

src/components/select/editForm/Select.edit.data.js
```diff
@@ -144,22 +144,24 @@ export default [
...
+            key: `data.${path}`
+          });
+        }
       });
 
       return newItems;
     },
     data: {
-      url: '/form/{{ data.resource }}'
+      url: '/form/{{ data.data.resource }}'
```
travist: This does not seem right.... why do we have to reference the data within data here?

PR #1511, opened 2019-07-25
Make form config available to eval context.

randallknutson:


src/Component.js
```diff
@@ -510,7 +510,8 @@ export default class Component {
       util: FormioUtils,
       user: Formio.getUser(),
       moment,
-      instance: this
```
travist: Why can't someone just use the following?...

```
{{ form.config }}
```

The form variable should be available in the evalContext already.

PR #1513, opened 2019-07-26
Interpolate attribute value

### LINKED ISSUE #1389: Javascript Custom Default value in HTML Element
In a HTML Element if I have a Custom Defaut Value like this:

```
value = data.downloadUrl;
```

how can I use it in `a` HTML Tag `<a href=""> ...`

image

In HTML Element the `value` (Javascript code) is never used - the functionality is not the same as the others components.

jftsaraiva:
Fixes #1389


travist:
Will you also make a pull request of this against the 4.x branch?

PR #1515, opened 2019-07-29
4.x

pixie79:
Revert change to search filter to allow users to specific. As per 3.x and ticket 1135


aminmc:
@randallknutson Can this PR be reviewed as soon as possible. We're unable to use search for a select box at the moment

travist:
We will pull this into the next build. Thanks for the contribution.  This will not be back-ported to 3.x since it is a breaking change.

Issue #1490, opened 2019-07-20
Firefox 68 and edit dialog component scroll background.


daebak74:
If you open https://formio.github.io/formio.js/app/builder with firefox the edit component modal scroll at the beginning than start to scroll the background.

randallknutson:
Do you know if something changed on this recently? We've tested this in firefox frequently and it hasn't been reported before. Could this also be some CSS that is causing the issue?

daebak74:
Reproduce the issue is simple go to formio demo page from firefox 
http://formio.github.io/formio.js/app/builder
Add any kind of component for example text field and try to scroll up and down until the top and botton you will see that the edit modal will block and start to scroll the background.

sanluiss:
Same issue, scroll not working properly on edit dialog (Firefox 68.0.1 ubuntu 19.04), tested on https://formio.github.io/formio.js/app/builder

jhen1422:
Thank you for the report. 

We were able to reproduce the issue and will discuss priority internally with our development team during our next planning meeting.

If you are needing this fixed quickly, please contact support@form.io for information on expediting a fix.

aiwebb:
I suspect this was caused by some recent changes to Firefox's scrolling behavior; they've recently changed their heuristics for deciding which component acts as the scroll target when there are multiple candidates.

I've explicitly allowed scrolling on modal content areas and explicitly disallowed scrolling on the document body while a modal is open. That appears to resolve the ambiguity and restore proper scrolling behavior in the latest version of Firefox.

daebak74:
Yes confirm with release 4 the problem is fixed. The demo online https://formio.github.io/formio.js/app/builder still present the problem because other release. There is online a demo with release 4 ?

aiwebb:
No demo builder for 4.x yet - we'll be updating that and the documentation for 4.x in the coming months.

The current builder will be updated on our next `gh-pages` branch update, which generally lags behind `master` a little bit. We'll be sure to update it after #1516 gets merged in.

PR #1516, opened 2019-07-29
Fix competing scroll behavior between body and modals

aiwebb:


aiwebb:
Resolves Firefox scrolling issue in #1490.

travist:
Also need this for 4.x?

PR #1517, opened 2019-07-29
[4.x] Fix competing scroll behavior between body and modals

aiwebb:


aiwebb:
Resolves Firefox scrolling issue in #1490.

PR #1522, opened 2019-07-31
Allow Sanitize Config for Forms options

edwinanciani:
## Change 
Add Dompurify configuration to allow video/image tags, attr, and URLs.

## Files Changed
Form.js

### DIFF
--- a/src/Form.js
+++ b/src/Form.js
@@ -177,12 +177,26 @@ export default class Form extends Element {
    * @returns {*}
    */
   sanitize(dirty) {
-    return sanitize(dirty, {
+    // Dompurify configuration
+    const sanitizeOptions = {
       ADD_ATTR: ['ref'],
       USE_PROFILES: {
         html: true
       }
-    });
+    };
+    // Allow tags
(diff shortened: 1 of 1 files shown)


src/Form.js
```diff
@@ -177,12 +177,26 @@ export default class Form extends Element {
...
       }
-    });
+    };
+    // Allow tags
+    if (this.options.sanitizeConfig && Array.isArray(this.options.sanitizeConfig.allowedTags) && this.options.sanitizeConfig.allowedTags.length > 0) {
+      sanitizeOptions.ALLOWED_TAGS = this.options.sanitizeConfig.allowedTags;
+    }
+    // Allow attributes
+    if (this.options.sanitizeConfig && Array.isArray(this.options.sanitizeConfig.allowedAttr) && this.options.sanitizeConfig.allowedAttr.length > 0) {
+      sanitizeOptions.ALLOWED_ATTR = this.options.sanitizeConfig.allowedTags;
```
randallknutson: There is a typo here "allowedTags" should be "allowedAttrs". Also, we need to ensure that "ref" is included even if they provide overrides.

Issue #1530, opened 2019-08-01
[BUG] ChoicesJS and Radio/Check boxes


aminmc:
## Environment
Please provide as many details as you can:

- Hosting type
  - [] Form.io
  - [ x] Local deployment
    - Version:  
- Formio.js version: 4.0.0.-rc-20
- Frontend framework: react-formio
- Browser: chromse
- Browser version:

## Steps to Reproduce
1. Create a wizard
2. Add a select component with ChoicesJS and URL source
3. Add second Page
4. Add radio with 2 options
5. Save form
6. Select from drop down page in Page 1
7. Go to Page 2 and select radio option
8. Page 1 selection has been wiped clean

## Expected behaviour
Going back to Page 1 after selecting an option should not clear the drop down selected in page 1. 

Folks this is a critical issue for us. This was working fine before 4.x, and this has effectively caused all of our forms to stop working when we have select boxes using ChoicesJS and radio buttons. Can someone please investigate this issue as soon as possible?

aminmc:
Also might be related to this:
https://github.com/formio/formio.js/issues/1492

wag110894:
Hello Amin,

Thanks for reaching out! We were able to discuss this item on our Afternoon Call and did deem it as a High Priority item. We will let you know once this item is resolved.

Thanks

travist:
This will be resolved when ^ is merged.  Should be later today.

aminmc:
This is working as expected.  Thank you guys!

Issue #1492, opened 2019-07-22
[BUG] Conditional logic on wizard causes weird behaviour 


aminmc:
## Environment
Please provide as many details as you can:

- Hosting type
  - [] Form.io
  - [x] Local deployment
    - Version: 
- Formio.js version: 4.0.0-rc.12
- Frontend framework:
- Browser: Chrome
- Browser version:

## Steps to Reproduce
See below:
https://codesandbox.io/s/vigilant-panini-d6999

After selecting option 'a' or 'b' from page 1 select box you will be presented with page 2. If you select any of the radio option the page then tries to submit.  This seems to be an issue when you use choicesjs as the drop down list. It works fine with HTML 5 drop down. Or if you do not put any conditional logic on the page,

For code or form JSON, please enclose in a code block:

```json
{
  type: "form",
  tags: [],
  components: [
    {
      title: "Page 1",
      theme: "default",
      breadcrumb: "default",
      breadcrumbClickable: true,
      buttonSettings: {
        previous: true,
        cancel: true,
        next: true
      },
      tooltip: "",
      customClass: "",
      collapsible: false,
      hidden: false,
      disabled: false,
      alwaysEnabled: false,
      tableView: false,
      key: "page1",
      tags: [],
      properties: {},
      customConditional: "",
      conditional: {
        json: "",
        show: "",
        when: "",
        eq: ""
      },
      nextPage: "",
      logic: [],
      attributes: {},
      overlay: {
        style: "",
        left: "",
        top: "",
        width: "",
        height: ""
      },
      type: "panel",
      label: "Page 1",
      components: [
        {
          label: "Select",
          labelPosition: "top",
          widget: "choicesjs",
          placeholder: "",
          description: "",
          tooltip: "",
          customClass: "",
          tabindex: "",
          hidden: false,
          hideLabel: false,
          autofocus: false,
          disabled: false,
          alwaysEnabled: false,
          tableView: true,
          multiple: false,
          dataSrc: "values",
          indexeddb: {},
          data: {
            values: [
              {
                label: "A",
                value: "a"
              },
              {
                label: "B",
                value: "b"
              },
              {
                label: "C",
                value: "c"
              },
              {
                label: "D",
                value: "d"
              }
            ],
            resource: "",
            json: "",
            url: "",
            custom: ""
          },
          valueProperty: "",
          template: "<span>{{ item.label }}</span>",
          defaultValue: "",
          searchEnabled: true,
          selectThreshold: 0.3,
[...]

PR #1537, opened 2019-08-02
Fixed issue where the wizard would reset the values from select compo…

travist:
…nents.


travist:
This fixes https://github.com/formio/formio.js/issues/1530

travist:
This also fixes https://github.com/formio/formio.js/issues/1492

PR #1542, opened 2019-08-05
Issue: 1541 fixed.

sudhakar29495:


src/components/file/File.js
```diff
@@ -237,15 +237,22 @@ export default class FileComponent extends BaseComponent {
...
       style: 'opacity: 0; position: absolute;',
       tabindex: -1, // prevent focus
       onChange: () => {
         this.upload(this.hiddenFileInputElement.files);
         this.hiddenFileInputElement.value = '';
       }
-    });
+    }
+    if (this.component.multiple) {
+      inputFile = {
```
travist: Why is a clone needed here? Why not just say ```inputFile.multiple = true```?

sudhakar29495: Yeah, you are right. I will change that way.

travist:
Needs rebase, then resubmit when it is ready.

PR #1561, opened 2019-08-14
semantic fixes

aminmc:
Minor semantic ui fixes for bulder


travist:
Going to go ahead and take this. Will be making some changes to all templates so make sure you fetch master before doing anything radical after the changes come in by the end of the week.

Issue #1573, opened 2019-08-16
SetBaseUrl fails to pass regex when added with '/' at the end


adornala:
Formio.setBaseUrl("https://sample.com/formio/")  fails, because regex couldn't parse '/' at the end.

` 
{
     key: "getUrlParts",
     value: function getUrlParts(url, formio) {}
}
`

Formio.js - Line: 835 - 847

travist:
Looks like it should just get trimmed before it is parsed. Pull requests welcome.

PR #1579, opened 2019-08-20
Fixed SetBaseUrl to pass regex when added with '/' at the end

adornala:
Fixed issue #1573 

Also added additional logging to notify user if Ambiguous letters are present in url.


src/Formio.js
```diff
@@ -104,7 +104,15 @@ export default class Formio {
 
     const hostparts = this.getUrlParts(path);
     let parts = [];
-    const hostName = hostparts[1] + hostparts[2];
+    let hostName = '';
+    if (hostparts && hostparts.length >= 2) {
+      hostName = hostparts[1] + hostparts[2];
+    }
+    else {
+      console.log('Ambiguous letter in url\'s might fail, please make sure setBaseUrl and ' +
```
travist: I would prefer not to do this (throwing an error if they are not greater than 3). I know that applications that use Open source servers would most likely fail here (since those don't have projects). Please remove this logic and we will accept the "trim" method you put in this file below.

adornala: Okay @travist . Is truthy check on hostparts ok? That's what is causing the issue too. When regex fails it's defined as null.

adornala: Also @travist don't you think application will fail if hostparts doesn't actually have length of 2 or above? Code might fail throwing Array out of bound error, isn't it!

travist:
Closed due to inactivity.

PR #1581, opened 2019-08-20
Edit Grid Modal

edwinanciani:
## Change
Open a modal when 'Add another' is clicked to add or edit data.

## Files Changed
EditGrid.js 
EditGrid.edit.templates.js

### DIFF
--- a/src/components/editgrid/EditGrid.js
+++ b/src/components/editgrid/EditGrid.js
@@ -339,7 +339,23 @@ export default class EditGridComponent extends NestedComponent {
     });
     editRow.components = this.createRowComponents(editRow.data, rowIndex);
     this.checkRow(this.data, editRow);
-    this.redraw();
+    if (this.component.modal) {
+      this.addRowModal(rowIndex);
+    }
+    else {
+      this.redraw();
+    }
+  }
+
+  addRowModal(rowIndex) {
+    const formComponents =  this.ce('div');
+    formComponents.innerHTML = this.renderComponents(this.editRows[rowIndex].components);
+    const dialog = this.component.modal ? this.createModal(formComponents) : undefined;
+    dialog.refs.dialogContents.appendChild( this.ce('button', {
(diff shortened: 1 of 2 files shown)


src/components/editgrid/EditGrid.js
```diff
@@ -402,7 +423,7 @@ export default class EditGridComponent extends NestedComponent {
     this.redraw();
   }
 
-  saveRow(rowIndex) {
+  saveRow(rowIndex, dialog) {
```
travist: Instead of this, I would prefer to contain the dialog.close within the ```addRowModal``` method. Basically turn it into this.

```js
addRowModal(rowIndex) {
    const formComponents =  this.ce('div');
    formComponents.innerHTML = this.renderComponents(this.editRows[rowIndex].components);
    const dialog = this.component.modal ? this.createModal(formComponents) : undefined;
    dialog.refs.dialogContents.appendChild( this.ce('button', {
      class: 'btn btn-primary',
      onClick: () => {
        this.saveRow(rowIndex);
        dialog.close();
      },
    }, this.component.saveRow || 'Save'));
    this.attachComponents(dialog, this.editRows[rowIndex].components);
  }	  }
```

src/components/editgrid/EditGrid.js
```diff
@@ -339,7 +339,23 @@ export default class EditGridComponent extends NestedComponent {
...
+    else {
+      this.redraw();
+    }
+  }
+
+  addRowModal(rowIndex) {
+    const formComponents =  this.ce('div');
+    formComponents.innerHTML = this.renderComponents(this.editRows[rowIndex].components);
+    const dialog = this.component.modal ? this.createModal(formComponents) : undefined;
+    dialog.refs.dialogContents.appendChild( this.ce('button', {
```
travist: This would also be ```undefined``` if this.component.modal is false. Should have an if statement around this.

PR #1589, opened 2019-08-22
Number component fix

vlad-shusterman:
fix bug with unsettable empty string


gshivam44:
I checked this in the online demo builder but it is not working.
Is this fixed in the online demo builder?

Issue #1625, opened 2019-09-04
[BUG] Console error when rendering builder


aminmc:
Hi All

We're using the latest version of the 4.x and when displaying the builder we're seeing the following:

<img width="1310" alt="Screenshot 2019-09-04 at 07 43 44" src="https://user-images.githubusercontent.com/290639/64231800-714b3700-cee8-11e9-9d78-d1542a6664ec.png">

This is a blank builder, on inspecting the line it's trying to do a forEach on a string:
<img width="1408" alt="Screenshot 2019-09-04 at 07 49 43" src="https://user-images.githubusercontent.com/290639/64231875-9f307b80-cee8-11e9-8c72-4cc549eeb708.png">

randallknutson:
Thanks. We are working on adding functionality for existing resources to the core library and it may be that our developer forgot to check if it exists first. I'll pass this along to them to get fixed.

randallknutson:
We're not sure why the components are coming in as a string but we can easily add a check and parse around it. Would this fix the issue? https://github.com/formio/formio.js/pull/1630

aminmc:
Strange that we see it with an empty builder. Nothing at this point has been added.

aminmc:
no longer an issue with 4.2.0-rc.7

PR #1630, opened 2019-09-04
Handle case when components array came as string

AlexeyDavyd:


src/WebformBuilder.js
```diff
@@ -334,7 +334,17 @@ export default class WebformBuilder extends Component {
         default: index === 0,
       };
 
-      eachComponent(resource.components, (component) => {
+      let { components } = resource;
+      if (!Array.isArray(components)) {
+        try {
+          components = JSON.parse(components);
```
travist: In what case does this happen?  This seems like a band-aid over a more fundamental problem. The components should never get passed in as a JSON string.

AlexeyDavyd: This was hotfix for that issue: https://github.com/formio/formio.js/issues/1625
I've asked Randall to provide more info about use case, but he advised me to only check if components is an array. So maybe we need to ask them to provide their environment, because I can't reproduce it on my machine.

travist: I think we need to investigate why this is coming in as a string. I don't want to include a hack to handle a malformed form.

travist: I don't think we will pull this in. We need to figure out why the "components" is being sent in as a string and not an array. That is the real bug we should be fixing instead of this... People may start abusing this new "feature"

travist:
I don't think we will pull this in. We need to figure out why the "components" is being sent in as a string and not an array. That is the real bug we should be fixing instead of this... People may start abusing this new "feature"

Issue #1652, opened 2019-09-10
[BUG] Infinite onChange events with Resource select boxes in a nested form


darabi:
## Environment
Please provide as many details as you can:

- Hosting type
  - [ ] Form.io
  - [x] Local deployment
    - Version: 
- Formio.js version: 4.0.0-rc.31 (via react-formio)
- Frontend framework: react-formio 4.0.0-rc.18
- Browser: Chrome + Firefox
- Browser version: Chrome 76.0.3809.100 Linux

## Steps to Reproduce

I am not able to create a small reproducible sample. I have a form with 3 nested forms; they have some Selects which reference Resources

In a simple case, everything works as expected, but in a more complex case, I have Webform.onChange being fired indefinitely.

## Expected behavior

A small number of onChange calls until everything has stabilised.

## Observed behavior

Infinite onChange calls.

randallknutson:
We've seen this happen if there are components have the same property name (key). Can you check if there are any duplicates that could be causing this?

darabi:
Thank you for your quick response.

I made sure all keys have a unique value but the behaviour remains the same.

darabi:
While debugging, I stumbled upon `NestedComponent.checkData` which looks like this:

```
  /**
   * A more performant way to check the conditions, calculations, and validity of
   * a submission once it has been changed.
   *
   * @param data
   * @param flags
   */
  checkData(data, flags, source) {
    flags = flags || {};
    let valid = true;
    if (flags.noCheck) {
      return;
    }

    // Update the value.
    let changed = this.updateValue(null, {
      noUpdateEvent: true
    }, source);

    // Iterate through all components and check conditions, and calculate values.
    this.getComponents().forEach((comp) => {
      if (comp.checkData) {
        valid &= comp.checkData(data, flags);
      }
      changed |= comp.calculateValue(data, {
        noUpdateEvent: true
      });
      comp.checkConditions(data);
      if (!flags.noValidate) {
        valid &= comp.checkValidity(data);
      }
    });

    // Trigger the change if the values changed.
    if (changed) {
      this.triggerChange(flags, changed);
    }

    // Return if the value is valid.
    return !!valid;
  }
```
If I interpret the code correctly, the lines

```
    // Update the value.
    let changed = this.updateValue(null, {
      noUpdateEvent: true
    }, source);
```
reset the value of the nested component completely to then iterate through the components, re-calculate the values and check the validity.

[...]

PR #1653, opened 2019-09-10
Select: do not trigger an update when setting a provided value (#1652)

darabi:
The argument to `setValue` is not passed to `getFlags` any more, so
`true` as second argument is never translated to `noUpdateEvent:
true`.

With this change, the described behaviour in #1652 is resolved.


travist:
Awesome! Thanks.

PR #1655, opened 2019-09-10
Fixed issue when skip&limit become undefined if select resource data type

dianatsevan:


src/utils/Evaluator.js
```diff
@@ -13,14 +13,10 @@ const Evaluator = {
...
       return _.noop;
     }
-    else {
-      return new Function(...params, func);
-    }
+
+    return new Function(...params, func);
   },
   template(template, hash) {
-    if (typeof template === 'function') {
```
travist: This will break pre-rendered templates. Why was this removed?

PR #1657, opened 2019-09-10
Revert "Additional validation for date\time component 4.x"

travist:
Reverts formio/formio.js#1645


travist:
I had to revert this for a couple of reasons.
  1.) It is way too intrusive.  We need an option on the Validation page that allows people to "opt-in" to this type of date validation and not make it by default. Also, it seems that if a value is required it will automatically trigger this validation. This is very disruptive for existing forms.

  2.) There is code that is sprinkled in "common" libraries that should be pulled into sub classes. It does not feel very compartmentalized.

This needs some work.

PR #1658, opened 2019-09-10
Fix performance of conditional logic within a nested component.

randallknutson:


travist:
Fixing in a different PR.

PR #1666, opened 2019-09-11
NestedComponent.checkData: no change trigger when resetting the value

darabi:
I think this is relevant for the behaviour described
in #1652: without this change v4.0.3 loops. With it,
the firing of change events stops without any observable
adverse effects.


travist:
Closing since my pull request https://github.com/formio/formio.js/pull/1660 fixes this problem.

PR #1688, opened 2019-09-15
enabling users to override fileService and delete a file

aminmc:
Currently attach and download file can be overridden by passing in fileService into the options configuration. However delete is being handled not in the same way. This PR enables the option to override the delete file by using fileService.  Happy to discuss this further. Our use case means we need to configure our fileService to do some pre processing.

### DIFF
--- a/src/components/file/File.js
+++ b/src/components/file/File.js
@@ -3,9 +3,10 @@ import { uniqueName } from '../../utils/utils';
 import download from 'downloadjs';
 import _ from 'lodash';
 import Formio from '../../Formio';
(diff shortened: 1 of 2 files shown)


randallknutson:
This looks good. The only thing is we need to make sure it works correctly for other storage providers as well.

aminmc:
I added the additional check of if deleteFile function exists. I can add it to the other providers. We need the override functionality. Do you want the others done first and merge ?

randallknutson:
Looks good. We'll merge this next build.

PR #1692, opened 2019-09-16
Wizard first page disabling

RomanLetsuk:


randallknutson:
This is a bit too heavy handed for changes. I'll push up another branch with more minimal changes.

PR #1698, opened 2019-09-17
updated select component delete events

RomanLetsuk:


src/components/select/Select.js
```diff
@@ -727,7 +727,7 @@ export default class SelectComponent extends Field {
         const { key } = event;
 
         if (['Backspace', 'Delete'].includes(key)) {
-          this.setValue(null);
```
travist: Why not make this ```this.setValue(this.emptyValue)```?

PR #1710, opened 2019-09-19
Fix issue where select html5 components can't restore value properly.

randallknutson:


randallknutson:
The issue is that at this line https://github.com/formio/formio.js/blob/master/src/components/select/Select.js#L1142 we look for the selectOption.element which is no longer set on html5 select components. This restores the functionality so it works correctly.

PR #1713, opened 2019-09-19
fix selection error is not clear after successfully validate

furytara:
error on select (component) should be clear


randallknutson:
Thanks for the fix! We already have a fix in place here https://github.com/formio/formio.js/pull/1708 and are waiting on a build.

randallknutson:
It is available in v4.1.0

furytara:
Fix in #1708 just fix in Component.js, we need to fix in Select.js too. Error was not clear if component is selection (dropdown)

Issue #1639, opened 2019-09-07
 How to localize form builder


Vedhachala:
Hi, 
We are using the angular wrapper for formio (form builder in specific)
Formiojs version : 3.21.5

I want to use form builder in other languages. It would help if you can tell the way to add i18n translations for form builder.

Thanks,
Vedhachala

randallknutson:
http://formio.github.io/formio.js/app/examples/language.html

Vedhachala:
@randallknutson

 I am able to do this for a form where i give the json and the translations for the strings in it. I followed the steps given [here](https://github.com/formio/angular-formio/wiki/Translations) and it works.

But i want to do the same to the builder , which is rendered by this code `<form-builder [form]="{
  components: []
}"></form-builder>`

To be specific, examples of strings i want to give my translations are: 

'Drag and Drop a form component'
.
'Basic Components' (tabs on the left)
'Text Field'
.
'Display' (In edit dialog)
'Validation'
'Logic'
.
'Custom Default Value'
'Calculated Value'

....etc

basically for all the components under basic, advanced, layout and data
also for the input fields (like 'label','Description','Calculated Value','Custom Default Value') in the edit dialog box under display, data, validation, api, conditional, logic tabs for every component.

How to achieve this ?

randallknutson:
These are supposed to be able to be translated the same as other values. However, I took a quick look and it appears we are missing the translation wrapper around some of these strings. For example, the "Drag and Drop a form component" is located here: https://github.com/formio/formio.js/blob/master/src/templates/bootstrap/builderPlaceholder/form.ejs

and should be wrapped in `{{t('Drag and Drop a form component')}}` so that it can be translatable. To fix this, we can either add these wrappers where they are needed or you can override the templates. You can see https://github.com/formio/formio.js/wiki/Form-Templating for how to do this. 

The Items like Display, Validation and others like that should already be translatable the same way as Forms.

Lastik:
How about DateTimeComponent (form renderer and viewer)? I was unable to figure out how to translate it. As I understand It wraps flatpickr, and flatpickr itself already has translation of some (not all) of the string constants. I was unable to make it localised through angular API. Some hint would be great!

Vedhachala:
@randallknutson 

I was able to do localization,
But i am seeing few issues , the contents inside all the code inputs (like in custom default value and caluclate value) are not getting changed.
[...]

Issue #520, opened 2018-06-12
Translation issue for File Component label "Drop files to attach, or"


airarrazaval:
**CONTEXT**

I am using Form.io to store project's translations using a Translations resource.  The structure of the resource is the following:

```json
{
  "components": [{
      "input": true,
      "inputType": "text",
      "label": "Label",
      "key": "label",
      "type": "textfield"
    },
    {
      "input": true,
      "inputType": "text",
      "label": "English",
      "key": "en",
      "type": "textfield"
    },
    {
      "input": true,
      "inputType": "text",
      "label": "French",
      "key": "fr",
      "type": "textfield"
    },
    {
      "input": true,
      "inputType": "text",
      "label": "Spanish",
      "key": "es",
      "type": "textfield"
    },
    {
      "input": true,
      "inputType": "text",
      "label": "Arabic",
      "key": "ar",
      "type": "textfield"
    },
    {
      "input": true,
      "inputType": "text",
      "label": "Chinese",
      "key": "zn",
      "type": "textfield"
    },
    {
      "input": true,
      "inputType": "text",
      "label": "Russian",
      "key": "ru",
      "type": "textfield"
    }
  ]
}
```

Where `label` is the label we want to translate.  The other components maps to a specific language code (defined by their key) of the label's translation.  

**ISSUE**

When the File Component is rendered it builds the following text node:

https://github.com/formio/formio.js/blob/76eb26435690ff361f035d6da81c11010c0bfae1/src/components/file/File.js#L271

As you can see, the text node appends extra spaces to both sides of the label `' Drop files to attach, or '`.  Because Form.io trims all inputs, this specific label cannot be translated.  Instead of using spaces in the text node, the same can be achieved with CSS.

**PROPOSED FIX**

Change

https://github.com/formio/formio.js/blob/388e6e4678e143418b1ee319415071ec090e9649/src/sass/formio.form.scss#L204-L207

to

```scss
.formio-component-file .fileSelector .glyphicon, .formio-component-file .fileSelector .fa {
  font-size: 20px;
  margin: 0 5px;
}
```

Change 

https://github.com/formio/formio.js/blob/388e6e4678e143418b1ee319415071ec090e9649/src/sass/formio.form.scss#L209-L211

to

```scss
.formio-component-file .fileSelector .browse {
    cursor: pointer;
    margin: 0 5px;
}
```

change

[...]

PR #1721, opened 2019-09-20
Send user-visible text in templates through i18next.t() function

darabi:
This is related to #520 and #1639. The localisation
strategy using i18next which is described in

https://github.com/formio/formio.js/wiki/Translations

can only work if the text is actually passed to
i18next.t(). This change sends some additional
user-visible strings to i18next.t() for localisation.


travist:
Thank you for this effort!  This will be included in our 4.2.0 build we are working on right now.

PR #1732, opened 2019-09-23
wip: accessibility

dianatsevan:


src/templates/bootstrap/checkbox/form.ejs
```diff
@@ -1,17 +1,19 @@
 <div class="form-check checkbox">
-  <label class="{{ctx.input.labelClass}} form-check-label">
+  <label class="{{ctx.input.labelClass}} form-check-label" id="label-{{ctx.component.key}}">
```
randallknutson: We are trying very hard to stay away from using ids as these forms can be duplicated on pages and if we use the key as the id it will cause errors. In order to link the labels with the inputs, we can use the "for" attribute and set it to the same as the "name" attribute on the input.

dianatsevan: Well, I tried to use the case that you suggested and want to say that it doesn't work without id. w3.org has an example of how it should be https://www.w3.org/TR/WCAG20-TECHS/H44.html But I also tried other techniques to link labels and input. So, aria-label works well for us

phucduong86: The html mark up should be
```
<label for="<value>"
<input id="<value>" type....>
```
And not the "id" attribute in the label.
So as long as we can find an attribute that is unique (maybe options.name since html5 accepts all characters). We can have:
```<label class="col-form-label {{ctx.label.className}} " for="{{ ctx.options.name }}">```
And in the input class 
```
      var attr = {
+        id: this.options.name,
        name: this.options.name,
        type: this.component.inputType || 'text',
        class: 'form-control',
        lang: this.options.language
      };
```

Comment (phucduong86) [reply src/templates/bootstrap/checkbox/form.ejs]: Ideally the radios and checkboxes should be wrapped into a fieldset with legend (https://webaim.org/techniques/forms/controls).
In my custom renderer I've done this to the radio form template:
```
    {% if (ctx.component.type == 'radio') { %}
    <fieldset id="{{ctx.id}}" class="{{ctx.classes}} form-group"{% if (ctx.styles) { %} styles="{{ctx.styles}}"{% } %} ref="component">
      {% if (ctx.visible) { %}
      {{ctx.children}}
      <div ref="messageContainer" class="formio-errors invalid-feedback"></div>
      {% } %}
    </fieldset>
    {% }  else { %}
    <div id="{{ctx.id}}" class="{{ctx.classes}} form-group"{% if (ctx.styles) { %} styles="{{ctx.styles}}"{% } %} ref="component">
      {% if (ctx.visible) { %}
      {{ctx.children}}
      <div ref="messageContainer" class="formio-errors invalid-feedback"></div>
      {% } %}
    </div>
    {% } %}
```
Combined with the change above I haven't really had any issues. However our application only render one form on a page.
What are your opinions on it @dianatsevan @travist @randallknutson

dianatsevan: Thanks for the response @phucduong86 
Both cases were investigated and would be taken into account

PR #1741, opened 2019-09-24
Added isInputComponent method in utils

Alena-Levina:


Alena-Levina:
@travist need to pull in formio.js version with this PR merged https://github.com/formio/formio.js/pull/1741 before merging this PR

PR #1745, opened 2019-09-24
Htmlelement self closing tags support

roflanKisel:


randallknutson:
Is there a reason that this pull request removes "form-group" from the classes?

travist:
> Is there a reason that this pull request removes "form-group" from the classes?

Because those classes are provided in the class method of the input components already.

PR #1770, opened 2019-09-27
updated dirty when edit grid row

RomanLetsuk:


src/components/editgrid/EditGrid.js
```diff
@@ -356,7 +356,7 @@ export default class EditGridComponent extends NestedComponent {
   editRow(rowIndex) {
     const dataValue = this.dataValue || [];
     const editRow = this.editRows[rowIndex];
-    editRow.dirty = false;
```
travist: Will this make errors show up when you create a new row?  It should not, but only show errors when you edit an existing row.

randallknutson: Yeah, this change is wrong. Rows should start out as pristine when opening them.

RomanLetsuk: It creates errors for items that we try to edit. If we don't do that and set dirty = false, we have the next logic:
editRow.components.forEach(comp => {
        comp.setPristine(!isDirty);
        valid &= comp.checkValidity(null, isDirty, editRow.data);
      });
and i did'n find any other place where dirty is setting to true. only when try to save row.

PR #1774, opened 2019-09-28
linking label with input id

phucduong86:
For accessibility improvement.

### DIFF
--- a/src/components/_classes/input/Input.js
+++ b/src/components/_classes/input/Input.js
@@ -22,6 +22,7 @@ export default class Input extends Multivalue {
       name: this.options.name,
       type: this.component.inputType || 'text',
       class: 'form-control',
+      id: this.component.key,
       lang: this.options.language
     };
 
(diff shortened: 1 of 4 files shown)


src/components/_classes/input/Input.js
```diff
@@ -22,6 +22,7 @@ export default class Input extends Multivalue {
       name: this.options.name,
       type: this.component.inputType || 'text',
       class: 'form-control',
+      id: this.component.key,
```
randallknutson: We've tried hard to stay away from using ids. This is because The same form can be on a page multiple times and different forms on the same page could have the same keys for components. The label should work with the "name" attribute of the component, not the id.

phucduong86: Thanks for the response @randallknutson 
Could we then somehow use label "for" with the "name" attribute, and set component's id same as its name?
It is a pretty important requirement for form accessibility to have the the label & input linked up.

phucduong86: Maybe you can use ```id:this.options.name``` and change label for to that id?

randallknutson: You don't need to set the id in order to set the label and input to be linked. If the label "for" is the same as the input "name" it will work correctly. We would definitely accept this as a pull request.

travist: We have a pretty big project working on this right now.  https://github.com/formio/formio.js/pull/1732

travist: I would vote to close this in favor of the changes being made to https://github.com/formio/formio.js/pull/1732

PR #1790, opened 2019-10-02
added columns extra small classes

RomanLetsuk:


randallknutson:
I don't know if this is the correct way to handle this. We should make it so that whatever column.width is set to is the breakpoint.

PR #1803, opened 2019-10-03
Table component style enhancements

roflanKisel:


src/components/table/editForm/Table.edit.display.js
```diff
@@ -80,5 +80,16 @@ export default [
     input: true,
     tooltip: 'Condense the size of the table.',
     weight: 704
-  }
+  },
```
travist: I don't think we need a configuration for this. If they don't want this to happen, they can just override the CSS in their own styles.

travist: I will make this change since I am getting a build together.

PR #1810, opened 2019-10-04
Added cell alignment option to table component.

roflanKisel:


randallknutson:
What is the use case of this? This seems like something better suited to a custom change for a customer. @travist any thoughts on this?

travist:
We talked through this, and I thought it was a legitimate request for new feature for Table components. In the case of the customer, they wanted to have a table of buttons, but them centered within the cells. It made sense to me to allow for a configuration on the Table component which would allow you to provide cell alignment.

Issue #1662, opened 2019-09-11
The input element's type ('email') does not support selectionStart


gregbown:
BUG Using Formio.js v4.0.3-8
Running the app like the demo at https://formio.github.io/formio.js/app/builder
Also verified that your demo throws the same error.

Add an email input to the form
paste an email address into the test rendered form
throws error because it is trying to set selectionStart on an input type email which does not have the property. 

See: [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange), according to the forms spec selectionStart, selectionEnd properties and setSelectionRange method apply only to inputs of types text, search

**Error**
```
formio.full.min.js:26 Uncaught DOMException: Failed to set the 'selectionStart' property on 'HTMLInputElement': The input element's type ('email') does not support selection.
```
**Just thought you should know this bug breaks the ability to use the email input completely. As soon as the error occurs no input is passed to the submission rendering the field useless.**

The fix I used was to check for the property first.  
src\components\_classes\multivalue\Multivalue.js

```
        if (element.selectionStart && element.selectionEnd) {
          element.selectionStart = selectionStart;
          element.selectionEnd = selectionEnd;
        }
```

jhen1422:
@gregbown 
Thank you for the report. We have a created a ticket for the issue and will review during our planning meeting next week. Please note, we welcome pull requests in this library. If you need this item expedited, please feel free to contact support@form.io for pricing.

gregbown:
This appears to be fixed in version 4.2.0-rc.7

gregbown:
This error is back in version 4.2.5 ?

randallknutson:
I'm not sure if we have fixed this yet. It is a valid issue that we need to fix. 

We need to wrap these lines in a check for the input type and not run it for other types. https://github.com/formio/formio.js/blob/2fff6bf0147fca1b7df05af605cd97851ed3c5c1/src/components/_classes/multivalue/Multivalue.js#L114-L115

randallknutson:
https://github.com/formio/formio.js/pull/1812

jhen1422:
This should be resolved in version 4.5.1

sergei-maertens:
Seeing this again in 4.12

AlexByte:
It should be reopened

AlexByte:
It should be reopened, error occurs in Chrome 49

jabcreations:
[...]

PR #1812, opened 2019-10-04
Add check of component type because of selection properties of input

AlexeyDavyd:
Should fix all 3 of them:

- https://trello.com/c/SnvUKJCI/395-email-email-data-is-not-saving
- https://trello.com/c/UNnagGu1/346-email-email-field-is-not-forcing-email-mask
- https://github.com/formio/formio.js/issues/1662

### DIFF
--- a/src/components/_classes/multivalue/Multivalue.js
+++ b/src/components/_classes/multivalue/Multivalue.js
@@ -112,7 +112,7 @@ export default class Multivalue extends Field {
           element.value = element.value.toLowerCase();
         }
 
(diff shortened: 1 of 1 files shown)


randallknutson:
In the original issue the check is for "selectionStart" and "selectionEnd" on the element. Is that a better idea than checking the type? 

```
if (element.selectionStart && element.selectionEnd) {
          element.selectionStart = selectionStart;
          element.selectionEnd = selectionEnd;
        }
```

Issue #1717, opened 2019-09-19
[BUG] data loss - Radio button not keeping value when navigating between pages


CatBakun:
## Environment
Please provide as many details as you can:

- Hosting type
  - [] Form.io
  - [x ] Local deployment
    - Version: 
- Formio.js version: 4.1.0
- Frontend framework: JS
- Browser: Chrome 
- Browser version: 76.0.3809.100 (Official Build) (64-bit)

## Steps to Reproduce
1. In the first page select a radio choice.
2. Navigate to second page by clicking in the page link (not "next" button)
3. Navigate back to page 1

## Expected behavior
The radio I selected in page 1 remains selected

## Observed behavior
The radio I selected is unselected.

Note: this form works on version 3. Other wizard forms work inversion 4.

## Example
`git clone https://github.com/CatBakun/test-formio`
`cd test-formio`
`npm ci` 
`npm run start`

randallknutson:
I tested this with a radio in a wizard and it worked as expected. Can you share a form where this isn't working?

CatBakun:
Hi @randallknutson  I have shared the example in a repo, at the end of this report

CatBakun:
Hi @randallknutson did you get the chance to try out my example and reproduce?
Some more context: This form was created with version 2, and used without this problem in version 3. But when moved to version 4, this form started to show this bug. It was hard to figure out because there is not error log when this happens.
Is there a list of breaking changes in components definitions between version 3 vs 4 or 2 vs 4? Maybe there is something in the form definition breaking the render value set of the components.

jhen1422:
@CatBakun I can confirm this is working on the latest version of our 4x renderer. Can you try on the latest?

CatBakun:
@jhen1422 No luck fter upgrading to 4.1.1, and later to 4.2.0-rc.7. Keep losing radio selections when navigating between pages. Your success try was with my example?
```
git clone https://github.com/CatBakun/test-formio
cd test-formio
npm ci
npm run start
```

ps-from-md:
I have also noticed an issue when using the radio control in a panel.  I will get a fiddle together to see if i can reproduce it.  Simply put, if i have the control on the main area it works fine.  However, if I nest it in a Panel, it does not set the value.

ps-from-md:
[...]

PR #1831, opened 2019-10-08
[BUG] 1717 Fixed issue where Radio control would not initialize with a value if …

ps-from-md:
…contained within a panel.

https://github.com/formio/formio.js/issues/1717


travist:
This causes issues with the SelectBoxes component (which derives from the Radio). We will need to revert this until we can come up with a better solution.

PR #1847, opened 2019-10-10
Bulma

YuriPetyhov:


travist:
Please just include the "src" files changed. We will update the lib folder when we build it.

Also, what use case is this solving? Do you have a jsFiddle of the problem?

travist:
I believe this will be fixed with https://github.com/formio/formio.js/pull/1852

PR #1866, opened 2019-10-14
Fix tab component stores schema of the child components twice

AlexeyDavyd:


src/components/tabs/Tabs.js
```diff
@@ -39,7 +39,7 @@ export default class TabsComponent extends NestedComponent {
     // We need to clone this because the builder uses the "components" reference and this would reset that reference.
     const components = _.cloneDeep(this.component.components);
     schema.components = components.map((tab, index) => {
-      if (index === this.currentTab) {
```
travist: This feels like a bandaid on top of the real problem. The only time that schema is usually called is when the form is in builderMode so I read this as skipping the needed logic for form builder. I would suggest to identify the real problem here.

PR #1873, opened 2019-10-15
Vpat/1 text field screen reader

dianatsevan:


dianatsevan:
[Text Field Screen Reader](https://trello.com/c/NHZK8pY9/1-text-field-screen-reader)

Issue #1734, opened 2019-09-23
Customg Template (bulma.io CSS)


unrealsolver:
Hello. Due to project needs I have to create a template for Bulma CSS Framework (https://bulma.io/) which we currently using.
Do anyone know any handy resources to look at?

heather-formio:
https://github.com/formio/bulma

PR #1875, opened 2019-10-15
Integrated Bulma

YuriPetyhov:


travist:
This is a great contribution! We may need to figure out how to not include this by default and have it so that it could be included in your application by importing the template directly and then assigning it. Give us some moments to figure out how this could work and write some documentation around it, and then we would be happy to include this template.

unrealsolver:
Fixes: https://github.com/formio/formio.js/issues/1734
@travist , this implementation is pretty much basic, but works fine for my form (150+ fields).
Me and @YuriPetyhov  going to keep working on it.

randallknutson:
We're discussing having some form of plugin architecture to easily add plugins to the renderer such as additional templates or custom components. We can't keep adding new css frameworks and components to the core renderer so we need to find a way to start breaking things out and making them optional. We would like to do that with this framework as well as some others that are in the works.

aminmc:
@travist @randallknutson it would be interesting to see how this would work. we have a single template file for a client which I would like to break out into the same structure as the other templates. I would like to publish these templates so other can consume. It would be great to able to add the template from some other npm registry and 'plugin' and register with formio.

randallknutson:
Yeah, it should support something like that. We're at the very preliminary stages of planning of this but will probably need to move fast on getting it implemented.

randallknutson:
Code is being developed here: https://github.com/formio/formio.js/pull/1916

Plugins are modules that can define their own templates, components and fetch interceptors. We're getting together an example for semantic for how we want to break it out. We haven't fully tested everything yet but this is how we want it to work.

Example for semantic: https://github.com/formio/semantic

randallknutson:
Please test this out and see if it works for you:
https://github.com/formio/bulma
https://www.npmjs.com/package/@formio/bulma

These won't work until this pull request is pulled into formio.js and a build is released.
https://github.com/formio/formio.js/pull/1939

Thanks for the contribution! If you need to make any changes, just open a pull request to the formio/bulma repo and we'll pull it in and build it.

travist:
We are going to close this Pull request in favor or using the repo we just created @ https://github.com/formio/bulma

PR #1876, opened 2019-10-15
Fixed sidebar styles in form builder.

roflanKisel:


src/sass/formio.form.builder.scss
```diff
@@ -168,3 +168,7 @@
 .help-block {
   margin: 0px;
 }
+
+.btn {
```
travist: Can we make this more specific?.... this should only affect buttons in the sidebar.

travist: easy fix... I will fix it.

PR #1882, opened 2019-10-16
updated disable setting for select component

RomanLetsuk:


src/components/select/Select.js
```diff
@@ -895,7 +899,7 @@ export default class SelectComponent extends Field {
     }
 
     // Force the disabled state with getters and setters.
-    this.disabled = this.disabled;
+    this.disabled = this.shouldDisabled;
```
randallknutson: I'm not sure this is correct yet. It will still get stuck if the parent is enabled. We need to move the setting of classes out of the disabled setter and call it in the attach method.

randallknutson: After checking more, I see that it will set the disabled correctly based on this even if the parent is disabled so this should be good.

PR #1886, opened 2019-10-16
Fix tab component stores schema of the child components twice

AlexeyDavyd:


travist:
This still feels broken. What happens if you switch between multiple tabs and make changes in the builder. This seems like the only changes that will be recorded are the ones that are in the "currentTab"

PR #1898, opened 2019-10-17
Added ability to deselect radio value.

roflanKisel:


travist:
Couple of comments on this one.

- I don't think we need a separate option for this... it should just work like this by default.
- https://github.com/formio/formio.js/pull/1898/files#diff-3a2391b0fdbb6539e390143265601564R69 should be wrapped up in the logic block

src/components/radio/Radio.js
```diff
@@ -8,6 +8,7 @@ export default class RadioComponent extends Field {
       inputType: 'radio',
       label: 'Radio',
       key: 'radio',
+      uncheckable: false,
```
travist: Please do not create a new option for this particular feature.

travist:
This breaks SelectBoxes. https://monosnap.com/file/7PQq8jcOiDY7l8wpALAc82Elli8WMI

PR #1899, opened 2019-10-17
Make tabs schema to take components form current tab only

AlexeyDavyd:


src/components/tabs/Tabs.js
```diff
@@ -40,7 +40,7 @@ export default class TabsComponent extends NestedComponent {
     const components = _.cloneDeep(this.component.components);
     schema.components = components.map((tab, index) => {
       if (index === this.currentTab) {
```
travist: I am not sure if this if statement is needed anymore. It seems that the tabs would not update if you make multiple changes to multiple tabs.

PR #1916, opened 2019-10-18
Add plugin architecture.

randallknutson:


src/Formio.js
```diff
@@ -11,6 +11,8 @@ import _intersection from 'lodash/intersection';
 import _get from 'lodash/get';
 import _cloneDeep from 'lodash/cloneDeep';
 import _defaults from 'lodash/defaults';
+import Templates from './templates/Templates';
```
travist: Now looking at this... This is going to make the "Formio.js" library (which is really for SDK usage) size blow up. We should only include this logic in the formio.form.js libs that already include these paths.

randallknutson: Yup. It won't even build this way since there are circular requires.

bios6574:
13-42 line

сб, 19 окт. 2019 г. в 05:07, Travis Tidwell <notifications@github.com>:

PR #1925, opened 2019-10-21
Fixing issue where the Select template would get overridden in Resource components.

travist:


lane-formio:
<div>Agent Lane Doughtie unlinked Freshdesk ticket <a href="https://formhelp.freshdesk.com/a/tickets/54599">54599</a> from this issue.</div>

PR #1930, opened 2019-10-22
vpat: 20 multiple values mode (text area)

dianatsevan:
- added discernible text for buttons
- implemented dynamic focus addition
- added special notifications for user on rows addition and removal


dianatsevan:
[Multiple value Text Area](https://trello.com/c/Z6M4PpEG/20-multiple-value-text-area)

src/components/_classes/component/Component.js
```diff
@@ -844,7 +844,8 @@ export default class Component extends Element {
 
     this.loadRefs(element, {
       messageContainer: 'single',
-      tooltip: 'multiple'
+      tooltip: 'multiple',
+      label: 'single'
```
YuryRybakSoftensity: Please use trailing commas where possible.

PR #1934, opened 2019-10-22
vpat: 24 tooltip content is not visible on focus

dianatsevan:


dianatsevan:
[Tooltip content is not visible on focus](https://trello.com/c/FUYTZHh9/24-tooltip-content-is-not-visible-on-focus)

PR #1939, opened 2019-10-22
Remove sketchpad and tagpad

randallknutson:


randallknutson:
These are being moved to another repo to be finished

PR #1948, opened 2019-10-24
Fix required validation border is being removed after correcting one field

AlexeyDavyd:


src/Wizard.js
```diff
@@ -318,6 +318,13 @@ export default class Wizard extends Webform {
     });
   }
 
+  checkData(data, flags) {
+    const dirty = this.currentPage.components.some(component => {
+      return !!data[component.component.key];
```
travist: This should probably use the ```isEmpty``` method on the component like this.

```
this.currentPage.components.some(component => !component.isEmpty());
``

File: src/Wizard.js [hunk 2]
```diff
@@ -318,6 +318,11 @@ export default class Wizard extends Webform {
     });
   }
 
+  checkData(data, flags) {
```
travist: This seems like it will cause some problems. I will test it out before including it in a new build.

Issue #1609, opened 2019-08-28
[BUG] : Show/hide based on condition causing wysiwyg editor initialized multiple times


Inaamhusain:
I have tried that [https://formio.github.io/formio.js/app/builder](https://formio.github.io/formio.js/app/builder) here.

Created radio button yes and no. and based on yes/no showing textarea and WYSIWYG editor.
When you switch multiple times it causing multiple time initialized.

Here is my form JSON to replicate,

`{
    "display": "form",
    "components": [
        {
            "label": "Show",
            "optionsLabelPosition": "right",
            "values": [
                {
                    "label": "yes",
                    "value": "yes",
                    "shortcut": ""
                },
                {
                    "label": "no",
                    "value": "no",
                    "shortcut": ""
                }
            ],
            "inline": false,
            "mask": false,
            "tableView": true,
            "alwaysEnabled": false,
            "type": "radio",
            "input": true,
            "key": "show",
            "defaultValue": "",
            "validate": {
                "customMessage": "",
                "json": ""
            },
            "conditional": {
                "show": "",
                "when": "",
                "json": ""
            },
            "customConditional": "",
            "properties": {},
            "reorder": false,
            "encrypted": false,
            "logic": [],
            "attributes": {}
        },
        {
            "label": "Detailed Description",
            "editor": "quill",
            "isUploadEnabled": false,
            "showWordCount": false,
            "showCharCount": false,
            "tableView": true,
            "alwaysEnabled": false,
            "as": "string",
            "wysiwyg": {
                "theme": "snow",
                "placeholder": "",
                "modules": {
                    "clipboard": {
                        "matchVisual": false
                    },
                    "toolbar": [
                        [
                            {
                                "size": [
                                    "small",
                                    false,
                                    "large",
                                    "huge"
                                ]
                            }
                        ],
                        [
                            {
                                "header": [
                                    1,
                                    2,
                                    3,
                                    4,
                                    5,
                                    6,
                                    false
                                ]
                            }
                        ],
                        [
                            {
                                "font": []
                            }
                        ],
                        [
                            "bold",
                            "italic",
                            "underline",
                            "strike",
                            {
                                "script": "sub"
                            },
                            {
                                "script": "super"
                            },
                            "clean"
                        ],
                        [
                            {
                                "color": []
                            },
                            {
                                "background": []
                            }
                        ],
                        [
                            {
                                "list": "ordered"
                            },
                            {
                                "list": "bullet"
                            },
                            {
[...]

PR #1954, opened 2019-10-25
fix(textarea): Submission value doesn't reflect

Inaamhusain:
Bugfix:
WYSIWYG editor Value doesn't reflect when we try to set submission value on change of something.

### DIFF
--- a/src/components/textarea/TextArea.js
+++ b/src/components/textarea/TextArea.js
@@ -521,6 +521,7 @@ export default class TextAreaComponent extends TextFieldComponent {
   }
 
   setValue(value, flags) {
+    this.dataValue = this.defaultValue;
     const skipSetting = _.isEqual(value, this.getValue());
     value = value || '';
     if (this.isPlain) {


travist:
This change would make it so that TextArea updates would always be perceived as "changed" and cause an infinite loop of change events. This will need some work.

shai93:
@travist 

hasChanged always return false because  newValue and  zdataValue always getting same value also in case of skipsetting. If you can provide solution for this that would be helpful

```
var newValue = value === undefined || value === null ? this.getValue() : value;
var changed = newValue !== undefined ? this.hasChanged(newValue, this.dataValue) : false;

   if (changed) {
        this.setWysiwygValue(newValue, skipSetting, function () {
          return _this7.updateOnChange(flags, changed);
        });
    }
```

travist:
Please provide a jsFiddle showing the issue so that we can determine root cause.

shai93:
This issue is based to [1609](https://github.com/formio/formio.js/issues/1609) in which wysiwyg initialized multiple times based on condition which got resolved after upgrading version to 4.0.0-rc.4 answered by @jhen1422.

But now we are facing new issue that onchange event submission object not updating wysiwyg editor value which is working for text-area component but not for wysiwyg editor

```onChange(event) {
    if (event.data) {
        submission.data.template = event.data["htmlfile"];
        this.formobj = submission;
    }
 }
```

So when i trying to update wysiwyg value on change event using submission object its not working.

Maybe wysiwyg editor not getting refresh or reset after updating its value on change event

This was working for version 3.10.2 but it was causing me issue that was mention in [1609](https://github.com/formio/formio.js/issues/1609).

Use this [jsfiddle](https://jsfiddle.net/k0bonc17/1) and upload any file.

shai93:
@travist  did you get a chance to look into it?

randallknutson:
This is still on our list of items to be fixed

travist:
The next release will include a refactored wysiwyg functionality. Please test this version and close this if the problem is resolved.

PR #1967, opened 2019-10-29
Fix Textarea not displaying on read only mode.

edwinanciani:


src/components/textarea/TextArea.js
```diff
@@ -525,6 +525,9 @@ export default class TextAreaComponent extends TextFieldComponent {
     value = value || '';
     if (this.isPlain) {
       value = Array.isArray(value) ? value.map((val) => this.setConvertedValue(val)) : this.setConvertedValue(value);
+      if (this.disabled || this.options.readonly) {
```
travist: I still believe this will trigger infinite loops. This should only happen if the value changes. The ```setValue``` method returns if the value changes so I think this code would be better.  **Also:  I believe the variable is called "readOnly" and not "readonly" so the camel case is going to get you there.**

```
const changed = super.setValue(value, flags);
if (changed && (this.disabled || this.options.readOnly) {
  this.triggerRedraw();
}
return changed;
```

travist: But please test this... don't just copy what put in the example above and assume that is right... actually set break points to determine that it functions the way it should.

PR #1968, opened 2019-10-29
Fix blank field in read only mode for textfield calendar

AlexeyDavyd:


travist:
I am refactoring the widget system seen @ https://github.com/formio/formio.js/pull/1981. Please test this bug with that branch and let me know if that fixes this problem.

PR #2001, opened 2019-11-04
Feat(TextField, TextArea): add spellcheck option

vlad-shusterman:


src/components/textfield/TextField.js
```diff
@@ -41,9 +41,7 @@ export default class TextFieldComponent extends Input {
     const info = super.inputInfo;
     info.type = 'input';
 
-    if (this.component.hasOwnProperty('spellcheck')) {
-      info.attr.spellcheck = this.component.spellcheck;
-    }
+    info.attr.spellcheck = this.component.spellcheck;
```
randallknutson: Is there a reason for this change? It seems like if the option isn't set it will cause the spellcheck to be set which is incorrect. It should NOT set on the input if it isn't set on the settings like with old forms.

PR #2021, opened 2019-11-06
Fixed focus shifting when typing in fields with Field Logic

Alena-Levina:


jerocon:
Hi, if a Field Logic has two defined actions, for example, to disable the field and remove the required one, the problem of loss of focus when typing in fields continues to occur

Alena-Levina:
@jerocon thanks for the notice! I fixed that issue in new PR: https://github.com/formio/formio.js/pull/2037

PR #2031, opened 2019-11-08
is breadcrumb clickable moved to top nav instead of buttons

aminmc:
Moved the check if breadcrumb is clickable to the top nav instead of the buttons (next, previous). The option breadcrumbSettings.clickable doesn't related to the buttons (in my opinion). If you set the value as true then the top breadcrumb header is clickable but the buttons became un-clickable. 
Also passing the option into the template context so people can override the look at feel of the breadcrumb with that option

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -201,15 +201,15 @@ export default class Wizard extends Webform {
   }
 
   attachHeader() {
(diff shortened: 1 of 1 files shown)


travist:
Yeah this was clearly a mistake.  Thanks.

PR #2032, opened 2019-11-08
Add label with red asterisk when label is hidden

AlexeyDavyd:


randallknutson:
Does this need to be fixed in bootstrap3 and semantic as well?

Please run "npm run test:updateRenders" to fix the tests.

travist:
@AlexeyDavyd please fix in these repos as well.
 - https://github.com/formio/bulma

PR #2045, opened 2019-11-13
Show default value in preview 

ihildebrandt:
This updates the preview component when editing and in the form builder so that the default value appears. 

Setting the dataValue to default rather than explicitly empty ensures that the field always shows the default value.

### DIFF
--- a/src/WebformBuilder.js
+++ b/src/WebformBuilder.js
@@ -878,6 +878,7 @@ export default class WebformBuilder extends Component {
   }
 
   updateComponent(component, changed) {
+    const defaultValueComponent = getComponent(this.editForm.components, 'defaultValue');
     // Update the preview.
     if (this.preview) {
       this.preview.form = { components: [_.omit(component, [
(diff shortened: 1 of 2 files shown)


ihildebrandt:
@travist Is there a more preferable way to handle this? I assume that the change to the dataValue method in Component.js is risky, but I couldn't find any other way to ensure that the preview component always showed the defaultValue.

travist:
This pull request is too intrusive. We cannot change the base component to use defaultValue instead of emptyValue.

JRiggenbach:
@travist Is there a proven way to get the default values to show in the Preview view of the component as they are being entered?  I've had only marginal success to get the default values to appear in the preview, more specifically with a custom component.  It seems to be just a "one and done" type refresh.  I keep thinking that I'm losing a reference to a my components model, but can't seem to prove that.  I've seen a few other posts around default values and the preview view so perhaps there is an actual reason why I can't get this to consistently update in realtime.

image

After I emit my components value once, the preview will not refresh on any subsequent emits of an updated component value.

Thanks

JRiggenbach:
@ihildebrandt,
If you don't mind me asking, did you end up making any changes for find a work around for the preview not updating?
Thanks...

travist:
Duplicate of https://github.com/formio/angular-formio/issues/591

PR #2048, opened 2019-11-13
Fixed tooltips for components inside datagrid.

roflanKisel:


src/templates/bootstrap/datagrid/form.ejs
```diff
@@ -10,7 +10,7 @@
       {% ctx.columns.forEach(function(col) { %}
         <th class="{{col.validate && col.validate.required ? 'field-required' : ''}}">
           {{ col.hideLabel ? '' : ctx.t(col.label || col.title) }}
-          {% if (col.tooltip) { %} <i ref="tooltip" class="{{ctx.iconClass('question-sign')}} text-muted"></i>{% } %}
```
travist: Please also fix the other templates that we have.

PR #2049, opened 2019-11-13
Fix github #1953: Override .defaultValue of subcomponents.

### LINKED ISSUE #1953: [BUG] Conditional visibility broken when depending on custom default value of nested form
## Environment
- Hosting type
  - [x] Form.io
- Formio.js version: 4.5.0
- Frontend framework: react-formio
- Browser: firefox 
- Browser version: 70.0

## Example on next.form.io:
* [main (= outer) form](https://api.form.io/project/5d1c9ef99cf71913daa0e494/form/5d1e273921470951960b2839)
* [nested (=inner) form](https://api.form.io/project/5d1c9ef99cf71913daa0e494/form/5d1c9f229cf719640ca0e4a6)

## Steps to Reproduce
1. Create an (inner) form with a checkbox component.
2. Add another component, conditionally visible when the checkbox is true.
3. Create an (outer) form with the above form as nested (sub-)form.
4. In that (nested-)form component, specify the custom default value for the inner form to be `{data: {apiKeyOfCheckbox: true}}`
5. Use the main (outer) form to enter data.

## Expected behavior
As the checkbox is correctly checked right from the beginning, the conditionally visible component should be visible from the start as well.

## Observed behavior
The checkbox is correctly displayed as set to true, but the conditionally visible component is still hidden. At some point, that component suddenly becomes visible, for example while typing into an unrelated textfield nearby.

When using custom JS to implement the 'Conditional' behavior instead of simple logic, one can observe that the visibility computation is indeed done, with the correct result in the final round. But that result is not reflected in the display on-screen, until something causes a refresh somehow.

formioBug

franke-mc:
Not sure about side-effects, but this appears to fix the problematic behavior.  Fixes #1953 .

### DIFF
--- a/src/components/form/Form.js
+++ b/src/components/form/Form.js
@@ -282,7 +282,10 @@ export default class FormComponent extends Component {
       }
 
       // Iterate through every component and hide the submit button.
+      // Override defaultValue with respective parts from old dataValue.
+      const oldData = this.dataValue ? this.dataValue.data : {};
       eachComponent(form.components, (component) => {
+        component.defaultValue = _.get(oldData, component.key, component.defaultValue);
         if (
           (component.type === 'button') &&
           ((component.action === 'submit') || !component.action)


travist:
We will include this for the next "minor" release which will go through a few "beta" releases to ensure there are no side effects to this. On the surface it looks fine.

franke-mc:
While the toy testcase example described here appears to work now, the general case involving more complex forms seems to require a more careful `customConditional` code which avoids setting `show` until the flag checkbox value is actually present in the data:
```
const flagComponentKey = 'showNumber';
if (data.hasOwnProperty(flagComponentKey)) {
    show = !!data[flagComponentKey];
}
```
To me, it seems unintuitive that this should be necessary, though.

franke-mc:
@travist: This needs to be merged.

travist:
I was in a debug session with a customer today and they had a very strange issue where they were unable to reset the subForm data. I am afraid this was the cause of that problem, so it does look like this has some pretty negative side effects and will need to be reverted. We will need to fix this issue in a different way.

PR #2052, opened 2019-11-14
Move displays and builders to their own registries.

randallknutson:


src/builders/Builders.js
```diff
@@ -0,0 +1,28 @@
+import _ from 'lodash';
+import pdf from './pdf/PDFBuilder';
```
travist: If we want to pull this into 4.x, then we cannot move file paths (since a lot of people reference the files directly to extend them). We can do this in the 5.x branch, but I would prefer to just have this file reference the builders against the paths they currently are. I know it is messy, but we will clean up the paths in 5.x.

randallknutson: Makes sense

randallknutson:
Manually merged into 5.x

PR #2053, opened 2019-11-15
vpat: 21 text area minimum length validation message is not announced by jaws

dianatsevan:


dianatsevan:
[Text Area Minimum Length validation message is not announced by JAWS](https://trello.com/c/juoDo1CZ/21-text-area-minimum-length-validation-message-is-not-announced-by-jaws)

PR #2056, opened 2019-11-15
Make builders and displays extendable.

randallknutson:


randallknutson:
Failed test is the wrong bootstrap version. Should be fine once updated.

PR #2062, opened 2019-11-18
Fix component redraw inside datagrid, editgrid

roflanKisel:


src/components/_classes/component/Component.js
```diff
@@ -534,6 +536,30 @@ export default class Component extends Element {
     return this.options.attachMode === 'builder';
   }
 
+  get calculatedPath() {
```
travist: This concerns me.  I will need to stare at this pull request to determine what it is actually fixing before we can commit this.

PR #2078, opened 2019-11-20
Fixed placeholder on initial render in select component.

roflanKisel:


travist:
Pretty sure this broke placeholders for "multiple" value selects.

https://monosnap.com/file/KNej4eomo9mxosqMkO4ZfAVbzI931e

PR #2092, opened 2019-11-21
Fix radio thinking false is true.

randallknutson:


src/components/radio/Radio.js
```diff
@@ -183,7 +183,7 @@ export default class RadioComponent extends Field {
         }
         break;
       case 'boolean':
-        value = !!value;
+        value = value.toString() !== 'false';
```
travist: Wouldn't this make value = '' be "true"? Shouldn't it be...

```
value = !(!value || value.toString() === 'false');
```

PR #2093, opened 2019-11-21
Bugfix/hidden margin

randallknutson:


randallknutson:
Accidentally included changes from another pull request but should still be safe to merge.

travist:
Well, there was a problem with previous pull request, so this will need to be pulled out?

PR #2096, opened 2019-11-21
Fix clear on hide not working since root is never set to pristine.

randallknutson:


Dev-Trilok:
@randallknutson @travist It seems that when fields are hidden, their values are cleared as expected, but they are not included in the form submission. Is there any way to include these hidden fields in the submission with a default empty value? For me this would ensure that the form submission accurately represents the complete state of the form, even for hidden components.
I would greatly appreciate any assistance or guidance on this . Thank you.

PR #2098, opened 2019-11-22
Allow translation of file error messages.

randallknutson:


src/components/file/File.js
```diff
@@ -475,33 +475,33 @@ export default class FileComponent extends Field {
...
           status: 'info',
-          message: 'Starting upload'
+          message: this.t('Starting upload'),
         };
 
         // Check file pattern
         if (this.component.filePattern && !this.validatePattern(file, this.component.filePattern)) {
           fileUpload.status = 'error';
-          fileUpload.message = `File is the wrong type; it must be ${this.component.filePattern}`;
+          fileUpload.message = this.t(`File is the wrong type; it must be ${this.component.filePattern}`);
```
travist: This should be this instead.

```
this.t('File is the wrong type; it must be {{ pattern }}', {
  pattern: this.component.filePattern
});
```

PR #2111, opened 2019-11-25
Allow attributes to be passed to modal creation in the form builder

ihildebrandt:
Passing attributes to the modal allows for CSS namespacing on the modal div. Necessary for including formio in a site which doesn't use bootstrap as the CSS framework.


travist:
Thanks @ihildebrandt! We will include this in the next build.

PR #2112, opened 2019-11-26
Update logic to get namespace

RomanLetsuk:


travist:
I don't think we can commit this... it would certainly break a lot of things since it changes behavior of a widely used method. Can you explain why you needed this?

RomanLetsuk:
@travist I've already discussed it with Randall, but the main reason is that line
https://github.com/formio/formio.js/blob/master/src/WebformBuilder.js#L520
The definition doesn't have a parent, the only instance has it.

travist:
@randallknutson you approve this one?  Seems pretty disruptive.

travist:
@RomanLetsuk I don't think we will be able to take this. I am going to revert it. After investigating what you are doing here, it seems that you are trying to fix component references (used in logic) so that their key can find the component no matter their path. The issue is that our builder is designed to handle same keys as long as they are in different paths. For example, it is OK to have "children.firstName" and "firstName" since they are in different contexts. I believe the real solution here should be to store the full path of the component in the logic settings instead of storing just the component key.

PR #2113, opened 2019-11-26
Allow editor source url and settings to be provided in form options

ihildebrandt:
This allows a user to use the form options to set the source url and default settings for third-party editors.


travist:
@ihildebrandt thanks again!

PR #2116, opened 2019-11-27
Fix issue with hidden checkbox values not being updated correctly.

### LINKED ISSUE #2067: [BUG] Conditional dependent on hidden checkbox is broken
## Environment
- next.form.io current

## Steps to Reproduce
0. create a form on next.form.io.
1. add a hidden text field 'text' with custom default value 'foo'
2. add a hidden checkbox 'flag' with calculated value (data.text === 'foo')
3. turn OFF 'clear value when hidden' on both hidden components created above
3. add a number field 'number' with visibility conditional on 'data.flag'

## Expected behavior
- 'text' is 'foo', 'flag' is true, and 'number' is visible

## Observed behavior
- 'text' is 'foo', 'flag' is assigned true, but 'number' does not see that and is not visible

## Example
See [this JSFiddle](https://jsfiddle.net/franke_mc/tznoh3sk/66) or [this form](https://api.form.io/project/5dd3ca7afcf37f3bf6fa513b/form/5dd3ca998b6323e0d941a192/) on form.io

Debug log in JS console:

> text :=  foo
> number: data.flag =  false
> number.show :=  false
> flag :=  true
> number: data.flag =  false
> number.show :=  false
> number: data.flag =  false
> number.show :=  false
> flag :=  true
> number: data.flag =  false
> number.show :=  false
> flag :=  true
> number: data.flag =  false
> number.show :=  false

Note: The example above works fine when then 'flag' checkbox is not hidden. In that case, the log reads:
> text :=  foo
> number: data.flag =  false
> number.show :=  false
> flag :=  true
> number: data.flag =  true
> number.show :=  true
> number: data.flag =  true
> number.show :=  true
> number: data.flag =  true
> number.show :=  true
> flag :=  true
> number: data.flag =  true
> number.show :=  true
> number: data.flag =  true
> number.show :=  true
> flag :=  true
> number: data.flag =  true
> number.show :=  true
> number: data.flag =  true
> number.show :=  true

franke-mc:
Fixes #2067 .

### DIFF
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -181,7 +181,7 @@ export default class CheckBoxComponent extends Field {
 
   setValue(value, flags) {
     flags = flags || {};
-    if (this.setCheckedState(value) !== undefined) {
+    if (this.setCheckedState(value) !== undefined || !this.component.clearOnHide) {
       return this.updateValue(value, flags);
     }
     return false;


travist:
I think the real issue here is that CheckboxComponent overrides the ```set dataValue``` method and does not use the following protections around setting the value.

https://github.com/formio/formio.js/blob/master/src/components/_classes/component/Component.js#L1819

I am going to commit my own version of this fix and provide you attribution.

Thanks.

travist:
This should work. 

https://github.com/formio/formio.js/commit/3e778449f2298248a9ef887d0f531630dba13341

franke-mc:
Are you sure? I'm not saying that your change is wrong, but how does this relate to the analysis in the issue?
I can't see how the committed change can fix the issue:
- I think that in the issue scenario, the checkbox component *does* have a key, but `clearOnHide` is `false`, so the added code should be a no-op.
- `this.dataValue = newValue;` happens in [`updateComponentValue` in Component.js](https://github.com/formio/formio.js/blob/c52c7a4232df32d3f30c47314ceb0af130d49d5e/src/components/_classes/component/Component.js#L2017), called from [`updateValue` in Component.js](https://github.com/formio/formio.js/blob/c52c7a4232df32d3f30c47314ceb0af130d49d5e/src/components/_classes/component/Component.js#L2035), called from [`updateValue` in Input.js](https://github.com/formio/formio.js/blob/c52c7a4232df32d3f30c47314ceb0af130d49d5e/src/components/_classes/input/Input.js#L170), which *NOT* being called from [`setValue` in Checkbox.js](https://github.com/formio/formio.js/blob/master/src/components/checkbox/Checkbox.js#L188) seems to be the problem here.

At the very least, the new commit needs some testing...

travist:
I looked at your fix and there were some issues with it. First of all... your code would have updated the value if it was "undefined" but clearOnHide was set which breaks the original logic. The other thing I noticed is that it seemed you were trying to protect against the value getting set when "clearOnHide" was set, but you were not checking if the checkbox was actually hidden or not. I then asked myself why this works for other components and not Checkboxes. I found what seemed to be an obvious reason. Checkbox's override the "set dataValue" method and do not provide the base classes protection against "clearOnHide" setting the value. Regardless if this fixes your specific bug or not, this was a glaring deficiency so I fixed it. I have not determined if it fixes this specific problem or not, but I would bet money that it does.

franke-mc:
Ok. I have a different theory: For the normal case of value not being undefined, the call to setCheckedValue() appears to return undefined iff there is no input (~ checkbox is hidden). Which means the original condition protects checkbox values from being updated when hidden. So basically I changed the condition from `!hidden` to `!hidden || !clearOnHide`, enabling updates to hidden checkboxes where necessary.

travist:
So it sounds like my "fix" did not resolve your use case? If not, then please make an updated pull request and we will re-evaluate it. Thanks.

PR #2120, opened 2019-11-27
Interpolate & translate labels.

### LINKED ISSUE #2117: [4.x] Interpolate component labels?
Would it be possible to allow templates for field labels, too? Or is it by design that this feature works in tooltips and html element contents, but not in labels?

franke-mc:
Fixes #2117 . Needs review & testing.

In an earlier version of this patch, this seemed to work well (tested with ordinary Text Field component labels):
testLabelTemplate

### DIFF
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -603,6 +603,19 @@ export default class Component extends Element {
   }
 
   /**
+   * Interpolate a template with data & row made available.
+   *
+   * @param {string} template - The template to interpolate.
+   */
+  i(template) {
+    const ctx = {
+      data: this.rootValue,
+      row:  this.data
+    };
+    return this.interpolate(template, ctx);
+  }
+
+  /**
    * Translate a text using the i18n system.
    *
    * @param {string} text - The i18n identifier.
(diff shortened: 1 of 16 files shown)


randallknutson:
The translate function already has params built in that can be used for substitution. See https://github.com/formio/formio.js/blob/master/src/components/_classes/component/Component.js#L611

Does that work for you instead of full on interpolation after translation?

franke-mc:
I'm not sure. My use case for wanting interpolation on ordinary labels is form development in a way that tries to avoid duplication of 'shared' parts of larger forms, thus implementing these 'shared' parts as subforms. But then there are occasions where tooltips or labels have to differ slightly depending on which form they are used in (= some kind of context, e.g. passed in via custom default value). For tooltips interpolation solves this quite well, so I thought that might be the best solution for labels as well. A typical tooltip is `{{MyUtils.selectByContext(data.context, {
case1: '...',
case2: '...'
})}}`

travist:
Closing due to inactivity. We didn't accept this since it was not entirely clear the need for this addition. If you can work up an example case in JSFiddle on why our current interpolation is not sufficient, we will be happy to revisit this addition.

franke-mc:
@travist : I don't have high stakes in this, and I can understand why interpolation might be overkill, but the need for translation of labels should be self-evident for any multi-lingual app. Am I missing something here?

franke-mc:
Ok, I can see now that the only case where I added translation to labels appears to have been checkboxes...

PR #2129, opened 2019-11-29
Fixed Wizard checkData validation order

airarrazaval:
Currently wizard checkData function performs validation checks before checking data, thus failing even when calculateValue is defined.  The order should be reversed and first check the component wizard data (thus triggering data evaluations) and then check the data's validity.

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -425,7 +425,7 @@ export default class Wizard extends Webform {
 
   checkData(data, flags) {
     const dirty = this.currentPage.components.some(component => !component.isEmpty());
-    return this.checkValidity(data, dirty, true) && super.checkData(data, flags);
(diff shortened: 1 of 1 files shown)


airarrazaval:
@travist for some reason the change did not merge.

travist:
I see that... that is very strange.  Let me try to pull in the change manually.

travist:
It seems like it cannot find the commit... let me revert and recommit.

travist:
Will you resubmit the change? It seems like your remote branch was deleted or is no longer found. that may have caused the irregular merge.

travist:
I went ahead and pushed it here. https://github.com/formio/formio.js/commit/ce5e77242fef53b94bb598eaa4e444752f68e8c0

Issue #2153, opened 2019-12-05
[BUG] Component id is always regenerated after a change in formbuilder


Romakita:
Hello Team,

Since this the 2019-04-15 (commit "Set unique id for each component"), the component id is always regenerated. It's a problem because in our project we use REST API to generate the front-end app and we use this id to store data in db (instead of the key).

In the code  of Element.js, I found that:

```javascript
    /**
     * The ID of this component. This value is auto-generated when the component is created, but
     * can also be provided from the component.id value passed into the constructor.
     * @type {string}
     */
    this.id = FormioUtils.getRandomComponentId();
```
The comment isn't aligned with the implementation. Originally the code was that:
```javascript
/**
     * The ID of this component. This value is auto-generated when the component is created, but
     * can also be provided from the component.id value passed into the constructor.
     * @type {string}
     */
    this.id = id || FormioUtils.getRandomComponentId();
```

## Expected behavior

- [] The component should not change when the component is already created.

Thanks for your job and your framework ;)

See you 
Romain

franke-mc:
Related: issue #1921

Romakita:
PR https://github.com/formio/formio.js/pull/2154

Romakita:
PR https://github.com/formio/formio.js/pull/2154

sagargg:
some projects required persisted unique keys, it is better if we make it configurable.

brendanbond:
I think the idea here (per [this pull request](https://github.com/formio/formio.js/pull/2154)) is that the `id` property isn't stored with the component definition and therefore should be regenerated each time the form is instantiated (unless the DOM already has an id). In other words, the `id` property is used internally by the renderer and isn't designed to be persisted anywhere else. I"m closing this as a "won't fix," although I'm curious why anyone fetching JSON from the Formio REST API or even from their own REST API can't just key into another property or store their own id.

Romakita:
Yes I totally agree. It was a bad idea to use id for our project ^^ sorry for this mistake!

see you

PR #2154, opened 2019-12-05
fix: Don't regenerate component.id when the component have an existing id

Romakita:
Should resolve the bug #2153

### DIFF
--- a/src/Element.js
+++ b/src/Element.js
@@ -29,7 +29,7 @@ export default class Element {
      * can also be provided from the component.id value passed into the constructor.
      * @type {string}
      */
-    this.id = FormioUtils.getRandomComponentId();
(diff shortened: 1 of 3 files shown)


travist:
I am a little concerned about this one. Normally I would accept this, but the only way this would work is if the "id" was persisted and as far as I know it should not be persisted. This means that this would be prone to be broken in the future.

randallknutson:
Travis is correct. The Id is not stored with the component definition and should be regenerated each time the form is instantiated except if the DOM already has an id. This is handled here: https://github.com/formio/formio.js/blob/master/src/components/_classes/component/Component.js#L869

travist:
won't fix.

PR #2166, opened 2019-12-09
FOR-2564: Add polyfills for Element and CustomEvent

RomanLetsuk:


src/polyfills/CustomEvent.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable */
+(function () {
+
+  if ( typeof window.CustomEvent === "function" ) return false;
+
+  function CustomEvent ( event, params ) {
```
travist: I would prefer to just use https://github.com/kumarharsh/custom-event-polyfill since it seems more robust and then just change this to...

```
import 'custom-event-polyfill';
```

For example, this will now break all Node.js inclusions since it doesn't check for empty window object.

src/polyfills/ElementPolyfill.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable */
+if (!Element.prototype.matches) {
```
travist: I'm ok with this just being written out, but there should at least be a link saying where this code came from... i.e. https://github.com/idmadj/element-closest-polyfill/blob/master/index.js

RomanLetsuk: @travist I got that from https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

travist: I meant in the actual code. Like in the comments have that link so when we run across it we know where it came from.

PR #2167, opened 2019-12-09
roll back flatpickr version to 4.6.2

RomanLetsuk:


travist:
Why?  We need a commit message or something explaining why we need to do this so that when I go an upgrade again I can be aware of why we downgraded.

RomanLetsuk:
@travist we need to roll back because version 6.4.3 has an issue that doesn't allow to clear input for the DateTime component.

travist:
You need to actually "pin" the version then...  ```^4.6.2``` will still pull in 4.6.3.

travist:
I'll pin it.

PR #2169, opened 2019-12-09
Fix (Component): bug with validity on default value

vlad-shusterman:


src/components/_classes/component/Component.js
```diff
@@ -2279,7 +2279,9 @@ export default class Component extends Element {
     }
     this.calculateComponentValue(data, flags, row);
     this.checkComponentConditions(data, flags, row);
-    return flags.noValidate ? true : this.checkComponentValidity(data, false, row);
+    const shouldCheckValidity = !this.builderMode && !this.options.preview && this.defaultValue;
+    const isValid =  shouldCheckValidity ? this.checkComponentValidity(data) : true;
```
travist: Why are we performing two ```checkComponentValidity``` checks now?

vlad-shusterman: What do you mean?
I have only one ```checkComponentValidity``` in line 2283.

vlad-shusterman: Did some performance improvement.

Issue #2024, opened 2019-11-07
[BUG] Using DataMap within an edit grid loses the data on Edit mode. 


sunshinesanty:
## Environment
Please provide as many details as you can:
Node: 10.10.0

- Hosting type
  - [] Form.io
  - [x] Local deployment
    - Version: latest.
- Formio.js version: 4.6.2
- Frontend framework: React 16.9.0
- Browser: All (Chrome, Firefox, IE 11, Edge, Safari)
- Browser version: Latest and -2 version, except IE 11

## Steps to Reproduce
1. Add a `DataMap` component in to an `Edit grid` (as form or wizard)
2. Render the form and select `Add Another`, to add an Edit grid entry. 
3. Enter the key value pair in `DataMap` component.
4. Save the iteration for `Edit grid`.
5. Edit the saved value.

## Expected behavior
The `DataMap` component should load with the correct number of Iterations with teh previously entered data.

## Observed behavior
The `DataMap` component loads with the correct number of Iterations but the data is lost.

## Example
https://codepen.io/sunsghinesanty/pen/QWWrvQv 

For code or form JSON, please enclose in a code block:

```js

var formElement = document.getElementById('formio');

var builder = new Formio.FormBuilder(document.getElementById("builder"), {
  display: 'form',
  components: [
        {
            "label": "Edit Grid",
            "disableAddingRemovingRows": false,
            "modal": false,
            "key": "editGrid",
            "type": "editgrid",
            "input": true,
            "components": [
                {
                    "label": "Data Map",
                    "key": "dataMap",
                    "type": "datamap",
                    "input": true,
                    "valueComponent": {
                        "type": "textfield",
                        "key": "key",
                        "label": "Value",
                        "input": true,
                        "hideLabel": true
                    }
                }
            ],
            "path": "editGrid"
        },
        {
            "type": "button",
            "label": "Submit",
            "key": "submit",
            "disableOnInvalid": true,
            "input": true
        }
    ],
  settings: {
    pdf: {
      "id": "1ec0f8ee-6685-5d98-a847-26f67b67d6f0",
      "src": "https://files.form.io/pdf/5692b91fd1028f01000407e3/file/1ec0f8ee-6685-5d98-a847-26f67b67d6f0"
    }
  }
}, {
  baseUrl: 'https://examples.form.io'
});

var onForm = function(form) {
  form.on('change', function() {
    subJSON.innerHTML = '';
[...]

PR #2200, opened 2019-12-17
Fixed EditGrid not setting Nested Components values on edit

airarrazaval:
Related to https://github.com/formio/formio.js/issues/2024

When restoring a row data context, it sets the context global data. Unfortunately, on nested components it does not sets the values of its components.  This pull request sets the component's value if it is present in the row context, thus correctly initializing nested component's components.


airarrazaval:
Example of this bug here: https://jsfiddle.net/airarrazaval/ehcfm4v8

travist:
Looks good. Thanks!

PR #2201, opened 2019-12-17
Fix day and select error messages.

randallknutson:


src/components/_classes/component/Component.js
```diff
@@ -2285,7 +2285,7 @@ export default class Component extends Element {
     }
     this.calculateComponentValue(data, flags, row);
     this.checkComponentConditions(data, flags, row);
-    const shouldCheckValidity = !this.builderMode && !this.options.preview && this.defaultValue;
+    const shouldCheckValidity = !this.builderMode && !this.options.preview && this.empty(this.defaultValue);
```
travist: Shouldn't this be ```!this.empty(this.defaultValue)```?

PR #2213, opened 2019-12-20
Fix (NumberComponent): add empty value for reset

vlad-shusterman:


randallknutson:
Unfortunately this isn't going to work. If we set the value to an empty string the server will reject it since it isn't a number.

src/components/_classes/component/Component.js
```diff
@@ -2020,7 +2020,7 @@ export default class Component extends Element {
    */
   updateComponentValue(value, flags) {
     flags = flags || {};
-    let newValue = (value === undefined || value === null) ? this.getValue() : value;
+    let newValue = ((value === undefined || value === null) && this.component.type !== 'number') ? this.getValue() : value;
```
travist: This should just be the NumberComponent overriding ```normalizeValue``` and not code within the Component that is specific to component.type.

vlad-shusterman: @travist Done.

PR #2231, opened 2019-12-27
Implemented Form Controller Execution

AlexeyDavyd:


src/Webform.js
```diff
@@ -862,9 +862,39 @@ export default class Webform extends NestedComponent {
     this.on('deleteSubmission', () => this.deleteSubmission(), true);
     this.on('refreshData', () => this.updateValue(), true);
 
+    this.executeFormController();
```
travist: This should get evaluated only when the form is ready. I think it should be...

```js
return this.formReady.then(() => {
  this.executeFormController();
  return this;
});
```

src/Webform.js
```diff
@@ -862,9 +862,39 @@ export default class Webform extends NestedComponent {
...
+  /**
+   * Performs an evaluation using the evaluation context of this component.
+   *
+   * @param func
+   * @param args
+   * @param ret
+   * @param tokenize
+   * @return {*}
+   */
+  evaluate(func, args, ret, tokenize) {
```
travist: Why do you need to rewrite this method?... Webform derives from Element which has this same method signature.

PR #2235, opened 2019-12-30
Fix (panel) : expand panel when validation errors

vlad-shusterman:


src/components/panel/Panel.js
```diff
@@ -36,6 +37,35 @@ export default class PanelComponent extends NestedComponent {
     return 'panel';
   }
 
+  checkValidity(data, dirty, row) {
+    data = data || this.rootValue;
+    row = row || this.data;
+    return this.checkComponentValidity(data, dirty, row);
+  }
+
+  everyComponent(fn, isPanel) {
```
travist: This would fail if you have other nested components within the Panel since it would not propagate the "isPanel" flag. There has to be a better way here since this seems a little like a hack.

vlad-shusterman: Updated.

PR #2241, opened 2020-01-03
[3.x] Disabled setting in build

RomanLetsuk:


travist:
I believe this was just for PDF but we have migrated over to the 4.x renderer. I am closing this PR.

Issue #2245, opened 2020-01-05
[BUG] Builder is broken in version: 4.8.0-rc.11


gregbown:
- Hosting type
  - [x] Form.io
  - [ x] Local deployment
    - Version: 4.8.0-rc.11
- Formio.js version: 4.8.0-rc.11
- Frontend framework: NodeJS/Express
- Browser: Chrome 
- Browser version:  79.0.3945.88 

## Steps to Reproduce
1. Drag any component from builder menu to layout
2. The component is not reflected in the change event data and is not rendered in preview
3. Rolling back to Formio.js version 4.8.0-rc.10 corrects the problem

## Expected behavior
Form builder should reflect added component in data
## Observed behavior
Form builder is broken/unusable because you cannot build forms
## Example
Version =<4.8.0-rc.10
```
builder change event 
display: "wizard"
components: Array(1)
	0:
		title: "Page 1"
		label: "Page 1"
		type: "panel"
		key: "page1"
		components: Array(1)
			0:
				label: "Text Field"
				labelPosition: "top"
				placeholder: ""
				description: ""
				tooltip: ""
				prefix: ""
				suffix: ""
				widget: {type: "input"}
```
Version >4.8.0-rc.10
```
builder change event 
display: "wizard"
components: Array(1)
	0:
		title: "Page 1"
		label: "Page 1"
		type: "panel"
		key: "page1"
		components: Array(0)
```

jhen1422:
@gregbown Thank you for the report. We are aware of this issue. Please use the previous version while we get this resolved.

travist:
Fixed with 4.8.0-rc.12

Issue #2243, opened 2020-01-04
[BUG] Component settings can't be changed after adding


hosney:
## Environment
Please provide as many details as you can:

- Formio.js version: v4.8.0-rc.11
- Browser: Google Chrome
- Browser version: 79.0.3945.88

## Steps to Reproduce
1.  Add 2 text fields
2. Try to change either of the text field settings (e.g. update label)
3. Try to delete any one of the text fields.

## Expected behavior
Settings should be updated.
Component should be deleted.

## Observed behavior
Settings not updated.
Component not deleted.

## Example
Issue can be reproduced at https://formio.github.io/formio.js/app/builder

jhen1422:
@hosney Thank you for the report. We are aware of this issue and will get this fixed. In the meantime, please use version 4.8.0-rc.10

travist:
Fixed with 4.8.0-rc.12

Issue #2240, opened 2020-01-03
[BUG] (Simple) Conditionals in Formbuilder Have No Form Components Listed


marcus-at-localhost:
## Environment
Please provide as many details as you can:

- Hosting type
  - [x] Form.io
  - [ ] Local deployment
    - Version: 
- Formio.js version: v4.8.0-rc.11 

## Steps to Reproduce
Create some fields
Make one depending on another
Go to tab "Conditional" 
"When the form component:" is empty 

## Expected behavior

List of Components

## Observed behavior

No List

## Example

From the Formbuilder Test Page https://formio.github.io/formio.js/app/builder

![image](https://user-images.githubusercontent.com/1337223/71719305-33223d80-2e1e-11ea-92ec-2ee35280baf2.png)

jhen1422:
@marcus-at-localhost Thank you for the report. We'll get this fixed.

marcus-at-localhost:
Just checked, v4.8.0-rc.10 works fine.

travist:
Fixed with 4.8.0-rc.12

PR #2248, opened 2020-01-07
Fixed issues with form builder elements not working.

travist:


travist:
Fixes https://github.com/formio/formio.js/issues/2245

travist:
Fixes https://github.com/formio/formio.js/issues/2240

travist:
Fixes https://github.com/formio/formio.js/issues/2243

PR #2260, opened 2020-01-13
Fix applying input format

RomanLetsuk:


dianatsevan:
[Number -When 'Require Decimal' setting is configured, field does not auto populate 0s into decimal place when no decimal place is entered](https://trello.com/c/rVjOR4VL/113-number-when-require-decimal-setting-is-configured-field-does-not-auto-populate-0s-into-decimal-place-when-no-decimal-place-is-en)

PR #2264, opened 2020-01-14
vpat: 95 Panel | Collapsible | Focus lost after panel is collapsed/expanded

MaksimFalei:


MaksimFalei:
[Panel | Collapsible | Focus lost after panel is collapsed/expanded](https://trello.com/c/YC4EWod5/95-panel-collapsible-focus-lost-after-panel-is-collapsed-expanded)

PR #2277, opened 2020-01-16
FOR-1763: Normalize date value

RomanLetsuk:


src/components/day/Day.js
```diff
@@ -317,13 +321,39 @@ export default class DayComponent extends Field {
...
+  }
+
   /**
    * Set the value at a specific index.
    *
    * @param index
    * @param value
    */
   setValueAt(index, value) {
+    value = this.normalizeValue(value);
```
travist: This should be getting called in ```updateValue``` which is executed before it gets to this point. It should also handle multiple values with multiple flag which this is getting called within each value. This means that your implementation changes behavior of how the normal normalizeValue should work.  This may need some work.

RomanLetsuk: @travist are you sure that Day Component may resolve multiple values?

PR #2278, opened 2020-01-16
Don't interpolate content when in builder mode so that it won't throw…

randallknutson:
… errors and shows the interpolation options.

### DIFF
--- a/src/components/content/Content.js
+++ b/src/components/content/Content.js
@@ -29,6 +29,9 @@ export default class ContentComponent extends Component {
   }
 
   get content() {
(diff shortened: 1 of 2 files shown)


dianatsevan:
[Number -When 'Require Decimal' setting is configured, field does not auto populate 0s into decimal place when no decimal place is entered](https://trello.com/c/rVjOR4VL/113-number-when-require-decimal-setting-is-configured-field-does-not-auto-populate-0s-into-decimal-place-when-no-decimal-place-is-en)

PR #2304, opened 2020-01-23
Feat(DateTime): add option to disable dates

vlad-shusterman:


src/widgets/CalendarWidget.js
```diff
@@ -179,6 +181,10 @@ export default class CalendarWidget extends InputWidget {
     return this.calendar ? this.calendar.altInput : null;
   }
 
+  get disabledDates() {
+    return this.settings.disabledDates ? this.settings.disabledDates.split(',') : [];
```
travist: The only issue with this is that it is only a matter of time before people request to disable dates by function.

https://flatpickr.js.org/examples/#disabling-dates

Or even providing a range of dates. By doing it this way, we are locking ourselves into a very limited scope of use cases.

I would recommend figuring out how to incorporate all aspects of both disabling and enabling the date ranges.

PR #2329, opened 2020-01-29
Modified the Evaluator so that it could be extended by external systems.

travist:


besmiralia:
Is there an example of how to extend the Evaluator. 
Ex. I want a field to calculate its value from a custom function. How do I inject the function?
```
value = myFunction(data.a, data.b);
```
Thank you

travist:
This can be done with the module system. Create a module like so...

```
{
  options: {
    form: {
      evalContext: {
        validatePhone: function(input) {
          return input.match(/^[2-9]\d{2}-\d{3}-\d{4}$/);
        }
      }
    }
  }
}
```

besmiralia:
Hi @travist 
Thank you for the example. I noticed that if I use a module, then I need to pass options (at least a simple blank object {}) to the createForm method otherwise it will fail to render. The form is initialized through an url.
The issue is at this line [#L28](https://github.com/formio/formio.js/blob/2fbeb878c2710bb974cb159c3556f3e6dbcb2e1f/src/Form.js#L28) when options is undefined.

randallknutson:
This might be a bug where the module system doesn't properly initialize the options if it isn't present. You shouldn't need to pass a blank option to the form

wag110894:
@besmiralia We have created a ticket for this and have added this to our backlog. We welcome a pull request or have a process to expedite if you are interested. 

For reference, the ticket number is FJS-1185.

Please let us know if you are interested in expediting and we can get the process going.

Thank you.

besmiralia:
Hi @wag110894 
Thank you for your feedback.

In this case, I think a simple fix like the below at [line 28](https://github.com/formio/formio.js/blob/2fbeb878c2710bb974cb159c3556f3e6dbcb2e1f/src/Form.js#L28) would do it:
```
if (Formio.options && Formio.options.form) {
   options = Object.assign(options || {}, Formio.options.form);
}
```

PR #2330, opened 2020-01-29
Add label alignment support

RomanLetsuk:


src/components/_classes/field/Field.js
```diff
@@ -5,6 +5,13 @@ export default class Field extends Component {
     if (this.noField) {
       return super.render(element);
     }
+    else if (this.isLegacyLabel && this.options.pdf) {
```
travist: So, this will only work with PDF forms?  If you found a generic way to solve this, then this can be pulled into the core renderer for all rendered forms.

randallknutson:
Is this specific to bootstrap or will it work with other frameworks as well?

travist:
@randallknutson it should require some testing, but it looks like it would work on all templates since he is using ```display: flex```. Good work @RomanLetsuk.

travist:
I personally don't even think this should be "legacy" support. Why don't we bring back the options of left and right alignment into the builder options as well.

PR #2365, opened 2020-02-04
File | Remove uploaded file link is not keyboard accessible

dianatsevan:


dianatsevan:
[File | Unsupported file type | Remove uploaded file link is not keyboard accessible](https://trello.com/c/BTBba7Ft/149-file-unsupported-file-type-remove-uploaded-file-link-is-not-keyboard-accessible)

PR #2390, opened 2020-02-10
Make editGrid use widget if input has a widget

alexandraRamanenka:


travist:
Can we find a better way to do this? I don't like that it is having to pass around this widgets variable to all these methods now.

Issue #2280, opened 2020-01-16
Override editForm api key validation?


PhilipeGatis:
Hi,
I`d like to ignore validation in "API" tab on field "KEY" on editForm.
Have some way?

```
const options = {
  builder: {
    data: false,
    premium: false,
  },
  editForm: {
    textfield: [
      {
        key: 'api',
        ignore: false,
        components: [
          {
            key: 'key',
            validate: false
          }
        ]
      }
    ]
  }
};
```

randallknutson:
I believe validate is an object. Have you tried setting it to validate: {}? That might work. Be aware though that the key is the most important field on the component definition as it controls where the value will be saved in the submission. If you don't make it required, weird things might start happening.

PhilipeGatis:
Yes, its an object. i try again with: 
```
const options = {
  builder: {
    data: false,
    premium: false,
  },
  editForm: {
    textfield: [
      {
        key: 'api',
        ignore: false,
        components: [
          {
            key: 'key',
            validate: {
              pattern: ''
            }
          }
        ]
      }
    ]
  }
};
```

I debbuged in chrome and the object is not overrided :(
![image](https://user-images.githubusercontent.com/4982462/72565476-b5841600-3890-11ea-9767-b6aa181e37b1.png)

randallknutson:
That looks like it should work the way you wrote it. It might be that the merge function isn't catching it correctly. Can you track through the code right above your screenshot to see if the merge is not working correctly?

If that doesn't work, we can get you on a support package and have some support resources for you.

PhilipeGatis:
@randallknutson 
I found the problem!
On function "unify Components" has an if clause where is verified this object has the property, if they have, don't merge.

https://github.com/formio/formio.js/blob/master/src/components/_classes/component/editForm/utils.js#L7 -> function
https://github.com/formio/formio.js/blob/master/src/components/_classes/component/editForm/utils.js#L15 -> first if clause
https://github.com/formio/formio.js/blob/master/src/components/_classes/component/editForm/utils.js#L20 -> second if clause
```
unifyComponents: function unifyComponents(objValue, srcValue) {
    if (objValue.key && srcValue.key) {
      if (objValue.skipMerge || srcValue.skipMerge) {
        return false;
      }

      if (objValue.label === 'pudim') {
        debugger
      }

      if (objValue.key === srcValue.key) {
        // Create complete objects by including missing keys.
        _lodash.default.each(objValue, function (value, prop) {
[...]

PR #2412, opened 2020-02-13
Improve(editFormUtils): add option to override editForm builder options

PhilipeGatis:
Ignore editForm api key validation? [#2280](https://github.com/formio/formio.js/issues/2280)

Now we can pass a overrideEditForm flag in options. example:
```
const options = {
  builder: {
    data: false,
    premium: false,
  },
  editForm: {
    textfield: [
      {
        key: 'api',
        ignore: false,
        components: [
          {
            key: 'key',
            overrideEditForm: true,
            validate:{}
          }
        ]
      }
    ]
  }
};
```

### DIFF
--- a/src/components/_classes/component/editForm/utils.js
+++ b/src/components/_classes/component/editForm/utils.js
@@ -12,12 +12,12 @@ const EditFormUtils = {
       if (objValue.key === srcValue.key) {
         // Create complete objects by including missing keys.
         _.each(objValue, (value, prop) => {
-          if (!srcValue.hasOwnProperty(prop)) {
+          if (objValue.overrideEditForm || !srcValue.hasOwnProperty(prop)) {
             srcValue[prop] = value;
           }
         });
         _.each(srcValue, (value, prop) => {
-          if (!objValue.hasOwnProperty(prop)) {
+          if (srcValue.overrideEditForm || !objValue.hasOwnProperty(prop)) {
             objValue[prop] = value;
           }
         });
--- a/src/components/_classes/component/editForm/utils.spec.js
+++ b/src/components/_classes/component/editForm/utils.spec.js
@@ -32,5 +32,16 @@ describe('Edit Form Utils', function() {
         { key: 'b', one: 1, two: 2, ok: true },
       ]);
     });
+
+    it('should override with "override" flag', () => {
+      const components = [
+        { key: 'a', label: 1, ok: true },
+        { key: 'a', label: 2, overrideEditForm: true }
+      ];
+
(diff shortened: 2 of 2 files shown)


travist:
My only comment is that I don't like such a generic property name for such a specific use case. "ignore" would not make me think that this is ONLY for ignoring the component when determining if it should allow an override. My suggestion is to call it ```overrideEditForm``` or something similar.

PhilipeGatis:
@travist changed :) thanks!

src/components/_classes/component/editForm/utils.js
```diff
@@ -12,12 +12,18 @@ const EditFormUtils = {
...
+            srcValue[prop] = value;
+          }
+          else if (!srcValue.hasOwnProperty(prop)) {
             srcValue[prop] = value;
           }
         });
         _.each(srcValue, (value, prop) => {
-          if (!objValue.hasOwnProperty(prop)) {
+          if (srcValue.overrideEditForm) {
+            srcValue[prop] = value;
```
travist, line 24: Shouldn't this say...

```js
if (srcValue.overrideEditForm) {
  objValue[prop] = value;
}
```

That being said, couldn't this be simplified with.

```js
_.each(objValue, (value, prop) => {
  if (objValue.overrideEditForm || !srcValue.hasOwnProperty(prop)) {
    srcValue[prop] = value;
  }
});
_.each(srcValue, (value, prop) => {
  if (srcValue.overrideEditForm || !objValue.hasOwnProperty(prop)) {
    objValue[prop] = value;
  }
});
```

PhilipeGatis: done

src/components/_classes/component/editForm/utils.js
```diff
@@ -12,13 +12,13 @@ const EditFormUtils = {
...
-          if (!srcValue.hasOwnProperty(prop)) {
+          if (objValue.overrideEditForm || !srcValue.hasOwnProperty(prop)) {
             srcValue[prop] = value;
           }
         });
         _.each(srcValue, (value, prop) => {
-          if (!objValue.hasOwnProperty(prop)) {
-            objValue[prop] = value;
+          if (srcValue.overrideEditForm || !objValue.hasOwnProperty(prop)) {
+            srcValue[prop] = value;
```
travist, line 21: I believe this should still say... ```objValue[prop] = value;``` You are changing it to be ```srcValue``` and I don't think that is correct.

PhilipeGatis: fixed up! sorry, my fault.

PR #2420, opened 2020-02-13
Fixed updating revision property on component change

AlexeyDavyd:


src/components/form/Form.js
```diff
@@ -84,8 +84,8 @@ export default class FormComponent extends Component {
     }
 
     // Add revision version if set.
-    if (this.component.formRevision || this.component.formRevision === 0) {
-      this.formSrc += `/v/${this.component.formRevision}`;
+    if (this.component.revisions || this.component.revisions === 0) {
```
travist: It should be singular. "revision"

travist:
Why are we changing the property name? We have customers that have already built forms with this configured, so won't this break their forms if they upgrade? Could we make this reverse compatible so that it will work with either "formRevision" or "revision"?...

travist:
@AlexeyDavyd I am still waiting for confirmation from you that this will not mess up forms that were built with 3.x builder and rendered in 4.x renderer with this patch. If you can confirm that there is no breaking change, then we will pull this in.

AlexeyDavyd:
@travist After my checks I can confirm, that it shouldn't break existing forms.

PR #2433, opened 2020-02-17
Added validation after page was changed

alexandraRamanenka:


ababic:
@travist / @alexandraRamanenka I could be wrong, but I believe there is an issue with this change, in that validation is triggered whether or not the form is in `readOnly` mode, which is does not match the current behaviour in [nextPage()](https://github.com/formio/formio.js/blob/master/src/Wizard.js#L403). Validation also appears to run twice now when nextPage() is used.

Is there an issue somewhere that explains what problem this is solving? it would be really nice to understand the context behind this change.

travist:
I agree @ababic.  @alexandraRamanenka please address @ababic concerns and describe the specific issues that were addressed with this change and also please fix the regression that was caused because of it.

alexandraRamanenka:
@ababic thanks for your attentiveness. It's really have been missed. The issue was that when we changed page with wizard's navigation, all the errors messages were disappeared. 
I created new solution for that https://github.com/formio/formio.js/pull/2487

ababic:
Thank you both. @alexandraRamanenka, it sounds like you have this in hand :)

PR #2439, opened 2020-02-18
Fixed styles to prevent hiding basic components button when scrolling

alexandraRamanenka:


travist:
This style should probably be added to the formio-app and not the core builder. I imagine we don't want to have all builders add this padding to the top.

travist:
Fixed in this commit. https://github.com/formio/formio-app/commit/a5c85c4d9be1ead69ce6324f7900ccac8fc620cd

PR #2440, opened 2020-02-18
Removed placeholder from defaultWysiwyg settings for "ace" editor

alexandraRamanenka:
Placeholders for 'ace' editor are added with checkAcePlaceholder method


travist:
It seems to me we can just remove the ```checkAcePlaceholder``` method instead of this and just let ACE use the placeholder params. This feels like we are introducing two wrongs to make a right.

travist:
Fixed this way https://github.com/formio/formio.js/commit/a483e72693009069fe7ec6b0b7cdb99f1bdf4d2c

PR #2445, opened 2020-02-19
File | Single File | File upload/remove should be announced by screenreader

dianatsevan:


dianatsevan:
[File | Single File | File upload/remove should be announced by screenreader](https://trello.com/c/K9yNhrkJ/105-file-file-upload-remove-should-be-announced-by-screenreader)

PR #2448, opened 2020-02-19
Select data are displayed in HTML render mode

TanyaGashtold:


travist:
Does this work if the value is complex? Like a data object?  For example, with a Resource select dropdown where the selected item is a submission?

TanyaGashtold:
Travis, unfortunately, I did not manage to make test in HTML mode with resource select dropdown where the selected item is a submission or any other case when selected data is complex because I did not find a way to insert those data in submission in HTML mode.
The problem I worked was connected with not displaying multivalue in HTML mode, this was because the multivalue select provides data as array and this was not taken into consideration. I did not make serious changes, just added a check whether value is an array is or not and used the itemValue function inside as  before. Now it works with multivalue.
Due to some problems with running tests, I had to create new pull request for this issue. Please check it. https://github.com/formio/formio.js/pull/2494.

PR #2452, opened 2020-02-20
Added Role API methods

airarrazaval:
The following methods have been added to interact with the project role endpoint:

- loadRoles
- loadRole
- saveRole
- deleteRole

The following properties are also registered:

- rolesUrl
- roleUrl (only when full role url is passed as path)
- roleId (only when full role url is passed as path)

Finally, changed a string regex search with a regex test method which is more efficient.


travist:
Looks good!  I am also working on more extensive docs for this and that is coming soon. Thank you for the contribution.

PR #2456, opened 2020-02-21
Fix(Evaluator): fix RexExp

vlad-shusterman:


travist:
@vlad-shusterman Please write some tests for this one.

PR #2461, opened 2020-02-24
Added trailing zeros in currency component. Please run 'test:updateRen…

TanyaGashtold:
…ders' before testing.


travist:
Please add some tests.

PR #2462, opened 2020-02-24
Bugfix/added formio form class to wizard

TanyaGashtold:
Travis, by mistake I made a branch not from master but from other branch, please merge first formio/formio.js: Pull Request 2461 (currency component) to see pure changes regarding formio form class addition.

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -929,9 +929,13 @@ export default class Webform extends NestedDataComponent {
     return this.ready;
   }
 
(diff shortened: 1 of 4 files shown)


travist:
Please cherry-pick this commit into its own branch so that we don't pull in any other commits.

TanyaGashtold:
Here is the link to new pull request: https://github.com/formio/formio.js/pull/2491

PR #2465, opened 2020-02-25
Improvement/changed validation message

TanyaGashtold:


src/components/_classes/component/Component.unit.js
```diff
@@ -62,7 +62,7 @@ describe('Component', () => {
       bad: {
         value: 'testte',
         field: 'firstName',
-        error: 'First Name must be shorter than 6 characters.'
+        error: 'First Name must contain 5 or less characters.'
```
travist: This sentence seems weird to me.  I would rather it say "First Name must be 5 characters or less"

src/components/_classes/component/Component.unit.js
```diff
@@ -47,7 +47,7 @@ describe('Component', () => {
       bad: {
         value: 't',
         field: 'firstName',
-        error: 'First Name must be longer than 1 characters.'
+        error: 'First Name must contain 2 or more characters.'
```
travist: Please change this to "First Name must be at least 2 characters."

src/components/_classes/component/Component.unit.js
```diff
@@ -47,7 +47,7 @@ describe('Component', () => {
       bad: {
         value: 't',
         field: 'firstName',
-        error: 'First Name must be longer than 1 characters.'
+        error: 'First Name must be at least 2 characters.'
```
travist: Also change this to "must have" so it is consistent with the minimum words messages.

PR #2469, opened 2020-02-25
Fix (DateTime) : change min/max components to datetime

vlad-shusterman:
Changed that to prevent adding invalid date in textfield by user.

### DIFF
--- a/src/components/datetime/editForm/DateTime.edit.date.js
+++ b/src/components/datetime/editForm/DateTime.edit.date.js
@@ -15,7 +15,7 @@ export default [
     input: true,
     key: 'enableMinDateInput',
     label: 'Use Input to add moment.js for minDate',
-    weight: 0,
+    weight: 10,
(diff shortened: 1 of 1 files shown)


src/components/datetime/editForm/DateTime.edit.date.js
```diff
@@ -11,20 +11,20 @@ export default [
     tooltip: 'Enables date input for this field.'
   },
   {
-    type: 'textfield',
+    type: 'datetime',
```
travist: This removes the ability to use moment to add min and max dates like the following.

```js
moment().subtract(10, \'days\')
```

vlad-shusterman: @travist 
Updated with simple mask since there is no possibility to add moment object to datetime input.
Hope that's enough.

travist:
How can I type "moment()" in this when there is an inputMask applied?

PR #2475, opened 2020-02-27
Fix(HTML) : bug with conditions on init

vlad-shusterman:


travist:
Please write some automated tests for this change.

PR #2481, opened 2020-02-28
Fix: Infinite setValue call on textarea

UladzimirKudzelka:


src/components/textarea/TextArea.unit.js
```diff
@@ -11,4 +14,18 @@ describe('TextArea Component', () => {
       Harness.testElements(component, 'textarea', 1);
     });
   });
+
+  it('setValue should be called only once', () => {
```
travist: 👏

src/components/_classes/component/Component.js
```diff
@@ -2145,7 +2145,7 @@ export default class Component extends Element {
       return changed;
     }
     const isArray = Array.isArray(value);
-    if (isArray && this.refs.input && this.refs.input.length !== value.length) {
+    if (isArray && this.type === 'textField' && this.refs.input && this.refs.input.length !== value.length) {
```
travist: This type of code should not be within the Component class. I would prefer that the TextFieldComponent extend this method to overwrite this logic. 

It seems like this is here so that the multi-valued field would add a new item if the value is set with an additional item. Regardless, there are many other fields that need to work with the multiple flag and this will break those components now.

UladzimirKudzelka: Travis, you mean to move this condition to setValue function inside TextFieldComponent?
    if (isArray && this.type = 'textField' && this.refs.input && this.refs.input.length !== value.length) {
      this.redraw();
    }

travist: I am saying that I don't think this fix is good since it is basically only enabling it for textFields and there are MANY other fields that enable multiple flags.

travist:
Fixed with https://github.com/formio/formio.js/pull/2548

PR #2486, opened 2020-03-02
Added sketchpad styles

alexandraRamanenka:


alexandraRamanenka:
Added sketchpad styles to the premium, but we also need to replace 
.component-rendering-hidden {
    visibility: hidden;
  } 
with display: none;
Because it sometimes causes conflicts with buttons or make cursor be blinking

PR #2487, opened 2020-03-02
Fixed Wizard validation is run in readOnly mode

alexandraRamanenka:


travist:
Please write a test.

src/Wizard.js
```diff
@@ -319,6 +316,14 @@ export default class Wizard extends Webform {
     return NativePromise.reject('Page not found');
   }
 
+  goToPage(index) {
```
ababic: I feel that having both a `setPage` and `goToPage` method may cause a little confusion going forward. Perhaps we could add an additional argument to `setPage` to toggle this behaviour instead, perhaps with a `false` default value?

randallknutson: Agreed. We should try to not duplicate nearly identical functions.

travist:
Looks good. Just waiting on some tests to ensure this continues to work in the future.

PR #2500, opened 2020-03-05
Fix (DataGrid) : disable child components

vlad-shusterman:


antonSoftensity:
[Wizard | Add role="tablist" and role="tab"](https://trello.com/c/bghrMY4H/186-wizard-add-roletablist-and-roletab)

travist:
Please write some tests.

PR #2503, opened 2020-03-05
Fixed validation of all fields when data is entered only in one field

TanyaGashtold:


travist:
We will fix this by reverting 2b3ef7fce42410ad5e2e3a779afa401dc4209d97

PR #2511, opened 2020-03-09
Bugfix/fixed noValidate flag in set submission

TanyaGashtold:


travist:
There are a lot of places in this code that seem like hacks (using the "fromSubmission" which is actually a flag that I am a little confused about is one of them). I will implement the tests and the fix in the Form.js separately and see if I can fix the tests a different way.

PR #2519, opened 2020-03-11
Feat(Component): add option for validateOnInit

vlad-shusterman:
Connected with https://github.com/formio/angular-formio/pull/490


travist:
This seems to be handled with the work that @TanyaGashtold is working on.

travist:
I accomplished this via the way we discussed.

PR #2527, opened 2020-03-13
Fixed calculated value is not set after validation on submit if  'All…

TanyaGashtold:
…ow override calculated value' is true

### DIFF
--- a/src/Webform.unit.js
+++ b/src/Webform.unit.js
@@ -5,11 +5,38 @@ import each from 'lodash/each';
 import Harness from '../test/harness';
 import FormTests from '../test/forms';
 import Webform from './Webform';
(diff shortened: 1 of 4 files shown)


travist:
@TanyaGashtold this breaks "Property Name" options in the Form builder.   I will need to revert it.  We need to write a test on how our Property Names function. Here is how you reproduce.

- Use the form builder
- Drag a single Text Field and give it a Label of "First Name"
- Now click on the API tab, and change the Property Name to say "testing"
- Now go back to the Label and change it to say "First Name 2".
- Go back to Property Name and you will see it was reset to "firstName2" even though I overrode the value to "testing". It should have stayed as "testing"

PR #2538, opened 2020-03-16
VPAT-208 Fixed error messages focus loss in IE11

antonSoftensity:


antonSoftensity:
[Wizard | IE11 | When an error message in the error section is in focus, pressing Enter button does not redirect to the form with the error](https://trello.com/c/QB9B8NKc/208-wizard-ie11-when-an-error-message-in-the-error-section-is-in-focus-pressing-enter-button-does-not-redirect-to-the-form-with-the)

PR #2539, opened 2020-03-16
WIP: Fix numbers not clearing on reset of submission.

randallknutson:
This issue occurs if you reset the underlying value of a number component. This is because the emptyValue of numbers is 'null' and we were checking null in updateComponentValue and defaulting back to the existing value. By changing this we override the null check if the emptyValue of a component is also null, allowing it to be removed.


travist:
Fixed with https://github.com/formio/formio.js/pull/2544

PR #2541, opened 2020-03-17
Add pdf on change handler

RomanLetsuk:


travist:
Which item is this fixing?

travist:
Also write tests please.

travist:
Did you even test this?  When I pulled it in and compiled it is gives me an error that says 

```
/Users/travistidwell/Documents/formio/modules/formio.js/src/PDF.js
  24:25  error  '_' is not defined  no-undef
```

PR #2545, opened 2020-03-18
Revert "Update iframe src building"

travist:
Reverts formio/formio.js#2532


travist:
@RomanLetsuk I'm sorry but we need to just revert this and probably keep it that way. It keeps causing too many issues with customers and also different browsers have issues with this method. We will stick with regular src iframes for now.

PR #2559, opened 2020-03-20
Fixed payment error

alexandraRamanenka:
Made form send a full submission instead of only submission's data when the form's action specified.


alexandraRamanenka:
@travist , there were entire submissions sent in the portal to the url from _form.action, so all urls from _form.action  (incluing '/payeezy') expects request's body to have the 'data' field, but it hasn't.

travist:
So how is it that Actions currently work? I would expect that they would also be broken?

PR #2566, opened 2020-03-23
VPAT-236 Changed variable types for future extension in vpat

antonSoftensity:


antonSoftensity:
[Button | All browsers | Add an option to pick theme for a button](https://trello.com/c/VANMJ4hz/236-button-all-browsers-add-an-option-to-pick-theme-for-a-button)

PR #2567, opened 2020-03-23
Fix(DateTime): add option to change from moment input to calendar

vlad-shusterman:


src/components/datetime/editForm/DateTime.edit.date.js
```diff
@@ -10,23 +10,79 @@ export default [
     weight: 0,
     tooltip: 'Enables date input for this field.'
   },
+  {
+    type: 'checkbox',
```
travist: Can we make this non-persistent so it won't get saved with the component JSON?

vlad-shusterman: @travist Do you mean I should move customConditional to component from json and add it automatically when component creates?

travist: No, I just mean to add the persistent flag to this component.  ```persistent: false``` or maybe it should be ```persistent: 'client-only'```.

Issue #2554, opened 2020-03-19
[BUG] Conditional element not shown in read mode


diondirza:
## Environment
- Formio.js version: v4.9.0-rc.12
- Frontend framework: React
- Browser: Chrome 
- Browser version: 80.0.3987.149

## Steps to Reproduce
1. Use this fiddle http://jsfiddle.net/diondirza/qevn1scu/2/
2. Look at **Hospitalization** radio when user select 2nd option we will see a text area for **Reason(s) for hospitalization** 
3. In edit mode this text area visible, but when you toggle `readOnly: true` this element vanishes.

## Expected behavior
The conditional element should appear in read mode as well. Working as expected in version v4.8.1

```

jhen1422:
@diondirza 
Thank you for the report. We were able to replicate this issue on our end as well and have created a ticket in our 4x branch. We will review this item during our planning meeting next week. Please note, we welcome pull requests in this library. If you need this item expedited, please feel free to contact support@form.io for pricing.

randallknutson:
This probably has to do with the conditional you have set on the conditional field. When in read only mode it won't calculate or default the value. Then, if you are depending on the default or calculated value for your conditional it will not evaluate to true.

iamsid:
Seeing this issue in v4.9.6.
As a workaround in my app I have kept the form as 'readOnly: false'.. and set the formio div as 'disabled' with opacity 70%.

VikkiAlenn:
Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #2573, opened 2020-03-26
Fixed issues with read only forms not applying conditions.

travist:


travist:
Fixes https://github.com/formio/formio.js/issues/2554

PR #2579, opened 2020-03-28
Removed the erroneous requirement that if you have a checkbox configu…

travist:
…red as a radio, it should still add the checkbox value to the payload.


travist:
For some reason, we were deliberately adding the individual checkbox data when it is configured as a radio. This is not correct since a checkbox configured as radio stores its data within the radio name property.

PR #2596, opened 2020-04-01
Revert "Fix github #1953: Override .defaultValue of subcomponents."

travist:
Reverts formio/formio.js#2049


travist:
This was causing issues where the subforms values could not be switched with other values for the subform.

Issue #2599, opened 2020-04-02
[BUG] react-formio updating read only props 


JovanT:
## Environment
Please provide as many details as you can:

- Hosting type
  - [ ] Form.io
  - [x] Local deployment
    - Version: 
- Formio.js version: 4.10.0-beta.3
- Frontend framework: react 16.13.1
- Browser: Google Chrome
- Browser version: 80.0.3987.162

## Steps to Reproduce
1. "redux-immutable-state-invariant": "^2.1.0" and "immer": "^6.0.3"
2. load SubmissionGrid or Webform on view 
3. error for grid is update this.component.id but for Webfrom is this.submmission.data

## Expected behavior
we don't have to update props send to formio render 
## Observed behavior
in case for Comonent.js and for Webform.js we have update on reference porps send as formio form configuration.
## Example
If possible, please provide a screenshot, live example (via [JSFiddle](https://jsfiddle.net/) or similar), and/or example code to help demonstrate the issue.

For code or form JSON, please enclose in a code block:

```js
// your code here
```

JovanT:
I already have fix and ill like to make pull request on formio.js

JovanT:
all is ok

PR #2600, opened 2020-04-02
react-formio updating read only props #2599

JovanT:
solving bug #2599


JovanT:
> This seems pretty disruptive and will most likely break other repos. The submission object is purposefully a "reference" so this will break other libraries that depend on it to be that way. Is it possible to fix this within the react-formio library where the submission is cloned before it is sent to setValue? Maybe also do the same for the component JSON that is passed to be rendered from within react-formio?

OK I'm going to solve it from react-formio it is also possible that thanks.

PR #2612, opened 2020-04-05
VPAT-252 Added an option to add column breakpoints

antonSoftensity:


antonSoftensity:
[Add an option to add column breakpoints to USWDS template](https://trello.com/c/SJxV9Txr/252-add-an-option-to-add-column-breakpoints-to-uswds-template)

PR #2629, opened 2020-04-09
FJS-844 implement basic fix for checkbox type radio submit values in tableView

mikekotikov:
image

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -872,7 +872,7 @@ export function fieldData(data, component) {
 
     // Fix for checkbox type radio submission values in tableView
     if (component.type === 'checkbox' && component.inputType === 'radio') {
-      return !!data[component.name];
+      return data[component.name] === component.value;
     }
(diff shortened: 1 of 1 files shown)


src/utils/utils.js
```diff
@@ -869,6 +869,12 @@ export function fieldData(data, component) {
     if (component.multiple && !Array.isArray(data[component.key])) {
       data[component.key] = [data[component.key]];
     }
+
+    // Fix for checkbox type radio submission values in tableView
+    if (component.type === 'checkbox' && component.inputType === 'radio') {
+      return !!data[component.name];
```
travist, line 875: ```data[component.name]``` will be a string that will be set to the ```component.value```.  You need to also check for component.value I believe like this.

```
return data[component.name] === component.value;
```

mikekotikov: Makes sense! Missed that thing, thanks for catching.

Issue #2613, opened 2020-04-06
[BUG] Form validation


pixie79:
## Environment
If you have a disabled or hidden field validation logic has stopped working on those fields. With FormIOJS 4.9.5 you can validate a disabled or hidden field but no longer. Whilst initially this may seem logical. We do have a form where we ask users to enter total number of hours worked. e.g 10. Then we ask them to give us the break down of the hours. At the bottom we have a text box with a calculated value to ensure that they have allocated all 10 hours. If not then upon submission with raise a validation error telling them the totals do not match.

Can we please revert this function?

- Hosting type
  - [X ] Local deployment
    - Version: 
- Formio.js version: 4.9.8

jhen1422:
@pixie79 Is this issue occurring on version 4.9.13?

pixie79:
Yes, Tested with latest as well

randallknutson:
Hidden components (components that are hidden and the actual hidden components) should not have validations since they by definition aren't active on the form. Disabled components SHOULD have validations since like in your use case they can be modified by other components.

pixie79:
Ok, It is not working with Disabled ones currently it appear the validation logic is not firing.

jhen1422:
@pixie79 
Thank you for the report. We were able to replicate this issue on our end as well and have added a ticket to our developer backlog. We will review this item during our planning meeting next week. Please note, we welcome pull requests in this library. If you need this item expedited, please feel free to contact support@form.io for pricing.

PR #2635, opened 2020-04-10
Don't skip validation for disabled components

gavinorland:
This reverts a change made in https://github.com/formio/formio.js/commit/129e5c8fcc81195dd5318fd726cbb2b680d14444 and is a fix for https://github.com/formio/formio.js/issues/2613.

### DIFF
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -2668,6 +2668,8 @@ export default class Component extends Element {
 
   shouldSkipValidation(data, dirty, row) {
     const rules = [
+      // Force valid if component is read-only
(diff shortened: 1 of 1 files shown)


travist:
This was put in place to originally fix "readOnly" forms from triggering all the validation errors. I think this is doable, but would like to propose to replace 

```
() => this.shouldDisabled,
```

with the following...

```
() => this.options.readOnly,
```

Would this work for your use case?

gavinorland:
Thanks for your reply, @travist.

AFAIK that will be fine for us. I've updated the branch to that effect, and if there does turn out to be any problem further down the line I could get back to you. 🙂

PR #2655, opened 2020-04-15
fixed impossibility to create select with resource because resource i…

TanyaGashtold:
…d identified as a number


TanyaGashtold:
There is no ticket for this issue, I just found out it when trying to reproduce another problem with select with resource

PR #2656, opened 2020-04-15
VPAT-244 Fixed focus not going to error section after validation in W…

antonSoftensity:
…izard


antonSoftensity:
[Form and Wizard | All browsers | Focus does not go to the 1st error in the error section after pressing Submit button in a Form or Next/Submit Form buttons in a Wizard if any errors occurred](https://trello.com/c/Hocp6hLH/244-form-and-wizard-all-browsers-focus-does-not-go-to-the-1st-error-in-the-error-section-after-pressing-submit-button-in-a-form-or-n)

PR #2689, opened 2020-04-21
Ldpm updates to fix pdf issues

ldpmcasebook:


travist:
To make this pull request easier to commit, I would do the following.

```shell
git fetch origin;
git reset origin/master --soft;
git add src
git commit -m "Your commit message";
```

This will create a SINGLE commit and only include the changes made to the "src" folder.

PR #2705, opened 2020-04-23
FJS-869/FJS-887

Tkachez:
since last fix there were some issues, so condition was added that prevents value calculation if it is not needed


src/components/_classes/component/Component.js
```diff
@@ -2549,37 +2549,41 @@ export default class Component extends Element {
...
-    this.checkRefreshOn(flags.changed);
-    if (flags.noCheck) {
-      return true;
-    }
-    this.calculateComponentValue(data, flags, row);
-    this.checkComponentConditions(data, flags, row);
-    if (flags.noValidate) {
-      return true;
-    }
+    if (flags && flags.changed && _.isEqual(flags.changed.instance.data, row)) {
```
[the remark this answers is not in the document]
Tkachez: @travist, hey, please check again!

src/components/_classes/component/Component.js
```diff
@@ -2549,37 +2553,39 @@ export default class Component extends Element {
...
-    this.checkRefreshOn(flags.changed);
-    if (flags.noCheck) {
-      return true;
-    }
-    this.calculateComponentValue(data, flags, row);
-    this.checkComponentConditions(data, flags, row);
-    if (flags.noValidate) {
-      return true;
-    }
+      data = data || this.rootValue;
```
travist, line 2556: This adds a space so that is why we see a diff here. Please remove the space so the commit is clean.

travist:
This seems like it would break if you made a calculation based on data that was outside your data context (row). Have you tested this against performing calculations within a data grid that uses data at the global level?  Please write some tests before we accept this.

Tkachez:
@travist, I'm a bit confused. Could you please clarify waht do you mean by 'calculation based on data that was outside your data context (row)'. Do you mean if there is condition in another component to set value inside data grid's component? Also a bit confused by  'calculations within a data grid that uses data at the global level'.
P.S. I've tested components that were inside datagrid and everything was working fine.

travist:
What this change is telling me is that the calculations will only be performed when the components data object is equal to the row context in which the check is being performed. This basically means that calculations will only work if they share the same data context. What I am saying is that this form would probably not work with this change...  

https://nhfsrgycfushxca.form.io/calculation
https://pro.formview.io/#/nhfsrgycfushxca/calculation

Tkachez:
Ok, thanks for the update, will dive into it tomorrow!

Tkachez:
Updated the logic. For now if label is changed, value will be calculated from label, but if we change value it will override calculated value from label. Still needs some unit tests.

src/components/_classes/component/Component.js
```diff
@@ -2391,6 +2392,14 @@ export default class Component extends Element {
       return false;
     }
 
+    if (flags.changed && !flags.changed.value) {
+      return false;
+    }
+
+    if (flags.changed && row.label && flags.changed.value !== row.label) {
```
travist, line 2399: And what if they have a field in their form named "Label"?  This will cause a "magic" bug that will only break if you name a textfield Label. Seems a little precarious.

src/components/_classes/component/Component.js
```diff
@@ -2369,6 +2369,7 @@ export default class Component extends Element {
 
     let firstPass = false;
     const dataValue = this.dataValue;
+    console.log(this.dataValue);
```
travist, line 2372: Please remove console.log.

Tkachez:
I guess I've found the solution that will work for every mentioned case. Will wait for approval before covering it with tests.

src/components/_classes/component/Component.js
```diff
@@ -2391,6 +2391,18 @@ export default class Component extends Element {
       return false;
     }
 
+    if (flags.changed && !flags.changed.value) {
+      return false;
+    }
+
+    if (flags.changed && flags.changed.value && !row.values && flags.changed.instance.data !== row) {
```
travist, line 2398: So, does this mean if I have a form and give it a label of "Values" (which will give it a key of "values"), that I should never calculate the value of it as long as the row context is different than the instance data?  I have to admit this is very strange, and does not seem to really solve the main problem.

travist:
Getting resolved in https://github.com/formio/formio.js/pull/2789

PR #2710, opened 2020-04-23
VPAT-314 Removed role alert from error messages container

antonSoftensity:
Presence of this role in error section means screenreader announcements will get interrupted by error section announcement every time error section is redrawn, which should not be happening.


antonSoftensity:
[Form | All br/sr | After one of the errors is fixed, SRs start announcing the error section with all the errors within it](https://trello.com/c/3wCq0nEf/314-form-all-br-sr-after-one-of-the-errors-is-fixed-srs-start-announcing-the-error-section-with-all-the-errors-within-it)

PR #2712, opened 2020-04-24
VPAT-335 Fixed currency input announcing invalid entry with any value

antonSoftensity:


antonSoftensity:
[Currency | Firefox, Chrome/NVDA, Chrome/Jaws | Invalid entry is announced with any value in the field](https://trello.com/c/BBW5TCSe/335-currency-firefox-chrome-nvda-chrome-jaws-invalid-entry-is-announced-with-any-value-in-the-field)

PR #2713, opened 2020-04-24
PDF 59 - Fixed redirection after hitting the pdf submit button

dianatsevan:


travist:
We should not be scrolling at all when they click the button. Why is this needed? If this is here to fix a problem with the page scrolling after they press the submit button, then this is more likely a bandaid on top of another problem. Please try to fix the original problem of why the page is scrolling to begin with.

PR #2715, opened 2020-04-27
Fixed submission display in read only mode in tree (ticket - FJS-207)

TanyaGashtold:


paulfalgout:
@travist  This seems like a really poor solution.  It clearly calls `redraw` twice one right after the other in the case of removal and if this is in the midst of a change event on the tree add/edit field it clears the add/edit form.  If this is truly only for readOnly why not wrap it in `if (this.options.readOnly)` such that it doesn't potentially break other use cases?

PR #2722, opened 2020-04-29
PDF 44 - Fixed an issue of not correct work of Date/Time component

dianatsevan:


src/widgets/CalendarWidget.js
```diff
@@ -70,7 +70,7 @@ export default class CalendarWidget extends InputWidget {
   loadZones() {
     const timezone = this.timezone;
     if (!zonesLoaded() && shouldLoadZones(timezone)) {
-      loadZones(timezone).then(() => this.emit('redraw'));
+      loadZones(timezone).then(() => this.rebuild());
```
travist, line 73: Did you actually test this?  ```this.rebuild``` is not a method in CalendarWidget, nor in InputWidget, no in Element.  I am pretty sure this will cause a JavaScript error. Please write a test to prove otherwise.

dianatsevan: Yeah, my mistake. I lost 'emit'.

travist:
Again I feel like you are just guessing here. Are you actually testing these changes?

First of all, widgets get their own instance of the EventEmitter seen @ https://github.com/formio/formio.js/blob/master/src/Element.js#L47

And there are only two events that widgets listen to https://github.com/formio/formio.js/blob/master/src/components/_classes/input/Input.js#L262 and "rebuild" is not one of them.

How is this doing anything?  It seems it would be the same result to just delete the whole line together, but that does not seem correct to me.

dianatsevan:
Yeah, I looked up more carefully and I got it. I totally agree with you. Will continue to work on it.

src/widgets/CalendarWidget.js
```diff
@@ -70,7 +70,7 @@ export default class CalendarWidget extends InputWidget {
   loadZones() {
     const timezone = this.timezone;
     if (!zonesLoaded() && shouldLoadZones(timezone)) {
-      loadZones(timezone).then(() => this.emit('redraw'));
+      loadZones(timezone).then(() => this.emit('rebuild'));
```
travist, line 73: Did you need to change this back to ```this.emit('redraw')```?  This event is not being handled currently so it serves no purpose to change it to ```rebuild```

dianatsevan: Omg, yes of course! I'm already confused and forgot to bring it back..

PR #2726, opened 2020-04-29
VPAT-353 Fixed error section links not working in nvda focus mode

antonSoftensity:


antonSoftensity:
[Form and Wizard | Chrome/NVDA | Pressing Enter when an error in the error section is in focus, focus does not go to the corresponding component](https://trello.com/c/wXT3AngP/353-form-and-wizard-chrome-nvda-pressing-enter-when-an-error-in-the-error-section-is-in-focus-focus-does-not-go-to-the-corresponding)

PR #2730, opened 2020-04-30
Add round function to calculate value context

alexandraRamanenka:


src/components/_classes/component/Component.js
```diff
@@ -2404,7 +2404,12 @@ export default class Component extends Element {
     const calculatedValue = this.evaluate(this.component.calculateValue, {
       value: dataValue,
       data,
-      row: row || this.data
+      row: row || this.data,
+      round: function round(number, precision) {
```
travist, line 2408: Can we just add this to the "utils" instead... that way they could use ```utils.round```

alexandraRamanenka: All right

PR #2736, opened 2020-04-30
PDF-63: Update events emitting

RomanLetsuk:


src/PDFBuilder.js
```diff
@@ -294,7 +294,7 @@ export default class PDFBuilder extends WebformBuilder {
         if (!this.options.noNewEdit) {
           this.editComponent(component.component, this.webform.iframeElement);
         }
```
travist, line 296: Shouldn't there also be an ```else``` statement and then ```emit('change', this.form)``` here?  Something like...

```js
if (!this.options.noNewEdit) {

}
else {
  this.emit('change', this.form);
}
```

PR #2746, opened 2020-05-04
Fjs-727

Tkachez:
Now there is a logic in attach method that removes unnecessary elements from dom. There is also 'Condesed' mode for breadcrumbs, but I wasn't quite sure of what it does. I guess it just removes padding, but I want to be sure, so I just skipped that option for now.

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -176,6 +176,18 @@ export default class Wizard extends Webform {
       [`${this.wizardKey}-link`]: 'multiple',
     });
 
+    for (const button in this.form.components[this.page].buttonSettings) {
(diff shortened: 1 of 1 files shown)


src/Wizard.js
```diff
@@ -176,6 +176,18 @@ export default class Wizard extends Webform {
       [`${this.wizardKey}-link`]: 'multiple',
     });
 
+    for (const button in this.form.components[this.page].buttonSettings) {
```
travist, line 179: Why not just add statements to the template that will conditionally add these elements. This seems very inefficient that they are being added to the DOM only to be immediately taken away. I would prefer to just add conditions to the templates themselves.

PR #2753, opened 2020-05-06
PDF 48 - Added 'Fixed size' checkbox for textarea edit form

dianatsevan:


travist:
Where is this being used?  Why is it needed?

dianatsevan:
> Where is this being used? Why is it needed?
Owing to this checkbox user have an ability to fix the set sizes of the textarea so that in view mode the component won't be expanded regarding how many text is in it. The customer wanted to have it adjustable.

travist:
Right, but what I see so far is just the ability to set the configuration. Where in the renderer does it actually do something?

dianatsevan:
Another PR was added to formio-pdf

travist:
So if this only applies to PDF forms, then can we only show this checkbox if the display of the form is "pdf"?

dianatsevan:
Discussed it with Yury and we decided to make it common because we can switch between pdf view and web form view and user should see all the options.

travist:
I think what we should do here is add this to the "Overlay Settings" which are already only applicable to PDF forms.

dianatsevan:
Okay, will do that and ping you.

dianatsevan:
Updated.

PR #2761, opened 2020-05-07
PDf 61 - Added a condition for panel opening if the pdf is being generated

dianatsevan:


alexmaie:
Hello FormIO team,

Would it be possible to get an ETA related to this bug?

Thanks
Alex

PR #2768, opened 2020-05-08
VPAT-363 Fixed issue with custom validation on blur

antonSoftensity:


antonSoftensity:
[USWDS-363 (FRB) Validate On Blur issue](https://trello.com/c/DybyTBF6/363-uswds-363-frb-validate-on-blur-issue)

PR #2771, opened 2020-05-09
FJS-916: add sanitize option to avoid malicious JS code

mikekotikov:


randallknutson:
Our philosophy is to allow any input but treat all entered data as potentially dangerous and sanitize on output. This is trying to sanitize the input which is not how to want it to work. Please fix this in the output side.

PR #2783, opened 2020-05-12
Remove pageQuery global cache

edwinanciani:


src/Formio.js
```diff
@@ -1251,10 +1251,6 @@ export default class Formio {
   }
 
   static pageQuery() {
-    if (Formio._pageQuery) {
```
travist, line 1254: There is no need to have ```Formio._pageQuery``` anymore. I would just turn this into a local variable to the method.

PR #2807, opened 2020-05-18
Add unique values property to select component

AlexeyDavyd:


src/components/select/Select.js
```diff
@@ -190,6 +191,14 @@ export default class SelectComponent extends Field {
       label: label
     };
 
+    const skipOption = this.component.uniqueValues
+      ? !!this.selectOptions.find((selectOption) => _.isEqual(selectOption.value, option.value))
```
travist, line 195: This would not work for Resource select, and if they have custom objects in the list. Maybe we should only show this option if they are using the "Values" option?

src/components/select/editForm/Select.edit.display.js
```diff
@@ -15,4 +15,12 @@ export default [
       ],
     },
   },
+  {
+    weight: 1230,
+    type: 'checkbox',
+    label: 'Unique values',
+    tooltip: 'Left only unique dropdown values.',
```
travist, line 22: Make this say... "Display only unique dropdown values."

src/components/select/Select.js
AlexeyDavyd: This is working for me with resource
image

src/components/select/editForm/Select.edit.display.js
```diff
@@ -15,4 +15,12 @@ export default [
       ],
     },
   },
+  {
+    weight: 1230,
+    type: 'checkbox',
+    label: 'Unique values',
+    tooltip: 'Display only unique dropdown values.',
```
travist, line 22: This should really say "Unique Options" so that it is not confused with Unique validation which validates the value is unique.  I would like to propose that we change anywhere it says "values" with "options", so the property will be called ```uniqueOptions``` as well.

src/components/select/Select.js
```diff
@@ -32,6 +32,7 @@ export default class SelectComponent extends Field {
       template: '<span>{{ item.label }}</span>',
       selectFields: '',
       searchThreshold: 0.3,
+      uniqueValues: false,
```
travist, line 35: Should be ```uniqueOptions```

PR #2809, opened 2020-05-18
Choice.js dropdown opens now on first click

### LINKED ISSUE #2801: [BUG] - ChoicesWrapper blocks opening dropdown on first click after clear
## Environment

- Hosting type
  - [x] Form.io
  - [ ] Local deployment
    - Version: 
- Formio.js version: 4.7.4
- Frontend framework: Angular
- Browser: Chromium based and Firefox
- Browser version: Chromium build 81.0.4044.138, Firefox 76.0.1

## Steps to Reproduce
1. Have a select component with choice.js 
2. Open it and select a value
3. Use the `Remove Item` button to clear the selection
4. Click on the container to open the dropdown
5. The dropdown doesn't open
6. Click it again; It opens

## Expected behavior
It should open the dropdown on the first click 

## Observed behavior
Opens on the second click. The first click just appends the class `is-focused`.  

## Example
If possible, please provide a screenshot, live example (via [JSFiddle](https://jsfiddle.net/) or similar), and/or example code to help demonstrate the issue.

Peek 2020-05-15 18-30

pkarakal:
When clearing the selected item in a dropdown using `choice.js` and then clicking the dropdown, the dropdown didn't open. It only came into focus. 

With this change, the dropdown opens on the first click and not the second 

Closes #2801

### DIFF
--- a/src/utils/ChoicesWrapper.js
+++ b/src/utils/ChoicesWrapper.js
@@ -51,7 +51,6 @@ class ChoicesWrapper extends Choices {
 
     this._onTabKey = this._onTabKey.bind(this);
     this.isDirectionUsing = false;
-    this.shouldOpenDropDown = true;
   }
 
   _handleButtonAction(activeItems, element) {
@@ -68,7 +67,6 @@ class ChoicesWrapper extends Choices {
       return;
     }
 
-    this.shouldOpenDropDown = true;
     super._handleButtonAction(activeItems, element);
   }
 
@@ -184,11 +182,6 @@ class ChoicesWrapper extends Choices {
   }
 
   showDropdown(...args) {
-    if (!this.shouldOpenDropDown) {
-      this.shouldOpenDropDown = true;
-      return;
-    }
-
     super.showDropdown(...args);
   }
 
--- a/src/utils/ChoicesWrapper.js
+++ b/src/utils/ChoicesWrapper.js
@@ -68,7 +68,7 @@ class ChoicesWrapper extends Choices {
       return;
     }
 
-    this.shouldOpenDropDown = false;
+    this.shouldOpenDropDown = true;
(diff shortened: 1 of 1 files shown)


src/utils/ChoicesWrapper.js
```diff
@@ -68,7 +68,7 @@ class ChoicesWrapper extends Choices {
       return;
     }
 
-    this.shouldOpenDropDown = false;
+    this.shouldOpenDropDown = true;
```
travist, line 71: This change claims that this variable is not needed. You may as well just delete all references to it since it checks to see if it is false later and then sets it to true. I can't imagine that this was added for nothing. I would need to consult with the developer who introduced this to see if this is alright to remove.

pavlkara1: Let me know if it's ok to remove this. If yes, I will proceed with the change you suggested :v:

pkarakal: Hey. Do you have any news on this MR? Should I close this PR?

travist: I have a hard time believing we would have added this ```shouldOpenDropDown``` flag for nothing. Your change pretty much rips it out, so I do have some concerns that this will break the original need for it. Go ahead and change this pull request to remove this entirely and we may pull this into a test build to try it out.

travist:
This is related to https://github.com/formio/formio.js/pull/2965.

pkarakal:
I have went head and removed any reference to `shouldOpenDropDown` variable as you requested.

To check this I transpiled the application running `npm run transpile`, linked it with one of my projects that use this and use the transpiled version to run the application. Nothing seemed broken and the dropdown opened on the first click.

travist:
We found some issues with removing this. @AlexeyDavyd is working on another solution.

PR #2810, opened 2020-05-18
FJS-907

Tkachez:
If there is calculated value, it will override select component's value that is inactive


QThans:
If there are no erros.please fix them as soon as possible.this bug has much impact

PR #2818, opened 2020-05-20
Feature/quill table

vlad-shusterman:
Added 2 dependencies:
``quill-table`` is module for tables
and ``quill`` is just in need for ``quill-table``

docs for ``quill-table`` - https://developer.aliyun.com/mirror/npm/package/quill-table-module

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
     "moment-timezone": "^0.5.28",
     "native-promise-only": "^0.8.1",
     "quill": "^2.0.0-dev.3",
-    "quill-table": "^1.0.0",
     "resize-observer-polyfill": "^1.5.1",
     "signature_pad": "^2.3.2",
     "string-hash": "^1.1.3",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,8 @@
     "moment": "^2.25.3",
     "moment-timezone": "^0.5.28",
     "native-promise-only": "^0.8.1",
+    "quill": "^2.0.0-dev.3",
+    "quill-table": "^1.0.0",
     "resize-observer-polyfill": "^1.5.1",
     "signature_pad": "^2.3.2",
     "string-hash": "^1.1.3",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "moment": "^2.25.3",
     "moment-timezone": "^0.5.28",
     "native-promise-only": "^0.8.1",
-    "quill": "^2.0.0-dev.3",
+    "quill": "^2.0.0-dev.2",
     "quill-table": "^1.0.0",
     "resize-observer-polyfill": "^1.5.1",
     "signature_pad": "^2.3.2",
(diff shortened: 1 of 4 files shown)


src/components/_classes/component/Component.js
```diff
@@ -7,12 +7,13 @@ import isMobile from 'ismobilejs';
...
 import * as FormioUtils from '../../../utils/utils';
 import Validator from '../../../validator/Validator';
+import TableModule from 'quill-table';
 import Templates from '../../../templates/Templates';
 import { fastCloneDeep, boolValue } from '../../../utils/utils';
 import Element from '../../../Element';
 import ComponentModal from '../componentModal/ComponentModal';
 const CKEDITOR = 'https://cdn.form.io/ckeditor/16.0.0/ckeditor.js';
-const QUILL_URL = 'https://cdn.form.io/quill/1.3.7';
+const QUILL_URL = 'https://cdn.quilljs.com/2.0.0-dev.2';
```
travist, line 16: seems risky to include a "dev" build of quill into our renderer.

vlad-shusterman: @travist 
This ``dev-2`` is the most stable from their ``devs``.
Unfortunately they still haven't released 2nd version and it is the only way how to provide tables for users.

travist: Is there any way we could lazy-load the table module?

vlad-shusterman: @travist no, didn't found any possibility to lazy load module.

vlad-shusterman: @travist Made it as option for textarea component.

src/components/_classes/component/Component.js
```diff
@@ -7,12 +7,13 @@ import isMobile from 'ismobilejs';
 import Formio from '../../../Formio';
 import * as FormioUtils from '../../../utils/utils';
 import Validator from '../../../validator/Validator';
+import TableModule from 'quill-table';
```
travist, line 10: Why can we not lazy load this?....

https://github.com/MrMadClown/quill-table/blob/master/dist/quill-table.js

We can add this to our CDN so that it is lazy-loadable, and the following should work.

```js
return Formio.requireLibrary('quill', 'Quill', _.get(this.options, 'editors.quill.src', `${QUILL_URL}/quill.min.js`), true)
      .then(() => {
return Formio.requireLibrary('quill-table', 'TableModule', `${QUILL_TABLE_URL}/quill.table.js`, true)
      .then(() => {

});
});
```

package.json
```diff
@@ -78,6 +78,7 @@
     "moment": "^2.25.3",
     "moment-timezone": "^0.5.28",
     "native-promise-only": "^0.8.1",
+    "quill": "^2.0.0-dev.3",
```
travist, line 81: We don't need this as a dependency now.

travist: I'll take care of it.

vlad-shusterman: @travist ok, thanks!

PR #2830, opened 2020-05-22
fixed unable to submit form with dataMap inside dataGrid/editGrid (ticket - FJS-951)

TanyaGashtold:


travist:
Please write some tests.

TanyaGashtold:
I added test

PR #2831, opened 2020-05-22
Fix/uip validation error display

mikekotikov:


travist:
Please write some tests

PR #2836, opened 2020-05-22
PDF 72 - Fixed an issue of PDF instances duplication and messy Formio.forms

dianatsevan:


src/Webform.js
```diff
@@ -918,13 +918,18 @@ export default class Webform extends NestedDataComponent {
...
+  destroy(deleteFromGlobal = false) {
     this.off('submitButton');
     this.off('checkValidity');
     this.off('requestUrl');
     this.off('resetForm');
     this.off('deleteSubmission');
     this.off('refreshData');
+
+    if (deleteFromGlobal) {
+      delete Formio.forms[this.id];
```
travist, line 930: What is wrong with just always deleting it from the global? Basically no need to pass around a flag to destroy.

dianatsevan: Because the instance won't be exist at all. The goal is to delete it only when it is necessary. The instances duplication happens when we are moving from edit tab to use and vise versa that is why new instances are created.

travist: But the only way to come back from a "destroy" is to re-create which will just create a new instance.  Isn't it a memory leak if we don't always delete it from here?

dianatsevan: Оn the contrary right now we have the only instance that we work with instead of what we had. Formio.forms was always being filled messy and unnecessary instances weren't deleted. Currently we leave just the only instance that we need and all other would be deleted.

PR #2840, opened 2020-05-26
Fixed issue where Promise collides with Angular ZoneAwarePromise

mohlendo:
Fixes an error in angluar-formio: https://github.com/formio/angular-formio/issues/552


travist:
Thank you @mohlendo!  I will get a new angular-formio released asap with this in place.

PR #2845, opened 2020-05-26
PDF 40 - Fixed element's location on its adding

dianatsevan:


randallknutson:
In the future, please write tests.

PR #2846, opened 2020-05-26
PDF 41 - Removed red asterisk from readOnly mode and submission as PDF

dianatsevan:


dianatsevan:
> This should only apply to PDFs, not every form. Please update so that it only work applies in pdf mode.

Done.

PR #2857, opened 2020-05-29
FJS-968: Fixed `eachComponent` including `htmlelement` components when `includeAll` is not provided

Alena-Levina:


Alena-Levina:
@travist I've added a test which should have failed because I did not add the fix yet, but I see that it's still passing on CircleCI: https://app.circleci.com/pipelines/github/formio/formio.js/482/workflows/c8e386a1-b96e-413d-b5a7-bd7b1cffe202/jobs/6873/steps 
And I also see no logs that any test were executed: https://monosnap.com/file/56COV8wGvwAWJUYk0JZfuK5Ic7Ksnq

Am I missing something, or is something really wrong with our CircleCI pipeline?

Alena-Levina:
@travist I've added the fix so now it should pass, but on previous check it should not and also as I understand it should log executed tests, so my questions above are still actual

Alena-Levina:
@travist I also added fix and tests for `content` components, so feel free to merge this PR

PR #2861, opened 2020-06-01
PDF 43 - Fixed an issue of disability to drag components over on Firefox

dianatsevan:


dianatsevan:
Supported by all browsers.

PR #2862, opened 2020-06-01
Row draft

alexandraRamanenka:


src/components/editgrid/editForm/EditGrid.edit.validation.js
```diff
@@ -16,5 +16,13 @@ export default [
     type: 'number',
     tooltip: 'The maximum length requirement this field must meet.',
     input: true
+  },
+  {
+    type: 'checkbox',
+    input: true,
+    weight: 105,
+    key: 'draft',
```
travist, line 24: I don't want to use the name ```draft``` as the property here since this makes it seem like the component JSON itself is in "draft" mode. I would also like to reserve this work for possible future use.  Can we call this "rowDrafts"? and then the label will say "Enable Row Drafts"

```
{
  type: 'checkbox',
  input: true,
  weight: 105,
  key: 'rowDrafts',
  label: 'Enable Row Drafts',
  tooltip: '--- same as you have ---'
}
```

PR #2865, opened 2020-06-02
S3 photo upload - update File.js

oneserve-alh:
Fix for uploading photos to S3 from mobile devices.


oneserve-alh:
Resolves #2865

travist:
Looks good, thanks!

travist:
I just realized that this forces all files to be named "photo-......png".  We will need to revert this until we figure out a way to make it work for all file types.

travist:
You will need to create a new Pull Request with this + fixes.

oneserve:
Created PR #2874 which includes this plus:

- Set blob.name to file.name so that the original naming convention is used.
- Removed unused "evt" param'.
- Removed unnecessary console.log().

PR #2869, opened 2020-06-02
fixed address component losing settings if the form was saved on next…

TanyaGashtold:
… and viewed on edge (ticket - FJS-933)


travist:
We decided to allow a breaking change between 4.9.x and 4.10.x for the address component due to the reason that it was "kind-of" broken anyway in 4.9.x and we do not suspect many people would have it in place. I am going to revert this logic to keep the code clean.

PR #2870, opened 2020-06-02
PDF 008 - Fixed to correctly set the default value of CheckboxAsRadio

dianatsevan:


src/components/checkbox/Checkbox.js
```diff
@@ -30,7 +30,9 @@ export default class CheckBoxComponent extends Field {
   }
 
   get defaultValue() {
-    return this.component.name ? '' : (this.component.defaultValue || false).toString() === 'true';
+    const { name } = this.component;
+
+    return name ? (this.component[name] || this.emptyValue) : (this.component.defaultValue || false).toString() === 'true';
```
travist, line 35: ```this.component[name]``` seems weird to me. This seems like you will return the default value of some random property of the component JSON?

dianatsevan: Think I should rename it to radioGroupName. The value of this property is the default value set to the radio group.

travist: Oh nevermind. I see that before you are doing ```const { name } = this.component```.  I thought that ```name``` could be any random string.

PR #2874, opened 2020-06-02
Correct image names on S3 upload

oneserve-alh:
Set blob.name to file.name so that the original naming convention is used.
Removed unused "evt" param'.
Removed unnecessary console.log().

### DIFF
--- a/src/components/file/File.js
+++ b/src/components/file/File.js
@@ -363,9 +363,9 @@ export default class FileComponent extends Field {
           window.resolveLocalFileSystemURL(success, (fileEntry) => {
               fileEntry.file((file) => {
                 var reader = new FileReader();
-                reader.onloadend = function(evt) {
+                reader.onloadend = function() {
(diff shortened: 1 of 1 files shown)


travist:
Looks good. Thanks!

src/components/file/File.js
```diff
@@ -362,7 +362,13 @@ export default class FileComponent extends Field {
         webViewCamera.getPicture((success) => {
           window.resolveLocalFileSystemURL(success, (fileEntry) => {
               fileEntry.file((file) => {
-                this.upload([file]);
```
oneserve, line 365: @travist @oneserve-alh 

Will below change work ?
Wouldn't "this" in onloadend function be assigned to FileReader context ?

oneserve-alh: You are right, the scope isn't correct. I'll verify and change to;

```
  reader.onloadend = (evt) => {
      var blob = new Blob([new Uint8Array(evt.target.result)], { type: file.type });
```

PR #2911, opened 2020-06-11
Add teext overflow for long select items

AlexeyDavyd:


travist:
What would this do for Tags component? This seems like it would not allow multiple tags per line.

PR #2916, opened 2020-06-11
PDF 10 - search and removal of the component on all nesting levels

dianatsevan:


travist:
We will need to wait to pull this in until we are ready to test another RC release.

PR #2922, opened 2020-06-12
fixed editGrid data display inside wizard in readOnly mode (ticket - FJS-1005)

TanyaGashtold:


src/components/editgrid/EditGrid.js
```diff
@@ -60,7 +60,7 @@ export default class EditGridComponent extends NestedArrayComponent {
     return `<div class="row">
   {% instance.eachComponent(function(component) { %}
     <div class="col-sm-2">
-      {{ component.getView(component.dataValue) }}
+      {{ component.getView(getRowComponentValue(component)) }}
```
travist, line 63: We cannot be cavalier about changing the defaultRowTemplates for EditGrid. It seems that this is indicative of a different problem. Why is ```component.dataValue``` not correct here?

PR #2932, opened 2020-06-15
FJS-1011: Fixed select component with JSON data source always having empty object as default value

Alena-Levina:


src/components/select/Select.js
```diff
@@ -106,6 +106,23 @@ export default class SelectComponent extends Field {
...
+      else {
+        firstValue = firstItem;
+      }
+      if (typeof firstValue === 'object' || !firstValue) {
+        return {};
+      }
+      else {
+        return '';
+      }
+    }
```
Alena-Levina, line 125: @travist could you please take a look at this logic and check if I didn't miss anything from our discussion on Friday?

src/components/select/Select.js
```diff
@@ -106,6 +106,23 @@ export default class SelectComponent extends Field {
...
+    if (this.component.dataSrc === 'json' && this.component.data.json) {
+      const firstItem = this.component.data.json[0];
+      let firstValue;
+      if (this.valueProperty) {
+        firstValue = _.get(firstItem, this.valueProperty);
+      }
+      else {
+        firstValue = firstItem;
+      }
+      if (typeof firstValue === 'object' || !firstValue) {
```
travist, line 119: It would be better to make the logic.

```js
if (firstValue && (typeof firstValue === 'string')) {
  return '';
}
return {};
```

Checking for typeof string is more deterministic since there are many things that give you a typeof object.

Alena-Levina: @travist changed as you suggested

PR #2939, opened 2020-06-16
Fix (Wizard): display panel navigation buttons according to page options

### LINKED ISSUE #2934: [BUG] "Panel Navigation Buttons" checkboxes have no effect
## Environment
Tested on https://formio.github.io/formio.js/app/builder with Google Chrome

## Steps to Reproduce
1. Create a Wizard
2. Open the settings of a page/panel
3. Uncheck the checkboxes under the label "Panel Navigation Buttons"

## Expected behavior
When a checkbox is unchecked, the related button should not appear in the generated form.

## Observed behavior
Unchecking the checkboxes has no effect. The buttons remain visible even if all checkboxes are unchecked.

## Explaination 
I managed to locate what may be the source of the bug in the file [Wizard.js](https://github.com/formio/formio.js/blob/master/src/Wizard.js#L556) : 

Each checkbox value is accessed with the instruction
`_.get(currentPage, 'buttonSettings.******')`,
but the actual values are in
`_.get(currentPage, 'component.buttonSettings.******')`

Replacing [the line 556](https://github.com/formio/formio.js/blob/master/src/Wizard.js#L556) by `const currentPage = this.currentPage.component;` is actually enough to fix the bug, but is it a valid fix ?

Baloche:
Closes #2934

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -552,37 +552,27 @@ export default class Wizard extends Webform {
     ]);
   }
 
-  hasButton(name, nextPage) {
-    const currentPage = this.currentPage;
-    if (name === 'previous') {
-      const show = firstNonNil([
-        _.get(currentPage, 'component.buttonSettings.previous'),
-        this.options.buttonSettings.showPrevious
-      ]);
-      return (this.getPreviousPage() > -1) && show;
+  hasButton(name, nextPage = this.getNextPage()) {
+    // get page options with global options as default values
+    const {
+      previous = this.options.buttonSettings.showPrevious,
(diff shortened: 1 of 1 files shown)


travist:
Thank you for the fix. This would be more performant if you made it like this.

```js
const buttonSettings = _.get(this.currentPage, 'component.buttonSettings');
```

Then just remove the other _.get methods referencing the buttonSettings elsewhere in the function.

Baloche:
I did a little bit more than what you asked, in an attempt to make the function more readable.
I hope it helps :)

travist:
Well done!  Thanks for the contribution.

PR #2945, opened 2020-06-16
Allow non-truthy computed values (e.g. zero).

nathanieltagg:
When computing a value, the code would check for false (falsey) values, and if so use the empty Value.

This is incorrect behavior: for example:
     value = data.a - data.b
should yield 0 if we enter the same number into both the 'a' and 'b' fields.  Instead, the field simply doesn't update, or updates to the incorrect 'emptyValue' entry.

Instead of checking for false, now compares explicitly to null or undefined (so that zero and false are possible valid calculated results)


travist:
Thank you for the contribution!  No need to commit the "dist" folder changes for pull request. You just need to include the "src" folder changes.  I will try to manually merge this and accept it, while also giving you attribution in the git logs.

nathanieltagg:
Whups, sorry! I thought I had done that!  I'm still new to pull-requests.  (I guess I submitted a branch instead of a tag)
— Nathaniel

PR #2947, opened 2020-06-16
VPAT-394/391 Added callbacks to further extension in VPAT

antonSoftensity:


travist:
We have a hook system in our renderer that would probably be better here.  You could use it like the following.

```js
this.hook('onClose');
```

Then, anyone can hook into this method by adding an option within hooks.

```js
hooks: {
  onClose: () => {
    // Called when this hook is executed.
  }
}
```

This can also be used asynchronously (see ```beforeSubmit```) and also be used to manipulate variables and state.

PR #2952, opened 2020-06-17
FJS-1002: Fixes an issue when components comparing by getter and property

RomanLetsuk:


travist:
Please write tests

travist:
This was my fault for not writing tests. I will pull it through and write some tests for it.

PR #2953, opened 2020-06-17
Fix/validation error messages contain html chars

mikekotikov:
Relates to the issue discover in angular material.
Before:
image

After:
image

Is safe for XSS attacks, works for IE since 10th version.


travist:
Please write tests

PR #2958, opened 2020-06-17
added consistent return for setToken

airarrazaval:
Currently `Formio.setToken` returns either void or a Promise, making it bothersome and having to wrap the setToken call in a promise in order to make it consistent.

This makes sure to always return a Promise and return either `null` (if token has been deleted) or the current user, always updating if the current user is `null`.

Other small fixes includes:

* `static get token()`: code refactoring
* `static set token(token)`: removed return value

### DIFF
--- a/src/Formio.js
+++ b/src/Formio.js
@@ -976,7 +976,7 @@ export default class Formio {
       Formio.tokens = {};
     }
 
-    return Formio.tokens.formioToken ? Formio.tokens.formioToken : '';
+    return Formio.tokens.formioToken || '';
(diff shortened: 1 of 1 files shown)


travist:
It seems this has caused some regressions in our portal where we can no longer log into the portal. We may need to revert this so that it does not block our next release.

travist:
Here is the fix.  https://github.com/formio/formio.js/pull/2978

What you had would not set the token if the previous one was an empty string (which is the initialized state)

PR #2965, opened 2020-06-18
Fix select dropdown does not open on first click after value clear

AlexeyDavyd:


src/utils/ChoicesWrapper.js
```diff
@@ -68,7 +68,6 @@ class ChoicesWrapper extends Choices {
       return;
     }
 
-    this.shouldOpenDropDown = false;
```
travist, line 71: This completely negates the need for this variable. For code cleanup, can we just remove it all other places it is referenced.

Issue #2948, opened 2020-06-16
Calculated values aren't checked for validation [BUG] 


nathanieltagg:
Formio.js (@latest, checked against 4.10.2) 

Computed fields apparently are not checked for validity.

Here is a fiddle: 
https://jsfiddle.net/nathanieltagg/n37hxypd/10/

The calculated value shows up, but no matter what entries are put in for 'a' and 'b', it never shows an error in the computed field.

nathanieltagg:
At first I thought the isDirty flag should be set in checkData() if the calculated value changes.  But that doesn't work. When the independent field is changed, it triggers a change events which runs the calculation and checkData, but then something apparently triggers it to run again, and on the second checkData() pass, the value is now no longer dirty and the message clears.

I tried to put in a check so that if this.calculatedValue was defined, the isDirty flag was always set (forcing the validation), but then I run into the problem that if the form is empty, the calculatedValue is often set to NaN instead of undefined (as of my most recent submitted patch allowing non-truthy calculated values).

This reveals another problem: it is possible to have NaN as a computed value (i.e. the form divides two numeric fields, resulting in a divide-by-zero situation) which SHOULD be checked for validation (and fail such checks if the input values are bad).   But it should NOT be checked for validation if it comes about because some of the input fields are as-yet undefined.

I know I certainly don't follow the validation logic; I can't see why checkData is called more than once after a change, so I'm really just tinkering with something I don't really understand.

airarrazaval:
I think this is the expected behavior as what you are saying is to trigger a validation error on a calculated value using values of other components.  The correct way to do it is to put the validation on the component that depends of the calculated one.

In your example you have:

* Number Component A
* Number Component B
* Number Component Result (disabled and calculated as A - B)

In this case the validation should be set in B, with the following condition:

```js
valid = input < data.a ? true : 'B should be less than A';
```

Maybe it is not so obvious using number components, but imagine you have the following:

* Text Field Component `First Name`
* Text Field Component `Last Name`
* Text Field Component `Full Name` (disabled and calculated as `First Name` + `Last Name`)

Suppose you want to have a valid `Full Name` (i.e. no numbers).   What you are proposing is that it doesn't matter if `Last Name` has invalid characters, but `Full Name` should not.  As you can see, this doesn't make sense and you should put validations on components that are editable by the user (which is the input you should be watching, not the one you are calculating).

nathanieltagg:
[...]

PR #2968, opened 2020-06-18
Attempt to validate fields that are calculated.

nathanieltagg:
Attempt to fix bug 2948, but it is rather ham-fisted.

If a Component has a this.calculatedValue, then always check for validity.
Exception: don't do it if the calculatedValue is NaN, which happens frequently if the form is empty. 

But that isn't really perfect, because you might want to validate a case where
value = data.a/data.b
which would generate an NaN that you want to validate against.

### DIFF
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -2577,7 +2577,7 @@ export default class Component extends Element {
     if (flags.noCheck) {
       return true;
     }
-    this.calculateComponentValue(data, flags, row);
+    const calcChanged = this.calculateComponentValue(data, flags, row);
     this.checkComponentConditions(data, flags, row);
     if (flags.noValidate && !flags.validateOnInit) {
       return true;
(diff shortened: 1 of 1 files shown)


travist:
This change kind of scares me. I will add this to our Monday discussion to talk through what the right approach to solving this issue will be.

nathanieltagg:
Fair enough - as I said, it's a hack where I don't really understand the cause of the problem. It's a workaround for me in the meantime.

wag110894:
@nathanieltagg, do you not agree with @airarrazaval on his comment that this is working as expected: https://github.com/formio/formio.js/issues/2948

nathanieltagg:
@wag110894  I would say that it may be working as intended, but not as (I) expected.  If calculated values cannot be validated, this should be disabled in the builder menu and documented.

But I would prefer just to make it work, because I see no reason, in principle, why it shouldn't work.

src/components/_classes/component/Component.js
```diff
@@ -2590,6 +2590,13 @@ export default class Component extends Element {
       !this.isEmpty(this.defaultValue) &&
       this.isEqual(this.defaultValue, this.dataValue);
 
+    // We want to always validate calculated values...
+    //  .. unless they haven't recently changed, and are NaN, which is frequently the case when the form is empty.
+    if (this.calculatedValue !== undefined && !(Number.isNaN(this.calculatedValue) && !calcChanged)) {
```
travist, line 2595: I am ok with this change as long as we do not make an assumption that calculated values should be numbers. You can create a calculated value as any data type. The name should really be computedValue

travist (changes requested):

travist (changes requested):
Requesting changes before we can pull this in. Please remove assumption calculated values must be numbers.

travist:
Closing due to lack of activity. We have also recently refactored calculated values (and we have not heard of any other customers complain about this), so I assume it is no longer a pressing issue.

PR #2972, opened 2020-06-19
Use this.i18next instance instead of imported module

shawnbot:
We've been trying to render multiple instances of the same form in different languages, but after digging around I discovered that Webform mutates the imported `i18next` module in a bunch of places rather than using its own instance (`this.i18next`), which can be passed in via `options.i18next`.

This adds a test for mutation of the imported `i18next` when Webform gets `options.i18next`, and updates the Webform class to always refer to its own instance.


travist:
Looks good! Thanks!

PR #2973, opened 2020-06-19
Improve the way oauth works to work with new server provider.

randallknutson:


src/components/button/editForm/Button.edit.display.js
```diff
@@ -34,6 +34,25 @@ export default [
...
     },
   },
+  {
+    type: 'select',
+    key: 'oauthProvider',
+    label: 'OAuth Provider',
+    input: true,
+    dataSrc: 'values',
+    weight: 111,
+    tooltip: 'The oauth provider to use to log in (8.x server only).',
```
travist, line 44: Doesn't 8.x server include the ```form.schema``` version?  Could be add a conditional to only show this field if the ```form.schema``` is above a certain version?

randallknutson: form.schema is not related to the server version so we can't rely on that. What I'd like to do is set this field as hidden as part of the portal app build for 7.x.

PR #2977, opened 2020-06-21
noNewModal = true to skip new component modal

besmiralia:
Add this flag to the predefined components to skip the modal popup.
Referring to the example:
https://formio.github.io/formio.js/app/examples/custombuilder.html

components: {
        firstName: {
          title: 'First Name',
          key: 'firstName',
          icon: 'terminal',
          schema: {
            label: 'First Name',
            type: 'textfield',
            key: 'firstName',
            input: true,
            **noNewModal: true**
          }
        },...

The components under the custom group can have the property noNewModal = true if you want to skip the modal.
Any other component that doesn't have the flag (ex. basic, data, etc) keep their regular behavior (modal popup when dragged/dropped on the form)

### DIFF
--- a/src/PDFBuilder.js
+++ b/src/PDFBuilder.js
@@ -293,7 +293,7 @@ export default class PDFBuilder extends WebformBuilder {
           width: schema.width
         };
 
-        if (!this.options.noNewEdit && !component.component.noNewModal) {
+        if (!this.options.noNewEdit && !component.component.noNewEdit) {
           this.editComponent(component.component, this.webform.iframeElement);
         }
         this.emit('updateComponent', component.component);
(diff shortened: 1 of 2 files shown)


travist:
We already have a flag for this called ```noNewEdit```.  You can add it to the builder like so.

```js
Formio.builder(document.getElementById('builder'), {}, {
  noNewEdit: true
});
```

besmiralia:
Hi @travist.
Yes, I noticed that flag, but once you initialize the builder with it, it applies to every component. What I created is a flag for a specific component.

Let's say, I created a custom group and there I have already predefined resource components(ex. FirstName, LastName, etc, just like the example). Now I don't want to waste time clicking save every time I drop them into the form, they are already predefined, there is nothing else to customize. But I want to leave the popup for new elements that I create, ex. new fields or maybe all layout components. Makes sense?

travist:
I am alright with this as long as we give it the same name ```noNewEdit```.  I would prefer to have a consistent name for a variable that does the same thing as the global option.

besmiralia:
Agreed, I'll make the change and resubmit

PR #2983, opened 2020-06-22
FJS-909: fixed dateTime renders suffix/prefix as string if form is saved in next/edge…

TanyaGashtold:
… and opened in portal


TanyaGashtold:
changes in Bootstrap3 and semantic templates:
https://github.com/formio/bootstrap3/pull/24
https://github.com/formio/semantic/pull/17

Issue #2627, opened 2020-04-08
[BUG] Multiple validation errors displayed


moeVdm:
## Environment
Please provide as many details as you can:

- Hosting type
  - [] Form.io
  - [z] Local deployment
    - Version: 
- Formio.js version: 4.7.8 => 4.9.9

We noticed this change going from 4.7.8 to 4.9.9.

## Steps to Reproduce
1. Create a wizard form with 2 pages
2. On the first page, add a text field with required validation
3. Add a custom validation, for example, valid = input.length > 0 ? true : 'should be longer';
(it doesn't make sense to do this instead of using the minimum length validation, but just for the sake of the example)
4. When using the form, do not enter anything in the text field and press "Next" to validate

## Expected behavior

Form validation should only show the error that the text field is required.

## Observed behavior

Form validation shows all validation errors:
- Text Field is required
- should be longer

In 4.7.8, it would only show that the field is required. Is this behavior wanted or is it a bug?

## Example

Here I used the form builder example on your website to reproduce the issue.

<img width="1108" alt="Screen Shot 2020-04-08 at 3 52 22 PM" src="https://user-images.githubusercontent.com/52581920/78827522-3c88ec00-79b1-11ea-9bb3-eacaf8f3a939.png">

For code or form JSON, please enclose in a code block:

```js
{
    "display": "wizard",
    "settings": {
        "pdf": {
            "id": "1ec0f8ee-6685-5d98-a847-26f67b67d6f0",
            "src": "https://files.form.io/pdf/5692b91fd1028f01000407e3/file/1ec0f8ee-6685-5d98-a847-26f67b67d6f0"
        }
    },
    "components": [
        {
            "title": "Page 1",
            "label": "Page 1",
            "type": "panel",
            "key": "page1",
            "components": [
                {
                    "label": "Text Field",
                    "spellcheck": true,
                    "tableView": true,
                    "calculateServer": false,
                    "validate": {
                        "required": true,
                        "custom": "valid = input.length > 0 ? true : 'should be longer';"
                    },
                    "key": "textField",
                    "type": "textfield",
                    "input": true
                }
            ],
            "input": false,
            "tableView": false
        },
        {
            "title": "Page 2",
            "label": "Page 2",
            "type": "panel",
            "key": "page2",
            "input": false,
            "tableView": false,
            "components": []
        }
    ]
}
```

randallknutson:
[...]

PR #2984, opened 2020-06-22
Localize alert messages, simplify t() methods

shawnbot:
This adds a new `alertMessage` localization string for the messages listed in the alert element displayed when validations fail:

image

The default value was `{{label}}. {{message}}`, but I've swapped out the period for a colon (`{{label}}: {{message}}`) since it sounds like [the period was a typo](https://github.com/formio/formio.js/issues/2627#issuecomment-630862768) anyway. The string gets `label` and `message` placeholder values (which can easily be added to [here](https://github.com/shawnbot/formio.js/blob/4574491b6eab9ef2d112ac667686447fb4570c3b/src/Webform.js#L1185)). I've included a test that localizes it as just `{{message}}` so that the text is just the validation message without the redundant component label, which I believe addresses @darabi's suggestion in https://github.com/formio/formio.js/issues/2627#issuecomment-631239557. (I would suggest simply using `{{message}}` as the default, but that feels like a breaking change to me.)

### Simpler, better `t()` methods
In this PR I've also simplified the Element and Component classes' `t()` methods so that they don't set i18next options each time, and I've moved those options to the `i18n.js` option defaults object. It doesn't look there actually are any tests for these options, but I think that #2972 solves the bug that repeatedly passing these options was meant to work around. However, tests to make sure that both 1) the default options (particularly `nsSeparator`) are respected and 2) they can be overridden via `i18n: { ...options }` would probably be good to round this out.

The big (but not, AFAIK, breaking) change this introduces is that you can now call `t()` with an array of strings, each of which [i18next tries to localize and falls back on the next value](https://www.i18next.com/principles/fallback#calling-with-fallback-keys) if none is found. This makes it much easier to localize form labels and descriptions in custom templates like so:

```diff
- ctx.t(ctx.component.label)
+ ctx.t([`components.${ctx.component.key}.label`, ctx.component.label])
```

which you could localize more explicitly in your `i18n` option object like this:

```js
i18n: {
  es: {
    components: {
      firstName: { label: 'Nombre' },
      lastName: { label: 'Apellido' }
    }
  }
}
```

### DIFF
--- a/src/Element.js
+++ b/src/Element.js
@@ -367,17 +367,11 @@ export default class Element {
   /**
    * Translate a text using the i18n system.
    *
-   * @param {string} text - The i18n identifier.
+   * @param {string|Array<string>} text - The i18n identifier.
    * @param {Object} params - The i18n parameters to use for translation.
    */
-  t(text, params) {
-    params = params || {};
-    params.nsSeparator = '::';
-    params.keySeparator = '.|.';
-    params.pluralSeparator = '._.';
-    params.contextSeparator = '._.';
-    const translated = this.i18next.t(text, params);
-    return translated || text;
+  t(text, ...args) {
+    return this.i18next.t(text, ...args);
   }
 
   /**
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -3,6 +3,7 @@ import moment from 'moment';
 import compareVersions from 'compare-versions';
 import EventEmitter from './EventEmitter';
 import i18next from 'i18next';
+import i18nDefaults from './i18n';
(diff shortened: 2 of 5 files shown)


travist:
This is great, and thank you for your contribution!  If I understand you correctly, this is a non-breaking change right? If so, then we can probably try to get it in the next RC rounds (we are wrapping up our testing efforts on 4.10.5-rc now so I won't bring it in there).

randallknutson:
This looks really good. We'll look at getting this pulled in soon.

PR #2989, opened 2020-06-25
FJS-1023:Clear on refresh is clearing populated data on form load for select component

edwinanciani:


src/components/select/Select.js
```diff
@@ -517,8 +517,8 @@ export default class SelectComponent extends Field {
     this.setItems(this.getCustomItems() || []);
   }
 
-  refresh() {
-    if (this.component.clearOnRefresh) {
```
travist, line 521: You need to handle the case where ```instance``` is null.

PR #2992, opened 2020-06-25
WIP FJS-971: fix entering period as a first character in currency input

AlehKatsiubaFormIO:


src/components/currency/Currency.js
```diff
@@ -73,6 +73,22 @@ export default class CurrencyComponent extends NumberComponent {
       inputElement: input,
       mask: this.numberMask || '',
     });
+    input.mask.destroy();
```
travist, line 76: Why is this getting destroyed as soon as it is created? This seems strange to me.

AlehKatsiubaFormIO: It is destroyed because maskInput function adds event listener to the input element to update mask, but we don't need it anymore because we add another listener with if statement. This lib doesn't provide a way to configure input mask without adding event listener.

travist: We control this library, so it could be reasonable to just add a way to configure the input mask without adding an event listener.

AlehKatsiubaFormIO: Here I add an afterUpdate callback

PR #2996, opened 2020-06-26
Fixes an issue where utils.eachComponent functon was not iterating through columns components

alexandraRamanenka:
No trello card, noticed when was working on GCFund tasks.

### DIFF
--- a/src/utils/formUtils.js
+++ b/src/utils/formUtils.js
@@ -97,7 +97,7 @@ export function eachComponent(components, fn, includeAll, path, parent) {
     if (!noRecurse) {
       if (hasColumns) {
         component.columns.forEach((column) =>
-          eachComponent(column.components, fn, includeAll, subPath(), parent ? component : null));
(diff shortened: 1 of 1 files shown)


src/utils/formUtils.js
```diff
@@ -97,7 +97,7 @@ export function eachComponent(components, fn, includeAll, path, parent) {
     if (!noRecurse) {
       if (hasColumns) {
         component.columns.forEach((column) =>
-          eachComponent(column.components, fn, includeAll, subPath(), parent ? component : null));
```
travist, line 100: This just tells me that the component JSON is incorrect. Columns should be an array of objects that contain their own "components" property.

alexandraRamanenka: @travist , the compoent's JSON is correct. It is an array of objects that containt  their own "components" property. But the **instance** of the Columns component has it own 'columns' property where each column is jsut an array of components' instances within the column.

PR #3001, opened 2020-06-30
VPAT-498 Changed flatpickr to be lazy-loaded and removed dependencies

antonSoftensity:


antonSoftensity:
[VPAT-498 | Replace Flatpickr with lazy-loaded version](https://trello.com/c/ID2coylb/498-vpat-498-replace-flatpickr-with-lazy-loaded-version)

PR #3018, opened 2020-07-08
FJS-1032: fixed an issue where btn checks validity when disableOnInvalid is not set and removes error classes when form is still invalid

TanyaGashtold:


TanyaGashtold:
all tests passed locally

PR #3022, opened 2020-07-09
FJS-1043: fixed validation on blur inside a panel

TanyaGashtold:


TanyaGashtold:
all tests passed locally

PR #3025, opened 2020-07-10
FJS-1022: fixed an issue where row alert errors list is not updated when editing editGrid modal row

TanyaGashtold:


TanyaGashtold:
all tests passed locally

src/components/editgrid/EditGrid.js
```diff
@@ -501,10 +502,27 @@ export default class EditGridComponent extends NestedArrayComponent {
         }
         else {
           this.alert.showErrors(editRow.errors, false);
+          editRow.alerts= true;
         }
       },
     }, this.component.saveRow || 'Save'));
 
+    this.on('componentChange', () => {
```
travist, line 510: This may cause some serious performance issues.  I would rather do this within the ```checkData``` loop which is called every change cycle.

TanyaGashtold:
Changes are committed

PR #3030, opened 2020-07-10
VPAT-498 Changed flatpickr to be lazy-loaded and removed dependencies

antonSoftensity:


AnthonyPhan:
May I ask, what was the reason for lazy loading the flatpickr? would this not affect formio.js libraries ability to work offline?

PR #3031, opened 2020-07-10
VPAT-502 Added 'for' attributes for labels and IDs for inputs

antonSoftensity:


antonSoftensity:
[VPAT-502 | Label | Add 'for' attribute for labels across different template frameworks](https://trello.com/c/s8B6Vl3W/502-vpat-502-label-add-for-attribute-for-labels-across-different-template-frameworks)

PR #3034, opened 2020-07-13
Fixed the Javascript SDK example

Directoire:
The example that was provided use the wrong script file, causing errors.


travist:
Thank you!

PR #3039, opened 2020-07-14
FJS-1035: Fix Conditional add button for EditGrid

DmitryNistratov:


src/components/editgrid/EditGrid.js
```diff
@@ -204,6 +204,15 @@ export default class EditGridComponent extends NestedArrayComponent {
       return super.init();
     }
 
+    this.prevHasAddButton = this.hasAddButton();
+    this.on('change', () => {
```
travist, line 208: Instead of doing this, just add a handler in the ```checkData``` method to perform this task.

src/components/editgrid/EditGrid.js
```diff
@@ -209,6 +209,7 @@ export default class EditGridComponent extends NestedArrayComponent {
     }
 
     this.components = this.components || [];
+    this.prevHasAddButton = this.hasAddButton();
```
travist, line 212: Shouldn't this be moved to the NestedArrayComponent class now?  I am concerned that all nested array components except EditGrid will "redraw" on first pass of checkData now.

DmitryNistratov: I hope I got it right

PR #3042, opened 2020-07-14
VPAT-529 Fixed issue with focus moving to 1st error onchange

antonSoftensity:


antonSoftensity:
[VPAT-529 | Wizard | Each time validation is triggered on a Wizard, focus automatically moves to the 1st error in the error section](https://trello.com/c/tDtTm27Q/529-vpat-529-wizard-each-time-validation-is-triggered-on-a-wizard-focus-automatically-moves-to-the-1st-error-in-the-error-section)

PR #3052, opened 2020-07-16
FJS-1040: fix recalculation of row numbers for DataGrid

DmitryNistratov:


heather-formio:
[FJS-1040 (McKinsey): Removing Rows inside of a data grid does not re-order the numbers](https://trello.com/c/C0MELSnW/1040-fjs-1040-mckinsey-removing-rows-inside-of-a-data-grid-does-not-re-order-the-numbers)

PR #3053, opened 2020-07-16
FJS-1050: added hide label option for panel component

TanyaGashtold:


TanyaGashtold:
changed templates: 
https://github.com/formio/semantic/pull/19
https://github.com/formio/bootstrap3/pull/26

PR #3059, opened 2020-07-17
Feat/uip 239 select options overlayed

mikekotikov:
@travist please see if the name of the option and it's description are suitable.
What it does, it makes select dropdown options float over other elements by specifying position: fixed.


travist:
I am alright that you made this happen only if the parent is a table. That is a good compromise.

PR #3064, opened 2020-07-21
FJS-1080: fix focus for TextField with calendar widget

DmitryNistratov:


travist:
Wouldn't a better fix be to change this line https://github.com/formio/formio.js/blob/master/src/components/_classes/input/Input.js#L115 to this...

```js
this.originalComponent.suffix = this.component.suffix = convertStringToHTMLElement(calendarIcon, '[ref="icon"]');
```

travist:
Should be fixed with https://github.com/formio/formio.js/pull/3068

PR #3072, opened 2020-07-23
3.x - fixed conditional fields are shown/hidden in all dataGrid rows when o…

TanyaGashtold:
…nly one row meets the condition


TanyaGashtold:
3.x

PR #3078, opened 2020-07-24
FJS-1109: replaced scope with alternative selector for working in IE11

TanyaGashtold:


travist:
We can also now remove the polyfill that roma introduced to get around this issue....

TanyaGashtold:
added a commit to remove the polyfill

PR #3080, opened 2020-07-27
FJS-1041: fixed multiple redrawing when applying mergeComponentSchema action

TanyaGashtold:


src/components/_classes/component/Component.js
```diff
@@ -1591,7 +1591,7 @@ export default class Component extends Element {
       return;
     }
 
-    const newComponent = fastCloneDeep(this.originalComponent);
+    const newComponent = fastCloneDeep(this.component);
```
travist, line 1594: If I am not mistaken, ```originalComponent``` was created to deal with Logic. This basically says that it is no longer needed?...

randallknutson:
This will not work correctly. The whole point of originalComponent is to store the original configuration of the component so it can have a clean set of logic applied to it. There needs to be a different way to resolve this issue.

src/components/_classes/component/Component.js
TanyaGashtold: If we always use originalComponent for merging schema, than new component schema is always compared with initial schema and the redraw is triggered. In my case I need to make comparison with previously merged schema to avoid redrawing and to get the result that the client needs. Seems like mergeComponentSchema is just not appropriate for the purpose that the client wants to achieve. I found another way how to get what they want without mergeComponentSchema. I close this PR, please merge my new PR:  https://github.com/formio/formio.js/pull/3110

PR #3088, opened 2020-07-29
FJS-886: fix conditional logic when Checkbox's Input type is set to r…

DmitryNistratov:
…adio


src/utils/formUtils.js
```diff
@@ -55,7 +55,8 @@ export function eachComponent(components, fn, includeAll, path, parent) {
     const hasRows = component.rows && Array.isArray(component.rows);
     const hasComps = component.components && Array.isArray(component.components);
     let noRecurse = false;
-    const newPath = component.key ? (path ? (`${path}.${component.key}`) : component.key) : '';
+    const componentKey = component.name ? component.name : component.key;
```
travist, line 58: This seems unnecessary and also scares me a little that we need to modify this method (which has been extensively tested for many different scenarios).

PR #3089, opened 2020-07-29
FJS-1044: 'OnBlur' Valdation is not working for email and phone number inputs

AlehKatsiubaFormIO:


travist:
Tests are failing on this one.

travist:
This will have other undesired consequences. ```checkData``` also performs calculations on change as well as performs conditional logic. This will basically make it so that if a component has "onBlur" validation, it's calculations and conditionals will fail to evaluate.  This needs more work.

AlehKatsiubaFormIO:
> This will have other undesired consequences. `checkData` also performs calculations on change as well as performs conditional logic. This will basically make it so that if a component has "onBlur" validation, it's calculations and conditionals will fail to evaluate. This needs more work

But conditional logic also checks is this method was called by form submission and only in this case it may skip checking. If it was called by onblur event (without fromSubmission flag), checking will be in normal way.

travist:
My point is that by skipping the ```checkData``` method if a component has "onBlur" and also if it is from the submission, will skip over EVERYTHING including calculations, conditionals, validations, etc... which does not fit the description of what you are trying to fix where "OnBlur validation is not working for email input". This change will affect all components and is pretty intrusive. Please find another way to solve the problem.

AlehKatsiubaFormIO:
> My point is that by skipping the `checkData` method if a component has "onBlur" and also if it is from the submission, will skip over EVERYTHING including calculations, conditionals, validations, etc... which does not fit the description of what you are trying to fix where "OnBlur validation is not working for email input". This change will affect all components and is pretty intrusive. Please find another way to solve the problem.

Yeah, I understood.

PR #3109, opened 2020-08-03
Fixes an issue where value of non-persistent component is not calculated and shown in the View/Edit mode

alexandraRamanenka:
No Trello card, found during testing the PAP


travist:
Tests are failing... Seems you also need to update the render test templates.

alexandraRamanenka:
> Tests are failing... Seems you also need to update the render test templates.

Yes, with updated templates all tests are passed

PR #3114, opened 2020-08-05
PDF-126: Add property to get pdf label

RomanLetsuk:


travist:
There is no need to put this in the core renderer... just include this in the Component override of the pdf renderer.

PR #3117, opened 2020-08-06
FOR-2699: now passing groupId and groupPermissions to file upload

Tkachez:
That is done to check permissions on the server side when using File component.


src/Formio.js
```diff
@@ -515,7 +515,7 @@ export default class Formio {
     });
   }
 
-  uploadFile(storage, file, fileName, dir, progressCallback, url, options, fileKey) {
```
travist, line 518: We cannot change the signature of this function without a major version release. Just append the arguments to the end please.

Tkachez: @travist, I moved the argument to the beginning, since the number of arguments varies and breaks existing logic

PR #3121, opened 2020-08-06
Change the display of panels for nested wizards inside parent wizards

DmitryNistratov:


src/components/form/Form.js
```diff
@@ -94,8 +94,22 @@ export default class FormComponent extends Component {
     if (this.component.revision || this.component.revision === 0) {
       this.formSrc += `/v/${this.component.revision}`;
     }
-
-    return this.createSubForm();
+    return this.createSubForm().then(() => {
+      const findRoot = (comp) => {
```
travist, line 98: This should not be necessary. If you needed to do this, then this means that https://github.com/formio/formio.js/blob/master/src/components/form/Form.js#L339 is broken. You should just be able to use ```this.root``` and that will point to the most root element, even when nesting occurs. Please verify.

DmitryNistratov: Yes, you are right, thanks!
This is an extra piece of code left after the old version

Issue #2040, opened 2019-11-12
use custom datapicker


imohammad74:
Hello.
Can i use custom  datapicker (for jQuery) in form io?
I want use a Persian calendar  for my form.

Thanks.

randallknutson:
You can write your own custom components and override the functionality of our standard components. This should allow you to use your own date picker. To do this, create a module and override the datetime component and change the render and attach methods. Then use your own templates to create the html the way you want it. The attach method can be used to attach the datepicker to the html elements.

Here is more documentation:
https://github.com/formio/formio.js/wiki/Modules
https://github.com/formio/formio.js/wiki/Custom-Components-API
https://github.com/formio/formio.js/wiki/Templates-API

m-eydimorad:
Hello @imohammad74 
I need to use a Persian datepicker custom component in formio similar to your need, Did you have it or develop it? If yes, Could you please share your source code?

travist:
This requires that you translate the Flatpickr module @ https://flatpickr.js.org/localization/.  You would need to do this in the ```customOptions``` of the datetime component like this.

```
// Not sure what the import would be here, but this gives you an idea.
import { Persian } from "flatpickr/dist/l10n/.....js";

const dateTime = FormioUtils.getComponent(form.components, 'dateTime');
dateTime.customOptions = {locale: Persian};
Formio.createForm(document.getElementById('formio'), form);
```

ptvce:
@travist  
import { Persian } from "flatpickr/dist/l10n/.....js"   >>   this file just change the caption like name of days or months, there are many diffrence between persian calendar and the others.
for develop a custom component of persian calendar we need to change moment.js but how?  I don't know.
Could you please help?

travist:
@ptvce this would need to be a custom component where you implement a new widget, which are found @ https://github.com/formio/formio.js/tree/master/src/widgets.

ptvce:
@travist   could u please tell me how i can create new widget?
 in order to create a new widget i created a new widget beside input and calendar widget, then i added it in index.js that is beside other widget and then i created a custom component that extend DateTime.js component, but now i don't know how use my widget in my component?
these are my files:
[JalaliCalendarWidget.js.txt](https://github.com/formio/formio.js/files/4465975/JalaliCalendarWidget.js.txt)
[index.js.txt](https://github.com/formio/formio.js/files/4465978/index.js.txt)
[...]

PR #3129, opened 2020-08-10
Pass "locale" (not "language") Flatpickr setting

### LINKED ISSUE #3130: [BUG] Datetime components do not respect the form language
## Environment

- Formio.js version: 4.10.4

## Steps to Reproduce
1. Create a form with a Date/time component
2. Set a form's language to a value other than `en`, e.g. in its render options: `{lang: 'es'}`
3. Import the Flatpickr localizations for your language, e.g. from [unpkg.com](https://unpkg.com/flatpickr@4.6.6/dist/l10n/es.js)
4. Focus the date/time component to activate the Flatpickr widget

## Expected behavior
Flatpickr should display in the same locale (language) as the form.

## Observed behavior
Flatpickr displays in English.

## Example
[This CodePen](https://codepen.io/shawnbot/pen/dyMYyNa?editors=1010) demonstrates how the locale is not respected (left) unless you pass the datetime component `customOptions: { locale: 'es' }` (right):

image image

I've proposed a simple fix in #3129.

shawnbot:
Flatpickr doesn't do anything with the `language` option passed here:

https://github.com/formio/formio.js/blob/60a9cb2c7a7107c1f2a83ed4f93766e9d95dde91/src/components/datetime/DateTime.js#L93

It does, however, [respect the `locale` option](https://flatpickr.js.org/localization/).

Fixes #3130.

### DIFF
--- a/src/components/datetime/DateTime.js
+++ b/src/components/datetime/DateTime.js
@@ -90,7 +90,7 @@ export default class DateTimeComponent extends Input {
       timezone,
       displayInTimezone: _.get(this.component, 'displayInTimezone', 'viewer'),
       submissionTimezone: this.submissionTimezone,
-      language: this.options.language,
+      locale: this.options.language,
       useLocaleSettings: _.get(this.component, 'useLocaleSettings', false),
       allowInput: _.get(this.component, 'allowInput', true),
       mode: 'single',


travist:
The only issue I see with this is our "language" option is a string that should be the current language of the form. The ```locale``` for Flatpickr is actually an object that is imported from the Flatpickr languages. I am not sure if this answer is still relevant, but I did answer this in a previous comment on how this could be achieved by doing the following.

https://github.com/formio/formio.js/issues/2040#issuecomment-591543927

This may be for version 3.x so not sure if it will still work (but it may). Regardless, this should probably be done in an easier way.

travist:
In that case, I approve! Thanks for the contrib @shawnbot!

PR #3138, opened 2020-08-12
Initialize Column with only 1 element

besmiralia:
The column component should be initialized with 1 element instead of two, because otherwise you won't be able to remove and leave it with only 1 column

### DIFF
--- a/src/components/columns/Columns.js
+++ b/src/components/columns/Columns.js
@@ -8,8 +8,7 @@ export default class ColumnsComponent extends NestedComponent {
       key: 'columns',
       type: 'columns',
       columns: [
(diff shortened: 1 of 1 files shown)


travist:
I don't like the side effect of this pull request. Where now if you drag and drop a columns component, it defaults to just one column and they would have to know to add a new columns and then change the widths to 6 each. I believe you are working to fix the bug where you cannot get the builder to allow only one columns. I would rather fix that in another way than do this.

besmiralia:
Thank you for the review. I see your point about UX here. I'll try to figure out another way to allow for less columns than the default.

PR #3144, opened 2020-08-13
DataGrid: fix noFirstRow option

mikekotikov:


src/components/datagrid/DataGrid.js
```diff
@@ -498,8 +504,10 @@ export default class DataGridComponent extends NestedArrayComponent {
     }
 
     const changed = this.hasChanged(value, this.dataValue);
-    this.dataValue = value;
-    this.createRows();
+    if (this.initRows) {
+      this.dataValue = value;
+      this.createRows();
+    }
```
juliapath, line 510: What is the purpose of this? This causes `setValue` to misbehave is noFirstRow/initEmpty is enabled, i.e. it does not add/remove rows, if the number of elements in the array is changed. This is still the case in the current version. Commenting here, because this is where this was introduced, so maybe there is some insight into this?

Issue #3145, opened 2020-08-13
[BUG] Behaviour for disabled text area is different between .js, Angular, and React versions


StephenOTT:
## Environment
Please provide as many details as you can:

- Hosting type
  - [x] Form.io
  - [x] Local deployment
    - Version: latest
- Formio.js version: latest
- Frontend framework: js, react, angular
- Browser: chrome
- Browser version: latest

When you have a disabled text area in Angular, the render is a disabled text area.  But when you have a disabled text area in formio.js it renders as a single line div. 

I would generally expect that the text area should render as a text area but in disabled mode.

Is this a bug or expected?  If expected then how is this supposed to be used?  as you lose text area resize and line breaks when it renders as a div

randallknutson:
This probably has to do with the different ways to disable a form and how they are handled by the different frontend libraries. Ideally they would all be the same but may have subtle differences. For example, if you set the form to disabled it will render each component in a disabled state. However, some renderers with actually switch the renderMode to "html" instead of "form" which causes an html version of the fields to be rendered which would result in the div instead of a wysiwyg. This issue is either that or the fact that angular may not be loading the wysiwyg library by default in disabled mode and so causes it not to attach properly.

StephenOTT:
Angualr loads the text area as a disabled textarea.  But it is formio.js and React that render it as a single line div (line breaks are lost).

wag110894:
Hello @StephenOTT. Thank you for reaching out! 

We have created a ticket for us to investigate why the disabled textareas are showing differently using Javascript, Angular, and React.  This item has been added to our backlog. We welcome a pull request or have a process to expedite as well. The ticket number is FJS-1167. Please let me know if you are interested in expediting and we can get the process started.

We will let you know once this issue has been resolved.

Thank you.

Tatsiana8:
Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #3179, opened 2020-08-25
FJS-1167-Fix-textarea-readonly-disable-rendering-behavior

GrellierMathieu:
Hello,

Regarding issue FJS-1167: https://github.com/formio/formio.js/issues/3145

I propose this fix (remove specific rendering when textarea is in readOnly or disable). As the component have the disabled property by default when disable or readOnly, there is no need for a custom rendering for the component.

Hope this will help.
Regards,
Grellier Mathieu

### DIFF
--- a/src/components/textarea/TextArea.js
+++ b/src/components/textarea/TextArea.js
@@ -64,13 +64,6 @@ export default class TextAreaComponent extends TextFieldComponent {
     const info = this.inputInfo;
     info.attr = info.attr || {};
     info.content = value;
-    if (this.options.readOnly || this.disabled) {
-      return this.renderTemplate('well', {
-        children: '<div ref="input" class="formio-editor-read-only-content"></div>',
-        nestedKey: this.key,
-        value
-      });
-    }
 
     return this.renderTemplate('input', {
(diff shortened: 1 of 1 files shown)


travist:
We will consider this, but it may break PDF's. We will need to do some testing on our end. Regardless, thank you for the contribution.

StephenOTT:
@travist Is this considered functional at the moment?  It seems that the readonly view for a text area is broken in general, as it does not render new line breaks.   The new line breaks are still in the source, but the readonly/disabled view of the text area does not render correctly.

travist:
Line breaks in PDF's is being solved with other fixes to our PDF system so they are getting resolved in another way. If we convert read only text area's to inputs, then this would also break use cases when users use Text Area with WYSIWYG (it would show HTML instead of render the result in the readonly view).  I believe we are going to close this PR assuming that we will fix the line break issue with other fixes (which we are already addressing in our latest PDF server builds).

travist:
Here is the pull request that fixes line breaks for Text areas.  https://github.com/formio/formio-viewer/pull/20

PR #3180, opened 2020-08-25
FJS-1187: fixed an issue where today date is not chosen on first click in calendar

TanyaGashtold:


TanyaGashtold:
all tests passed locally

PR #3184, opened 2020-08-25
FJS-1136: Fixes an issue where all the components inside a Nested Form are editable when Disabled is checked

alexandraRamanenka:


src/components/form/Form.js
```diff
@@ -94,6 +94,11 @@ export default class FormComponent extends Component {
     if (this.component.revision || this.component.revision === 0) {
       this.formSrc += `/v/${this.component.revision}`;
     }
+
+    if (this.component.disabled) {
+      this.options.readOnly = true;
```
travist, line 99: I believe this.options is a reference that all components share. I worry about side effects here. Will you confirm if that is the case? It may be that we clone the options at some point in component creation, but I don't believe so.

alexandraRamanenka: Do you mean that options are shared between all the components of the Nested Form? Even if so, it is correct, because if we set Nested Form to disabled, then every component inside it should be disabled.

travist: It is also shared with components outside of the Form component since it is all references. What I am worried about is if just the form component is disabled, that it will cause all fields to be read only in the entire form (since they all share the same options). I suppose a test would be to have a disabled form, and then force a redraw() of the form maybe a second later. If all fields end up in readOnly mode, then this fix has some pretty serious side effects.

alexandraRamanenka: > It is also shared with components outside of the Form component since it is all references. What I am worried about is if just the form component is disabled, that it will cause all fields to be read only in the entire form (since they all share the same options). I suppose a test would be to have a disabled form, and then force a redraw() of the form maybe a second later. If all fields end up in readOnly mode, then this fix has some pretty serious side effects.

Got it, will do it in another way

alexandraRamanenka: I checked it using more complex form, but only the components inside disabled Nested Form were readOnly. I still changed it and set readOnly property to the subForm options only, though.

PR #3185, opened 2020-08-25
modified has nested field logic in Select.js

Tkachez:


src/components/select/Select.js
```diff
@@ -195,7 +195,7 @@ export default class SelectComponent extends Field {
 
     if (data.data) {
       // checking additional fields in the template for the selected Entire Object option
-      const hasNestedFields = this.component.template.match(/(?<=item\.data.)\w*/g)[0];
+      const hasNestedFields = this.component.template.match(/(?<=item\.data.)\w*/g) && this.component.template.match(/(?<=item\.data.)\w*/g)[0];
```
travist, line 198: Why perform two regular expressions? I would prefer that you just create a variable of the result of the match, and then just use that so that it is only executed once.

Tkachez: @travist, done

PR #3197, opened 2020-08-27
Feat/datagrid initialize empty

mikekotikov:
Remade the option to not show first row into option to initialize the DataGrid entirely empty. This makes quite less misunderstanding if multiple default values are set.


mikekotikov:
Strange, tests pass ok at my local

src/components/datagrid/DataGrid.js
```diff
@@ -60,9 +62,9 @@ export default class DataGridComponent extends NestedArrayComponent {
   get defaultSchema() {
     return DataGridComponent.schema();
   }
-  // Checks if should init the first row (only when no defaultValue and just one row set)
+
   get initRows() {
-    return this.builderMode || !_.isEqual(this.defaultValue, this.emptyValue) || !this.component.noFirstRow;
```
travist, line 65: This breaks reverse compatibility with forms that already have the old configuration set. Can we change this logic to also check for ```noFirstRow```?

mikekotikov: Done

src/components/datagrid/DataGrid.js
```diff
@@ -60,9 +62,13 @@ export default class DataGridComponent extends NestedArrayComponent {
...
   }
-  // Checks if should init the first row (only when no defaultValue and just one row set)
+
+  get initEmpty() {
+    return this.component.initEmpty || this.component.noFirstRow;
+  }
+
   get initRows() {
-    return this.builderMode || !_.isEqual(this.defaultValue, this.emptyValue) || !this.component.noFirstRow;
+    return this.builderMode || this.path === 'defaultValue' || !this.initEmpty;
```
travist, line 71: Why did you remove the check if the value is empty? This will break the situation where they must now always click "Add Another" if they want to add a new row to a datagrid instead of initializing it with a row open, right?

mikekotikov: I had to remove it because it appeared to be not a reliable check. Previously I only did it to define the datagrid component that is situated on "Data" tab of datagrid settings where default value is set. 
This check was needed for the situation if you checked an option to hide first row and then wanted to set a default value.
The problem was that you could not properly do it with the rows being hidden.
Now it works such way: if `initEmpty` is set, you can still set default values as normal, but when form is used they will be hidden and only shown gradually when you click "Add Another".

travist: So are you saying that this will cause a regression?  When I have a datagrid on my form, it will ALWAYS initialize to no rows and I have to click "Add Another" to add one, unless i have this checkbox checked?  What happens if I do not set the ```noFirstRow``` variable?  I would assume it would always add an initial first row, but this code makes me believe it will still show as empty?  Am I wrong?

mikekotikov: Here is how it works
https://drive.google.com/file/d/1yczigiLccdsc_9z8NcBMjiMHsACL1zg3/view?usp=sharing

mikekotikov: noFirstRow option set in JSON will work the same

PR #3208, opened 2020-09-01
PDF-128: Fix html element excluding from callback

RomanLetsuk:


src/PDFBuilder.js
```diff
@@ -261,14 +261,11 @@ export default class PDFBuilder extends WebformBuilder {
...
     let container = [];
-    let originalComponent = null;
-    eachComponent(this.webform._form.components, (comp, path, components) => {
-      if (comp.id === component.component.id) {
-        container = components;
-        originalComponent = comp;
-        return true;
-      }
-    });
+    const originalComponent = this.webform._form.components.find(comp => comp.id === component.component.id);
```
travist, line 264: This will break when you have fields nested inside of panel components that have been mapped to the PDF forms.

PR #3219, opened 2020-09-03
FJS-1101: adds option allowing to scroll to the top of Wizard's page when user navigates to it

alexandraRamanenka:


src/components/panel/editForm/Panel.edit.display.js
```diff
@@ -123,6 +123,16 @@ export default [
           cancel: true,
           next: true
         },
+      },
+      {
+        weight: 50,
+        label: 'Scroll to the top when open',
```
travist, line 129: I believe we now want this to be the default. Please remove the option, and then just make it the default. Also please write a test for this.

travist (changes requested):

travist:
This will be added to 4.13.x release.

PR #3224, opened 2020-09-04
FJS-1117: fixed an issue where validation errors are shown for invalid default value on form load 

TanyaGashtold:


TanyaGashtold:
all tests passed locally

src/components/_classes/component/Component.js
```diff
@@ -2628,18 +2628,10 @@ export default class Component extends Element {
     this.checkComponentConditions(data, flags, row);
 
     if (flags.noValidate && !flags.validateOnInit) {
-      if (flags.fromSubmission && this.rootPristine && this.pristine && this.error && flags.changed) {
```
travist, line 2631: It seems that this should still be here. From what I can tell, this covers a different use case than an initial validation check.

travist (changes requested):

src/components/_classes/component/Component.js
TanyaGashtold: Yes, you are right, it is better to leave this logic. I changed the PR. All tests passed locally.

PR #3235, opened 2020-09-09
FJS-1178: adding an ability to use click on previous pages using brea…

DmitryNistratov:
…dcrumbs even when they are disabled

### DIFF
--- a/src/components/panel/editForm/Panel.edit.display.js
+++ b/src/components/panel/editForm/Panel.edit.display.js
@@ -96,6 +96,17 @@ export default [
         }
       },
       {
+        input: true,
+        type: 'checkbox',
+        label: 'Allow Previous',
+        key: 'allowPrevious',
+        defaultValue: false,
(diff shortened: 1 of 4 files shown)


travist:
Please write a test. Also, I do not like the variable name ```enabledWhenVisited```.  I would like to suggest ```allowPrevious``` since that is what we call it in the wizard mode. This option should also be available in the editForm's for Panel's, and should only show up if they have the "clickable" option for breadcrumbs unchecked, since they may only want to allow previous navigation.

src/components/panel/editForm/Panel.edit.display.js
```diff
@@ -95,6 +95,13 @@ export default [
           json: { '!==': [{ var: 'data.breadcrumb' }, 'none'] }
         }
       },
+      {
+        input: true,
```
travist, line 99: This needs a tooltip. Also, it should be conditional and only show up if they have breadcrumbClickable set to false.

travist (changes requested):

PR #3237, opened 2020-09-10
FJS-1103: When you hit the enter button, then you see the field not saving the year correctly.

AlehKatsiubaFormIO:


travist:
Needs tests.

elibus:
Please check #3277 
I think blocking the event propagation on enter won't fix the case when the blur event is triggered after clicking on another component.

travist:
I agree with @elibus. We also need to handle this case as well. @AlehKatsiubaFormIO please handle this case as well.

AlehKatsiubaFormIO:
@travist @elibus 
Case when the blur event is triggered after clicking on another component already have been fixed [here](https://github.com/formio/formio.js/pull/3227/files).

elibus:
@AlehKatsiubaFormIO #3227 introduced a new bug reported in #3277 which is unfortunately closed but the bug is still there. 
There is a codepen in the ticket, if you open it you will see that when a date is picked and you next click on any other component, the date picked is has year 3 years ahead.

travist (changes requested):
Need to better handle blur event handling.

PR #3239, opened 2020-09-11
FJS-1054: disables Subbmit button while uploading files

alexandraRamanenka:


src/templates/bootstrap/tagpad/form.ejs
```diff
@@ -0,0 +1,30 @@
+<div class="formio-tagpad-container clearfix">
```
travist, line 1: Why were these templates added?

travist (changes requested):

PR #3242, opened 2020-09-11
Fixed an issue where DateTime is rendered incorrectly & ignoring Format on mobile devices

alexandraRamanenka:
Don't know how to test since it was a very specific issue which appeared randomly.  I tested manually using my phone to make sure that it will work as expected. 
The problem is that Flatpickr mistakenly recognizes browser as a mobile and performs some additional stuff which renders additional elements and allows to use native mobile OS date picker. That options also makes the component ignore Format option on mobile devices. 
After disabling that option, it still works on mobile devices, but using the same calendar picker as in the desktop browser and with correct format.


travist:
This will be included in 4.13.x build.

travist:
merged manually.

PR #3244, opened 2020-09-11
Conditional to disable helplinks in builder

edwinanciani:


randallknutson (changes requested):
options.helplinks should be these possible options. 
1. url (default to https://help.form.io) - prepend all the help docs with this url so the help can be whitelabeled as well.
2. false - hide the help links altogether.

The value should default to 'https://help.form.io' for the help url.

You may need to change the links on each component to be relative and then prepend the help url.

PR #3250, opened 2020-09-14
FJS-1206: fixed an issue where conditional wizard pages are not shown/hidden if…

TanyaGashtold:
… wizard contains nested form


TanyaGashtold:
all tests passed locally

PR #3254, opened 2020-09-15
FJS-1116: add AddWizardPage component for premium

DmitryNistratov:


src/Wizard.js
```diff
@@ -116,6 +118,27 @@ export default class Wizard extends Webform {
       }
     });
 
+    this.on('renderRows', (newPage) => {
```
travist, line 121: This code "feels" like it will break something. Please join dev support so we can talk through what you are doing here.

travist (changes requested):

travist:
This will go in a separate module.

PR #3255, opened 2020-09-15
PDF-156: Update empty value validating for form component

RomanLetsuk:


RomanLetsuk:
All tests passed locally

PR #3259, opened 2020-09-15
Enable Setting Headers To S3

SalTor:
This PR introduces a pattern for setting (and potentially overriding) header information during the S3 uploadFile step.

This option can be set after form creation. Example:

```javascript
Formio.createForm(/* ... params */).then(form => {
  form.fileService.options.headers = {
    'Content-Disposition': 'attachment',
  }
})
```

This particular PR only introduces one header that would be valid, but theoretically this could be adapted to work for other existing headers.

This particular header has been relevant to a feature I have been working on in my own project and I replicated these changes by copying the s3.js file into my project and using it via `Formio.use({ providers })`

_P.S. Please ignore the myriad of linting commits, I wanted this diff to be as small as possible and for unknown reasons my linter in my editor wasn't jiving with the syntax of the original file._


src/providers/storage/s3.js
```diff
@@ -5,9 +5,23 @@ const s3 = (formio) => ({
     return XHR.upload(formio, 's3', (xhr, response) => {
       response.data.fileName = fileName;
       response.data.key = XHR.path([response.data.key, dir, fileName]);
+      const setXhrHeaders = (xhrToSet) => {
```
travist, line 8: I actually see this being valuable for all XHR file upload requests and not just S3.  Can we move this into the ```XHR.upload``` method?

travist (changes requested):

src/providers/storage/s3.js
SalTor: Absolutely!

SalTor:
@travist I've moved these changes into xhr.js as you suggested.

The implementation isn't what I expected when I initially set out to do this, namely I expected to be able to set these headers after initialization but it turns out you can only set headers after calling `xhr.open()` and so that is the reason for the implementation.

Further, dropbox was an exception to the implementation since it doens't use the `XHR.upload` method like the `s3` and `azure` storage modules do. Instead, it just implements an explicit initialization of a `new XMLHttpRequest()`

I'm unaware of whether or not this last part is an intentional design decision, maybe something specific to dropbox, but I think it could benefit from sharing the implementation that s3 and azure currently re-use.

travist:
Looks good to me!

PR #3260, opened 2020-09-16
VPAT-604 Added options argument to setAlert for customization in VPAT

antonSoftensity:


antonSoftensity:
[VPAT-604 | Tooltip | All browsers | When page is zoomed in tooltip isn’t visible as expected and style of error’s tooltip is different than component’s tooltip](https://trello.com/c/rVJRVVFR/604-vpat-604-tooltip-all-browsers-when-page-is-zoomed-in-tooltip-isnt-visible-as-expected-and-style-of-errors-tooltip-is-different-t)

PR #3286, opened 2020-09-21
FJS-1230 & FJS-1231: fixed an issue where conditional wizard pages and their values are no…

TanyaGashtold:
…t displayed after setting submission


TanyaGashtold:
all tests passed locally

PR #3302, opened 2020-09-25
fixed an issue where it is not possible to hide label of some compone…

TanyaGashtold:
…nts inside data grid


TanyaGashtold:
This is customer ticket. All tests passed locally.

PR #3306, opened 2020-09-29
FJS-1116: add DynamicWizard component for premium

DmitryNistratov:


travist (changes requested):
This code should be in the Premium library where you override the Wizard class to introduce this additional and necessary logic. I do not believe this should be in the open source core.

Issue #3108, opened 2020-08-03
[BUG] Dynamic Select Filtering - Refresh On does not work


kzhang-dsg:
## Environment
- Hosting type
  - [] Form.io
  - [x] Local deployment
    - Version: 
- Formio.js version: 4.10.x and above 
- Frontend framework: none
- Browser: Chrome
- Browser version: 84.0.4147.105

## Steps to Reproduce
1. Follow the documentation to setup Dynamic Select Filtering: https://help.form.io/tutorials/walkthroughs/dynamic-select-filtering/
2. Set the "Refresh On" to "Make" for the Model field.
3. Test the Dynamic Select Filtering

## Expected behavior
The Model options should refresh when the make field is changed.

## Observed behavior
The Model options do not refresh when the make field is changed for formio.js version: 4.10.x and above.

## Additional Notes:
1. When downgrading the formio.js version to 4.9.26, the refresh on function starts working,
2. For formio.js version: 4.10.x and above, if you set  "Refresh On" to "All", it will also work. But this is not really a workaround, because then the "Clear value on refresh" will clear the selection on any field changes.

wag110894:
@kzhang-dsg have you seen this example: https://formio.github.io/formio.js/app/examples/external.html?

kzhang-dsg:
@wag110894

Yes I have. 

Forgot to mention. the bug exists only with HTML5 widget. The Choose.js widget appears to work fine.

We need to use HTML5 widget for native mobile selection UI. The Choose.JS does not work very well out of the box on Mobile.

wag110894:
Could you provide what Mobile Device you are using so we can test against the same device? can you also provide a video of the issue taking place?

kzhang-dsg:
For local development, we just use Chrome. Mobile is just for future support. 

It is easy to reproduce. I just copied the example: https://formio.github.io/formio.js/app/examples/external.html and changed the widget to HTML5, you will see the issue: https://jsfiddle.net/uove48mc/

BTW,  the Choose.js widget also does not fully work. In summary:

1. For Choose.js, The model options refresh on change of the Make field. But it does NOT clear the value on change of the Make field.
3.For HTML5, The model options does NOT refresh on change of the Make field. It also does NOT clear the value on change of the Make field.

wag110894:
[...]

PR #3307, opened 2020-09-29
FJS-1144 Dynamic Select Filtering - Refresh On does not work

kzhang-dsg:


travist:
This makes sense to me, but I do believe that this may cause regressions in other places of the renderer. I also believe that this is the same thing as just removing the ```changePath``` variable and check all together. I would prefer to save this for the ```4.13.x``` release since it may require more extensive testing.

kzhang-dsg:
Hi @travist, I agree. I kept the `changePath` variable for regression so that if there are any place that depends on the `changePath` will still work (but it should not be any).

travist:
The issue I see happening is if you have two components in a form with the same key (like one in a DataGrid and the other at the root), this will trigger erroneous refreshes on the other fields. This may require some more thought. What is the bug you are trying to fix here? Maybe we can fix this another way?

kzhang-dsg:
Hi @travist, the issue I was trying to fix is: FJS-1144 a.k.a. https://github.com/formio/formio.js/issues/3108

travist:
I believe we fixed this in master branch. Please confirm.

PR #3323, opened 2020-10-01
FJS-1245: fixed an issue where dataGrid with initEmpty option hides rows when v…

TanyaGashtold:
…iewing a submission


TanyaGashtold:
all tests passed locally

PR #3329, opened 2020-10-05
FJS-1115: fix ordering for nested wizards with conditional logic

DmitryNistratov:


barclayd (approved):

src/Wizard.js
```diff
@@ -418,6 +400,30 @@ export default class Wizard extends Webform {
     this.allPages = allComponents;
   }
 
+  getSorteredComponents({ components, originalComponents }) { // sorts components if they were shuffled after the conditional logic
```
jamiebuckley, line 403: Minor point, should this be "getSorted" rather than "getSortered"?

DmitryNistratov: Yes, certainly. Thank you

jamiebuckley (approved):

PR #3330, opened 2020-10-05
Fix/uip 247 localization issues

mikekotikov:
@travist , please see if such implementation with `<translate>` tag is ok. For now I've only used it for the places that were complaints about. If the functionality is good to go can spend time to apply it to the rest of static templates.


mikekotikov:
Tests passed ok to me, after the last commit

travist (changes requested):
There has to be a better way to do this instead of introducing a ```<translate>``` html tag. This is very disruptive and will probably break legacy forms, etc. We will not be able to accept this as is.

PR #3331, opened 2020-10-05
FJS-1099: adding overrideBreadcrumbClickableOption to fix logic of cl…

DmitryNistratov:
…ickability of breadcrumbs

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -105,6 +105,11 @@ export default class Wizard extends Webform {
       showCancel: !this.options.readOnly
     });
 
+    if (!this.isSecondInit) {
+      this.isClickableDefined = this.options?.breadcrumbSettings?.hasOwnProperty('clickable');
+      this.isSecondInit = true;
+    }
+
     this.options.breadcrumbSettings = _.defaults(this.options.breadcrumbSettings, {
       clickable: true
     });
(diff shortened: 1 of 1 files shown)


src/Wizard.js
```diff
@@ -18,6 +18,7 @@ export default class Wizard extends Webform {
    *    - breadcrumbSettings.clickable: true (default) determines if the breadcrumb bar is clickable or not
    *    - buttonSettings.show*(Previous, Next, Cancel): true (default) determines if the button is shown or not
    *    - allowPrevious: false (default) determines if the breadcrumb bar is clickable or not for visited tabs
+   *    - overrideBreadcrumbClickableOption: false (default) determines if breadcrumbClickable overrides breadcrumbSettings.clickable
```
travist, line 21: Can we call this ```overrideBreadcrumbClickable```?  

Also, you need to add a default in the ```schema```. Otherwise every new panel component will include this configuration.

travist (changes requested):

travist:
I don't like the idea that we have a new configuration whose sole purpose it is to override another configuration. This seems really strange to me. Why do we need to do it this way?

src/Wizard.js
```diff
@@ -105,6 +105,11 @@ export default class Wizard extends Webform {
       showCancel: !this.options.readOnly
     });
 
+    if (!this.isSecondInit) {
+      this.isClickableDefined = _.clone(this.options?.breadcrumbSettings?.hasOwnProperty('clickable'));
```
travist, line 109: Why do you need to _.clone this?...  It is set to a boolean so therefore is not a reference.

travist (changes requested):

PR #3355, opened 2020-10-09
FJS-1293: fixed signature infinite loop in html render mode

TanyaGashtold:


TanyaGashtold:
all tests passed locally

PR #3362, opened 2020-10-12
Fix typo in class name in align.ejs.

azizk:
I think this must be a typo. The class "filed-content" is used nowhere.


travist:
I think you are right! Thanks!  We will pull this into the 4.13.x release.

PR #3363, opened 2020-10-12
WebFormBuilder Search Enhancements

besmiralia:
* Added debounce for search
* When searching, the components are grouped in a grouped named `searchFields`, so this group was not being searched


travist:
Looks great!  We will bring this into the 4.13.x branch after we stabilize 4.12.x

PR #3364, opened 2020-10-13
Nested: fix untouched panel collapse and then open triggered validation

mikekotikov:
Not for a ticket, just fixing a bug I did not notice previously.


mikekotikov:
@travist tests pass ok at my local.

PR #3366, opened 2020-10-14
FJS-1330: fixed an issue where date value is not set or incorrectly s…

TanyaGashtold:
…et in dateTime/textArea component after openning collapsed panel

### DIFF
--- a/src/components/_classes/input/Input.js
+++ b/src/components/_classes/input/Input.js
@@ -221,10 +221,6 @@ export default class Input extends Multivalue {
     return null;
   }
 
-  getValueAsString(value, options) {
-    return super.getValueAsString(value, options);
-  }
-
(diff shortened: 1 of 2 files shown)


src/components/_classes/input/Input.js
```diff
@@ -222,7 +222,7 @@ export default class Input extends Multivalue {
   }
 
   getValueAsString(value, options) {
-    return super.getValueAsString(this.getWidgetValueAsString(value, options), options);
+    return super.getValueAsString(value, options);
```
travist, line 225: Isn't this the same thing as just removing the method all together?

TanyaGashtold: Yes) You are right. Just deleted it.

PR #3372, opened 2020-10-15
Fixed an issue where conditions do not work for components inside EditGrid

alexandraRamanenka:
No card, found by myself

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -283,9 +283,10 @@ export function checkCondition(component, row, data, form, instance) {
     return checkCustomConditional(component, customConditional, row, data, form, 'show', true, instance);
   }
   else if (conditional && conditional.when) {
-    if (instance?.parent && conditional.when.startsWith(instance.parent.path)) {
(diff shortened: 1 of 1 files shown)


src/utils/utils.js
```diff
@@ -278,14 +278,20 @@ export function checkJsonConditional(component, json, row, data, form, onError)
...
-    return checkCustomConditional(component, component.customConditional, row, data, form, 'show', true, instance);
-  }
-  else if (component.conditional && component.conditional.when) {
-    return checkSimpleConditional(component, component.conditional, row, data);
+  const { customConditional, conditional } = component;
+  if (customConditional) {
+    return checkCustomConditional(component, customConditional, row, data, form, 'show', true, instance);
+  }
+  else if (conditional && conditional.when) {
+    if (instance?.parent && conditional.when.startsWith(instance.parent.path)) {
```
travist, line 286: Would this break if the parent was a panel called "Child" and there was a field inside of it called "Child First Name"?

PR #3396, opened 2020-10-20
Asking for components removal (React FormIO #334)

mm-dsibinski:
Fixing FormIO React [item #334](https://github.com/formio/react-formio/issues/334)

Added 'askForComponentsRemoval' parameter to FormBuilder options. When set, there's always a question asked before removing any component from the FormBuilder.
Otherwise, it behaves like previously, so only asking when removing "grouping" containers (columns etc.).

### DIFF
--- a/src/WebformBuilder.js
+++ b/src/WebformBuilder.js
@@ -959,15 +959,16 @@ export default class WebformBuilder extends Component {
       return;
     }
     let remove = true;
-    if (
-      !component.skipRemoveConfirm &&
+    const removingComponentsGroup = !component.skipRemoveConfirm &&
       (
         (Array.isArray(component.components) && component.components.length) ||
         (Array.isArray(component.rows) && component.rows.length) ||
         (Array.isArray(component.columns) && component.columns.length)
-      )
-    ) {
-      const message = 'Removing this component will also remove all of its children. Are you sure you want to do this?';
+      );
+
+    if (this.options.askForComponentsRemoval || removingComponentsGroup) {
+      const message = removingComponentsGroup ? 'Removing this component will also remove all of its children. Are you sure you want to do this?'
+        : 'Are you sure you want to remove this component?';
       remove = window.confirm(this.t(message));
     }
     if (!original) {
--- a/src/WebformBuilder.js
+++ b/src/WebformBuilder.js
@@ -966,7 +966,7 @@ export default class WebformBuilder extends Component {
         (Array.isArray(component.columns) && component.columns.length)
       );
 
-    if (this.options.askForComponentsRemoval || removingComponentsGroup) {
+    if (this.options.alwaysConfirmComponentRemoval || removingComponentsGroup) {
       const message = removingComponentsGroup ? 'Removing this component will also remove all of its children. Are you sure you want to do this?'
         : 'Are you sure you want to remove this component?';
       remove = window.confirm(this.t(message));


mm-dsibinski:
Hi guys, is something wrong with the PR or it just takes that long to process? Maybe there's something wrong with the PR? I'd be glad to know. @randallknutson ?

travist:
Pull request looks good. We just are working on getting the next 4.12.x release out before moving onto the 4.13.x items. We have many changes staged and we have to be responsible when taking new changes so that they can go into the QA process and get approved. My hope is to have 4.12.1 released either this week or next, and I will pull in all outstanding PR's at that time.

mm-dsibinski:
Great, that's clear, thanks for answering @travist and good luck with the release 😉

src/WebformBuilder.js
```diff
@@ -959,15 +959,16 @@ export default class WebformBuilder extends Component {
...
       (
         (Array.isArray(component.components) && component.components.length) ||
         (Array.isArray(component.rows) && component.rows.length) ||
         (Array.isArray(component.columns) && component.columns.length)
-      )
-    ) {
-      const message = 'Removing this component will also remove all of its children. Are you sure you want to do this?';
+      );
+
+    if (this.options.askForComponentsRemoval || removingComponentsGroup) {
```
travist, line 969: This seems like the incorrect logic...  

If ```askForComponentsRemoval``` is true, it will always ask for confirmation, and if it is false, then it will work like it always has where it will only ask if the skipRemoveConfirm is set to false. Seems like inverted from what you want.  

I think the following is what you are after...

```
if (!this.options.skipRemoveConfirm && removingComponentsGroup) {
```

This way it is reverse compatible, and will only skip if you provide a ```skipRemoveConfirm``` within the options. Your thoughts?

travist (changes requested):
Needs work.

src/WebformBuilder.js
mm-dsibinski: I think what you proposed in otherwise of what I need 😉
We want a possibility to **always ask before removing a component**. In the solution you proposed, the only change would be that I can pass `skipRemoveConfirm` in options which - when set - would make it NOT asking for the component's removal. The goal is otherwise - to ALWAYS ask for components removal, independently whether it's a group of components or a single component.

 The previous behavior was that we only ask for removal if the components group is removed and `skipRemoveConfirm` is not defined or false in the component's options (`component.skipRemoveConfirm`).  wanted to keep it backwards-compatible, so I introduced a new option called `this.options.askForComponentsRemoval`. When this is set to true, it will always ask before removing any component. If this is not set (the default, current behavior) it works as previously (now).

That's why you have this line:
`if (this.options.askForComponentsRemoval || removingComponentsGroup) {`
because it will ask for components removal in one of two cases:

- `removingComponentsGroup` is true, which represents the old (current) conditions

- `this.options.askForComponentsRemoval` which is a newly added option that can be provided to the builder

I guess it clarifies, but let me know if we're still not on the same page 🙂

mm-dsibinski: @travist are we on the same page here?

travist: I will need to run this locally to verify that it is reverse compatible. At first glance, this looks like it would change behavior on forms that were built without this change. If that is the case, then we will not be able to take this "as is".

mm-dsibinski: I don't think it will affect the forms build without this change. As far as I tested, the code without a new option should work as it used to.
Basically the previous condition (`!component.skipRemoveConfirm && (... arrays checking code..)` ) has been placed within `removingComponentsGroup` variable. If this variable is set, the dialog is displayed with the same message as previously. If it's not true, the dialog is skipped. Just as it was before.
But of course run your tests and let me know.

travist: yes you are right. Looks good.

travist: Maybe you really meant to make the variable say... ```alwaysConfirmRemoval```

mm-dsibinski: Ah, now I see what you mean. My intention adding this feature was to add one, new possibility - to allow **asking for every component's removal**. I assumed that we should **always ask for grouping components' removal** (columns, fieldset etc.), because its been implemented like that.
With these assumptions, **if you do not set `askForComponentsRemoval` at all or set it to `false` you'd get the old (current) behavior**.
And exactly, maybe just the naming is an issue here. I see two options:

1. Rename variable `askForComponentsRemoval` to `alwaysConfirmRemoval` or maybe `askForAllComponentsRemoval`
2. Implement the other behavior you assumed could be possible, i.e. not asking for ANY components removal (neither single components nor grouping ones) when `askForComponentsRemoval` is `false`.

To be honest, from the functional point of view, I think we should keep to always ask before removing a grouping component. I wouldn't let disabling it with an option. If the user has many components grouped in a grouping component it would be very frustrating to accidentally delete it all. What do you think?

travist: I am alright changing the variable name so it is more descriptive of what is going on. We could be overly verbose and make it ```alwaysConfirmComponentRemoval```

mm-dsibinski: Done and committed 😉

PR #3406, opened 2020-10-21
FJS-1301 & FJS-1302 Regressions on examples of github (Problem with Formio global object)

yavorovsky:
…condition for embeded event

### DIFF
--- a/_layouts/default.md
+++ b/_layouts/default.md
@@ -67,9 +67,10 @@
 <hr />
 <div>
   <h4 class="text-center text-muted">powered by <img src="{{ site.baseurl }}/app/logo.png" class="mr-2" style="height: 1.2em;" /></h4>
-  <p class="text-center text-muted mb-0" style="font-size: 0.8em">Copyright © Form.io LLC 2019. All rights reserved</p>
+  <p class="text-center text-muted mb-0" style="font-size: 0.8em">Copyright © Form.io LLC 2020. All rights reserved</p>
 </div>
 <p class="text-center text-muted mb-0" style="font-size: 0.8em">Renderer v<span id="renderer-version"></span></p>
+<script src="{{ site.baseurl }}/dist/formio.full.js"></script>
 <script type="text/javascript">
   document.getElementById('renderer-version').innerHTML = Formio.version;
 </script>
--- a/_layouts/default.md
+++ b/_layouts/default.md
@@ -70,7 +70,9 @@
   <p class="text-center text-muted mb-0" style="font-size: 0.8em">Copyright © Form.io LLC 2020. All rights reserved</p>
 </div>
 <p class="text-center text-muted mb-0" style="font-size: 0.8em">Renderer v<span id="renderer-version"></span></p>
+{% if Formio == undefined %}
 <script src="{{ site.baseurl }}/dist/formio.full.js"></script>
+{% endif %}
 <script type="text/javascript">
   document.getElementById('renderer-version').innerHTML = Formio.version;
 </script>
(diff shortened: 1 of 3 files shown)


_layouts/default.md
```diff
@@ -67,9 +67,10 @@
 <hr />
 <div>
   <h4 class="text-center text-muted">powered by <img src="{{ site.baseurl }}/app/logo.png" class="mr-2" style="height: 1.2em;" /></h4>
-  <p class="text-center text-muted mb-0" style="font-size: 0.8em">Copyright © Form.io LLC 2019. All rights reserved</p>
+  <p class="text-center text-muted mb-0" style="font-size: 0.8em">Copyright © Form.io LLC 2020. All rights reserved</p>
 </div>
 <p class="text-center text-muted mb-0" style="font-size: 0.8em">Renderer v<span id="renderer-version"></span></p>
+<script src="{{ site.baseurl }}/dist/formio.full.js"></script>
```
travist, line 73: This is already included on line 32.  Why are we doing it again here?

src/formio.embed.js
```diff
@@ -11,6 +10,8 @@ while (i--) {
 
 if (thisScript) {
   const Form = require('./formio.form').Form;
```
travist, line 12: This could also be written as 

```
const { Form, Formio } = require('./formio.form');
```

travist (changes requested):

_layouts/default.md
yavorovsky: If we want to have an access to Formio on DOM we should use scripts on the body. On the line 32 scripts includes on like a setting on the head. But we are using Formio.version also and when we're trying to include the embed file it's breaking global variable Formio because of an unknown reason. I couldn't find out what is the reason but I know setting scripts for DOM is better to include it on the body.

src/formio.embed.js
yavorovsky: Yes, sure

_layouts/default.md
travist: This is probably because the "embed" overrides the "instance" that is already defined. I think the solution here is to add an if statement to the pages so that they will not include the formio.form.js at the beginning and the "embed" page would set this flag since it includes it in the embed.

travist (changes requested):

travist:
Not sure if your check for Formio will work. What I was saying is something like this... You would add the following to the top of the embed.md page.

```
---
title: JavaScript Embedding
layout: vtabs
section: examples
noFormio: true
weight: 30
---
```

Then, you can add the following in this script ...

```
{% if page.noFormio == nil %}
<script src="{{ site.baseurl }}/dist/formio.full.js"></script>
{% endif %}
```

travist (changes requested):
Please resolve latest comment.

yavorovsky:
And I suppose just {% if page.noFormio %} in this case 
And yes, you're right it looks better, thank you

_layouts/default.md
```diff
@@ -67,9 +67,12 @@
 <hr />
 <div>
   <h4 class="text-center text-muted">powered by <img src="{{ site.baseurl }}/app/logo.png" class="mr-2" style="height: 1.2em;" /></h4>
-  <p class="text-center text-muted mb-0" style="font-size: 0.8em">Copyright © Form.io LLC 2019. All rights reserved</p>
+  <p class="text-center text-muted mb-0" style="font-size: 0.8em">Copyright © Form.io LLC 2020. All rights reserved</p>
 </div>
 <p class="text-center text-muted mb-0" style="font-size: 0.8em">Renderer v<span id="renderer-version"></span></p>
+{% if page.noFormio %}
```
travist, line 73: But what this says is you only are going to include the renderer if they have "noFormio" provided in the page... this means the only page that will include this is the inline embed page.

travist (changes requested):

_layouts/default.md
travist: I will fix this myself.  We need to get a build going.

PR #3409, opened 2020-10-22
PDF-204: Remove setting primary project id to download pdf

RomanLetsuk:


travist:
I may prefer to make the change on the server instead where this parameter does not do anything. The reason is because this will now be a huge problem for those who upgrade the renderer but do not upgrade the PDF Server. We would like to make it so that if they upgrade the PDF server it fixes the "project" resolution on the server side and just ignores the project property from the renderer.

RomanLetsuk:
> I may prefer to make the change on the server instead where this parameter does not do anything. The reason is because this will now be a huge problem for those who upgrade the renderer but do not upgrade the PDF Server. We would like to make it so that if they upgrade the PDF server it fixes the "project" resolution on the server side and just ignores the project property from the renderer.

@travist I think we can close this PR, fixed on PDF Server side. https://github.com/formio/pdf-server/pull/35

PR #3411, opened 2020-10-22
Fixes an issue where values of conditionally visible components inside EditGrid are not saved on the 6.x server

alexandraRamanenka:
No Trello card, found by Dmitry

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -1215,7 +1215,7 @@ export function getDataParentComponent(componentInstance) {
     return;
   }
   const { parent } = componentInstance;
-  if (parent && parent.isInputComponent) {
+  if (parent && (parent.isInputComponent || parent.input)) {
     return parent;
   }
   else {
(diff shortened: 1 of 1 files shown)


src/utils/utils.js
```diff
@@ -1215,7 +1215,7 @@ export function getDataParentComponent(componentInstance) {
     return;
   }
   const { parent } = componentInstance;
-  if (parent && parent.isInputComponent) {
```
travist, line 1218: I do not see "input" as a property defined on the Component instance. Did you mean to say ```parent.component.input```?

travist (changes requested):

src/utils/utils.js
alexandraRamanenka: This is only for 6.x server, it passes component's schema with 'parent' field which is a parent's schema. So, I just make all the calculations correct calculating path of the component and adding it to that schema, this part now is in the PR for formio, but I'm able to move it here. So, I will update that PR tomorrow.

PR #3412, opened 2020-10-22
FOR-2419: Changed the logic of filtering and added searching by existing resource

yavorovsky:


travist:
Saving this for 4.13.x release.

PR #3414, opened 2020-10-22
Allow 0 for labelWidth and labelMargin.

azizk:
Hi!

I think labelWidth and labelMargin should allow 0 as a value. At least labelMargin should work with 0.
~~My changes also constrain the properties to a number because a string compared to a number ("..." >= 0) will always be false.~~
Edit: actually never mind, because of implicit conversion "2" >= 0 gives true.

What do you think? :)


travist:
I'm alright with this, but it will need to wait for the 4.13.x release.

PR #3416, opened 2020-10-25
Style button to delete data in confirmation dialog as btn-danger

elibus:
Clicking on the button "Yes, delete it" will drop all changes not persisted. I think this should be a btn-danger rather than a btn-primary.
For your consideration.


travist:
We will consider this for 4.13.x

Issue #3392, opened 2020-10-19
[BUG] Formio.js sets Ace options incorrectly when ace is imported outside Formio.js


Sefriol:
## Environment
Please provide as many details as you can:

- Hosting type
  - [ ] Form.io
  - [x] Local deployment
    - Version: 
- Formio.js version: 4.1.12
- Frontend framework: VueJS
- Browser: Chrome
- Browser version: Version 86.0.4240.75

## Steps to Reproduce
1.  Install Ace and Formio.js within webpack environment
2. Open FormEditor. Try to edit a button.
3. Error will appear where it tries to use HTML file as a mode file.

## Expected behavior
It should use `ace-builds/webpack-resolver`'s file path
## Observed behavior
It uses HTML file

## Conclusion and how to solve
I do not go deeply into what is causing the problem, or how to trigger it. I will go through your code and explain why this happens:

https://github.com/formio/formio.js/blob/38fc9c184f0785285e0a2abd86f5d143990fc810/src/components/_classes/component/Component.js#L2066
Here you try to set a settings object into the editor. Unfortunately this settings object also includes mode, which is in format `html`, `json`, `javascript` or so on.
This will work if you have a non-webpack environment. However in other cases, mode path needs to be defined similarly as in
https://github.com/formio/formio.js/blob/38fc9c184f0785285e0a2abd86f5d143990fc810/src/components/_classes/component/Component.js#L2071
For example `ace/mode/html`. Otherwise it will raise an error on line 2069, because it tries to use `html` as key for the mode instead of `ace/mode/html`. 

i.e It could be fixed something like this (Fast MVP):
```javascript
return Formio.requireLibrary('ace', 'ace', _.get(this.options, 'editors.ace.src', ACE_URL), true)
      .then((editor) => {
        editor = editor.edit(element);
        editor.removeAllListeners('change');
        const settingsObj = Object.assign({}, settings);
        settingsObj.mode = `ace/mode/${settings.mode}`
        editor.setOptions(settingsObj);
        editor.getSession().setMode(settingsObj.mode); //Most likely not needed
        editor.on('change', () => onChange(editor.getValue()));
        return editor;
      });
```

daneformio:
Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #3417, opened 2020-10-26
Fix Ace in webpack environments

Sefriol:
In relation to https://github.com/formio/formio.js/issues/3392

Not sure if this causes problems but opened this to jump start the discussion.

### DIFF
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -2068,7 +2068,7 @@ export default class Component extends Element {
         editor = editor.edit(element);
         editor.removeAllListeners('change');
         editor.setOptions(settings);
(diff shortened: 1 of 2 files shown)


Sefriol:
So far this seems to fix this in our environment.

Sefriol:
i.e. the issue is that in webpack environment, if you have `ace` already defined outside of `formio.js`, it will raise an error when it tries to load the mode from a file that does not exist. `settings` has an attribute `mode` which is written as `json`, `html`, `js` etc. This is fine in non-webpack environments, however in webpack environments this will raise an error. It should be written as `ace/mode/json`,  `ace/mode/html`,  `ace/mode/js` etc.

travist:
Seems perfectly reasonable.

PR #3418, opened 2020-10-26
FJS-1349: fix EditGrid reset after cancel

DmitryNistratov:


src/components/editgrid/EditGrid.js
```diff
@@ -941,6 +941,13 @@ export default class EditGridComponent extends NestedArrayComponent {
       this.setNestedValue(component, editRow.data, flags);
     });
   }
+
+  setPristine(pristine) {
+    super.setPristine(pristine);
+    if (pristine) {
+      this.editRows = [];
```
travist, line 948: This may cause unintended consequences. You are making a bad assumption that the only time that pristine is set is when the EditGrid is canceled. I can think of many cases when pristine is set to true when this is not the case.

travist (changes requested):

travist (changes requested):

src/components/editgrid/EditGrid.js
```diff
@@ -941,6 +941,11 @@ export default class EditGridComponent extends NestedArrayComponent {
       this.setNestedValue(component, editRow.data, flags);
     });
   }
+
+  resetValue() {
+    super.resetValue();
+    this.editRows = [];
```
travist, line 947: Now I think this will cause memory leaks. We should be destroying all components before setting this to empty. I think we need to change it to this.

```js
emptyRows() {
  this.editRows.forEach((editRow) => {
    editRow.components.forEach((comp) => {
      comp.destroy();
    });
  });
  this.editRows = [];
}
```

Then you will just call ```this.emptyRows();``` here.

travist (changes requested):

src/components/editgrid/EditGrid.js
travist: Actually it may need to be...

```js
emptyRows() {
  this.editRows.forEach((editRow, index) => this.destroyComponents(index));
  this.editRows = [];
}

```

PR #3419, opened 2020-10-26
FJS-1348: remove the DynamicWizard component from the Webform builder

DmitryNistratov:


src/WebformBuilder.js
```diff
@@ -123,7 +123,7 @@ export default class WebformBuilder extends Component {
       if (this.groups[group] && this.groups[group].components) {
         this.groups[group].componentOrder = Object.keys(this.groups[group].components)
           .map(key => this.groups[group].components[key])
-          .filter(component => component && !component.ignore)
+          .filter(component => component && !component.ignore && component.key !== 'dynamicWizard')
```
travist, line 126: You don't do this here. Just add "ignore" property in the component builderInfo static method.

travist (changes requested):

PR #3424, opened 2020-10-27
FJS-1311: Clicking on the errors not redirected to the corresponding field in the Child Wizard

AlehKatsiubaFormIO:


src/components/panel/Panel.js
```diff
@@ -61,4 +63,11 @@ export default class PanelComponent extends NestedComponent {
       }
     });
   }
+
+  getComponent(path, fn, originalPath) {
+    if (this.parent instanceof Wizard && this.parent.parent instanceof FormComponent) {
```
travist, line 68: What if the Panel is inside of another panel and then inside of the form?... wouldn't this break?  I believe the following would work.

```
if (this.root.parent && this.root.parent instanceof FormComponent) {
  ...
}
```

travist (changes requested):

src/components/panel/Panel.js
AlehKatsiubaFormIO: Yes, that makes sense.

PR #3425, opened 2020-10-27
fix multiple files disabled allowed to browse during upload

mikekotikov:


src/components/file/File.js
```diff
@@ -325,7 +325,9 @@ export default class FileComponent extends Field {
     if (this.refs.fileBrowse) {
       this.addEventListener(this.refs.fileBrowse, 'click', (event) => {
         event.preventDefault();
-
+        if (!this.component.multiple && this.statuses.some(fileUpload => fileUpload.status === 'progress')) {
```
travist, line 328: Will this break IE11?  I am not sure if ```.some``` will be converted but please verify.

travist (changes requested):

PR #3431, opened 2020-10-27
FJS-1372: Fixes an issue where Rendered Form and Submission are not updated when form is changed

alexandraRamanenka:


src/Webform.js
```diff
@@ -647,9 +647,8 @@ export default class Webform extends NestedDataComponent {
       // If provided form is not a valid JSON object, do not set it too
       return NativePromise.resolve();
     }
-
     // Create the form.
-    this._form = form;
+    this._form = _.fastCloneDeep(form);
```
travist, line 651: It should be Utils.fastCloneDeep

travist (changes requested):

PR #3432, opened 2020-10-27
FJS-1348: remove the DynamicWizard component from the Webform builder

DmitryNistratov:


DmitryNistratov:
Related to https://github.com/formio/premium/pull/56
@travist I couldn't find a way to determine whether the component is in the Wizard or not inside the builderInfo. I did this way, but maybe there is a better solution

src/WebformBuilder.js
```diff
@@ -123,7 +123,7 @@ export default class WebformBuilder extends Component {
       if (this.groups[group] && this.groups[group].components) {
         this.groups[group].componentOrder = Object.keys(this.groups[group].components)
           .map(key => this.groups[group].components[key])
-          .filter(component => component && !component.ignore)
+          .filter(component => component && !component.ignore && !component.ignoreForForm)
```
travist, line 126: If you are using ```ignoreForForm``` which is the same thing as ```ignore``` why not just set the ignore flag??? This seems unnecessary. I think you need to keep digging.

travist (changes requested):

src/WebformBuilder.js
DmitryNistratov: Not absolutely the same. I need to remove the component from the list for the WebformBuilder, but leave for the WizardBuilder, because in the WebformBuilder it works incorrectly. The formation of the list is inherited from the WebformBuilder.
But if you confirm that I can reserve "ignore" as "ignore only for form", I will of course do it.

PR #3437, opened 2020-10-28
FJS-1355: add disableSiblings logic for the builder

DmitryNistratov:


src/WebformBuilder.js
```diff
@@ -822,6 +822,24 @@ export default class WebformBuilder extends Component {
...
+      let isCompAlreadyExists = false;
+      eachComponent(this.webform.components, (component) => {
+        if (component.key === draggableComponent.key) {
+          isCompAlreadyExists = true;
+          return;
+        }
+      }, true);
+      if (isCompAlreadyExists) {
+        this.webform.redraw();
+        this.webform.setAlert('danger', `You cannot add more than one ${draggableComponent.key} component to one page.`);
```
travist, line 838: It should not allow ANY siblings, so this error is incorrect by providing the "key" of the draggable component.

travist (changes requested):

PR #3439, opened 2020-10-28
FJS-1092: Wizard form page navigation bar needs to be responsive

AlehKatsiubaFormIO:


src/sass/formio.form.scss
```diff
@@ -1229,3 +1229,7 @@ div[read-only="true"] {
     padding-left: 12px;
   }
 }
+
+.pagination {
```
travist, line 1233: Wouldn't this change ALL paginations on the page and not just the header?  I would think that you would get a little more selective with this CSS.

travist (changes requested):

src/sass/formio.form.scss
AlehKatsiubaFormIO: Yep, my fault.

Issue #3442, opened 2020-10-28
[BUG] FormioUtils.eachComponent does not work with nested column components


JonathanFrias:
## Environment
`FormioUtils.eachComponent` does not properly walk through the document. I've attached a minimal example below. I think this will fix some other issues, like https://github.com/formio/formio.js/issues/3293 

I believe this used to work, as I have code depending on this that used to work

- Hosting type
  - [] Form.io
  - [x ] Local deployment N/A
    - Version: latest
- Formio.js version: latest
- Frontend framework: N/A
- Browser: N/A
- Browser version: N/A

## Steps to Reproduce
1. View attached HTML document

## Expected behavior
`FormiotUtils.eachComponent` should walk all the components
## Observed behavior
This function skips an element.

## Example
This happens when you have a columns Component inside of a panel component. I might be able to have a patch to solve this issue, but it appears that the issue is on or around this line.https://github.com/formio/formio.js/blob/master/src/utils/formUtils.js#L100.
It correctly determines that columns component has columns that should be iterated, and it expects a flat array of components to process. However when dealing with a columns component, it is a 2D nested array of components, and will not traverse the data structure properly. I have produced a minimal example below.

```html
<html>
  <head>
    <link rel='stylesheet' href='https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css'>
    <link rel='stylesheet' href='https://unpkg.com/formiojs@latest/dist/formio.full.min.css'>
    <script src='https://unpkg.com/formiojs@latest/dist/formio.full.min.js'></script>
    <script type='text/javascript'>
      window.onload = function() {
          Formio.createForm(document.getElementById('formio'), {
            "components": [
                {
                    "title": "TestPanel",
                    "collapsible": false,
                    "key": "testPanel",
                    "type": "panel",
                    "label": "Panel",
                    "input": false,
                    "tableView": false,
                    "components": [
                        {
                            "label": "TestColumn",
                            "columns": [
                                {
                                    "components": [
                                        {
                                            "label": "Skipped Element",
                                            "tableView": true,
                                            "key": "skippedElement",
                                            "type": "textfield",
                                            "input": true,
                                            "hideOnChildrenHidden": false
                                        }
                                    ],
                                    "width": 6,
                                    "offset": 0,
                                    "push": 0,
                                    "pull": 0,
                                    "size": "md"
                                },
                                {
                                    "components": [],
                                    "width": 6,
                                    "offset": 0,
[...]

PR #3443, opened 2020-10-28
Fix FormUtils.eachComponent

JonathanFrias:
Handle 2D arrays in response to https://github.com/formio/formio.js/issues/3442

This is my first contribution onto this project FYI

### DIFF
--- a/src/utils/formUtils.js
+++ b/src/utils/formUtils.js
@@ -99,7 +99,7 @@ export function eachComponent(components, fn, includeAll, path, parent) {
         component.columns.forEach((column) => {
           const components = Array.isArray(column) ? column : column.components;
           return eachComponent(components, fn, includeAll, subPath(), parent ? component : null);
-        }
+        });
       }
 
       else if (hasRows) {
--- a/src/utils/formUtils.js
+++ b/src/utils/formUtils.js
@@ -96,8 +96,10 @@ export function eachComponent(components, fn, includeAll, path, parent) {
 
     if (!noRecurse) {
       if (hasColumns) {
-        component.columns.forEach((column) =>
-          eachComponent(column.components, fn, includeAll, subPath(), parent ? component : null));
+        component.columns.forEach((column) => {
+          const components = Array.isArray(column) ? column : column.components;
+          return eachComponent(components, fn, includeAll, subPath(), parent ? component : null);
+        }
       }
 
       else if (hasRows) {


src/utils/formUtils.js
```diff
@@ -96,8 +96,10 @@ export function eachComponent(components, fn, includeAll, path, parent) {
 
     if (!noRecurse) {
       if (hasColumns) {
-        component.columns.forEach((column) =>
-          eachComponent(column.components, fn, includeAll, subPath(), parent ? component : null));
+        component.columns.forEach((column) => {
```
JonathanFrias, line 99: I'm not sure, but there's a chance we may even want to do this with the other calls to `eachComponent` in this function. I didn't include that here, as I'm just trying to be conservative.

JonathanFrias:
Hacks I resort to :/
Found a way to monkey patch this. 

```javascript
function monkeyPatchedFunc(components, fn, includeAll, path, parent) {
  if (!components) return;
  path = path || '';
  components.forEach((component) => {
    if (!component) {
      return;
    }
    const hasColumns = component.columns && Array.isArray(component.columns);
    const hasRows = component.rows && Array.isArray(component.rows);
    const hasComps = component.components && Array.isArray(component.components);
    let noRecurse = false;
    const newPath = component.key ? (path ? (`${path}.${component.key}`) : component.key) : '';

    // Keep track of parent references.
    if (parent) {
      // Ensure we don't create infinite JSON structures.
      component.parent = clone(parent);
      delete component.parent.components;
      delete component.parent.componentMap;
      delete component.parent.columns;
      delete component.parent.rows;
    }

    // there's no need to add other layout components here because we expect that those would either have columns, rows or components
    const layoutTypes = ['htmlelement', 'content'];
    const isLayoutComponent = hasColumns || hasRows || hasComps || layoutTypes.indexOf(component.type) > -1;
    if (includeAll || component.tree || !isLayoutComponent) {
      noRecurse = fn(component, newPath, components);
    }

    const subPath = () => {
      if (
        component.key &&
        !['panel', 'table', 'well', 'columns', 'fieldset', 'tabs', 'form'].includes(component.type) &&
        (
          ['datagrid', 'container', 'editgrid', 'address'].includes(component.type) ||
          component.tree
        )
      ) {
        return newPath;
      }
      else if (
        component.key &&
        component.type === 'form'
      ) {
        return `${newPath}.data`;
      }
      return path;
    };

    if (!noRecurse) {
      if (hasColumns) {
        component.columns.forEach((column) => {
          const components = Array.isArray(column) ? column : column.components;
          return monkeyPatchedFunc(components, fn, includeAll, subPath(), parent ? component : null);
        })
      }

      else if (hasRows) {
        component.rows.forEach((row) => {
          if (Array.isArray(row)) {
            row.forEach((column) =>
              monkeyPatchedFunc(column.components, fn, includeAll, subPath(), parent ? component : null));
          }
        });
      }

      else if (hasComps) {
        monkeyPatchedFunc(component.components, fn, includeAll, subPath(), parent ? component : null);
      }
    }
  });
}
```

I was tempted to do a straightforward method replace, but you can't do a straightforward monkey patch

```javascript
    FormioUtils.eachComponent = monkeyPatchedFunc

   // TypeError: Cannot set property eachComponent of #<Object> which has only a getter
```

But this works:

```javascript
    const _FormioUtils = {
      ...FormioUtils,
      eachComponent: monkeyPatchedFunc,
    };

   _FormioUtils.eachComponent(/*..*/)
```

This is hacky though... please help @travist?

travist:
We will stare at this much closer once we roll out the next release. Tagging with 4.13.x.

travist:
I am not sure if we will be able to accept this. This assumes that the columns component schema is nested arrays, which is not the case. Seems like you just need to fix the form schema to be the correct schema and then nested columns will work just fine.

travist:
Nested columns inside panels work just fine. 

Just go to https://formio.github.io/formio.js/app/sandbox and load the following sandbox 5fc5b9d578e15845c1dc987b

valentinmarinro:
Hi @travist and @JonathanFrias, I have the same issue, I'm trying to make the form responsive by using Columns component as much as possible.
My scenario is the following:
Collapsible Panel -> Columns split in 2 columns with widths 2 and 10 -> in the second column(the one with width 10) I added another Columns component that I split in 4 columns with width 3 each

**I'm using v4.12.4**

FormioUtils.getComponent and FormioUtils.searchComponents failed to get the textfield 'salutation' that's inside the second Columns component and this is because FormioUtils.eachComponent fails to find the salutation component.

I've created a sandbox: [601319c98aa7c24a94dd1b00](https://formio.github.io/formio.js/app/sandbox)
Also attached a stripped down schema with above use case [schema.txt](https://github.com/formio/formio.js/files/5889748/schema.txt)

In my scenario the following methods in FormUtils are not working correctly, except form.GetComponent
                
```js
// does not return the component inside the second Columns component
let flattenComponents = FormioUtils.flattenComponents(form.components, true);
console.log("flattenComponents", flattenComponents);

//FAIL to return the salutation component
let component = FormioUtils.getComponent(form.components, 'salutation', true);
console.log("getComponent", component);

//FAIL to find any text field component                
let components = FormioUtils.searchComponents(form.components, {
  'type': 'textfield'
});                                
console.log("searchComponents", components);
    
//FAIL: finds the 'salutation' component twice             
FormioUtils.findComponent(form.components, 'salutation', null, function (component, newPath, components){
    let id = component.id;   
    console.log("findComponent", component);
});
                
//WORKS: the only one that works is form.getComponent;                
let salutation = form.getComponent('salutation');
```

PR #3444, opened 2020-10-29
Fixed adress component label in editgrid header

AlexeyDavyd:


travist:
Changing the default header template in EditGrid makes me very nervous. Can you explain why this was needed to solve an issue with Addresses?

AlexeyDavyd:
> Changing the default header template in EditGrid makes me very nervous. Can you explain why this was needed to solve an issue with Addresses?

Without flag set to true code have been skipping nested components.

travist:
Right, but we have some customers that have hundreds of nested fields within an edit grid. This would make the default template suddenly show all hundred fields in the template. I don't think we can accept this. Can we just fix this for just Address components within an editgrid instead of affecting ALL components?

travist:
Closing so we can fix this in a different way.

PR #3450, opened 2020-10-30
FJS-1306: Fixes an issue where JSON Select example is not loaded in IE11

alexandraRamanenka:


app/examples/multiform.md
```diff
@@ -182,6 +182,7 @@ workflow.form = {
   <div id="workflow"></div>
   <script type="text/javascript">
   {% raw %}
+  debugger;
```
travist, line 185: Will you remove the debugger statement?

travist (changes requested):

PR #3451, opened 2020-10-30
VPAT-664 Changed setAlert signature to more flexible extension in VPAT

antonSoftensity:


antonSoftensity:
[VPAT-664 | Different error list displays at the top of the form (not the one used in the VPAT)](https://trello.com/c/9MDfI5d5/664-vpat-664-different-error-list-displays-at-the-top-of-the-form-not-the-one-used-in-the-vpat)

src/Webform.js
```diff
@@ -1084,7 +1084,7 @@ export default class Webform extends NestedDataComponent {
     }
     if (message) {
       this.alert = this.ce('div', {
-        class: classes || `alert alert-${type}`,
+        class: (options && options.classes) || `alert alert-${type}`,
```
travist, line 1087: This needs to be reverse compatible... maybe something like.

```js
class: (typeof options === 'string') ? options : ((options && options.classes) || `alert alert-${type}`)
```

travist (changes requested):

src/Webform.js
antonSoftensity: The thing is, 'classes' parameter was introduced to be used in VPAT and recently I changed it to options to accommodate more customizable stuff into one object. I'm not sure if it needs to be reverse compatible, only if there is a scenario when an older version of VPAT is used with newest renderer.

PR #3458, opened 2020-11-04
FJS-1175 [feat]: adds an ability to NestedForm to use original revision for submissions viewing 

alexandraRamanenka:


travist:
This will be part of 4.13.x release.

Issue #3357, opened 2020-10-12
EditGrid row select event


krixerx:
EditGrid has currently 2 classes which are used as row events: editRow and removeRow.

Could we please have one more event "selectRow"?
It does nothing except it would throw the event with currently selected row, so that the row could have selectable (not editable).

With "selectRow" we could build very interesting and useful forms together with field-logic.

PR #3461, opened 2020-11-05
FJS-3357 enable selection of closed EditGrid rows

kamsi:
Hi @travist 

This should solve #3357 . It would be cool if you could review it. The way I implemented is totally OK for our use case (talked to @krixerx ). But maybe you would rather have it handled differently?

### DIFF
--- a/src/components/editgrid/EditGrid.js
+++ b/src/components/editgrid/EditGrid.js
@@ -302,6 +302,9 @@ export default class EditGridComponent extends NestedArrayComponent {
     let openRowCount = 0;
     this.rowElements.forEach((row, rowIndex) => {
       const editRow = this.editRows[rowIndex];
+      if (this.dataValue[rowIndex].isRowSelected) {
+        row.classList.add('selected');
+      }
       if (this.isOpen(editRow)) {
         this.attachComponents(row, editRow.components);
         this.addEventListener(this.saveRowElements[openRowCount], 'click', () => this.saveRow(rowIndex, true));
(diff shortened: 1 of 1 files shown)


travist:
Thanks for the contribution! We will probably include this in 4.13.x. We are releasing 4.12.1 first and will then pull in these pr's.

kamsi:
@travist Turns out I will need to add some more state handling to the PR. We need to know which rows in the EditGrid have been selected between Form reloads. Not sure about conventions on handling such cases. I would propose to add an entry into the row's data like `isRowSelected` - that would give us enough control for our needs. Unsure about your take on using data object like this...

travist:
I would rather just add a hook to this component that would enable you to add checkboxes to the rows external to the renderer itself. If you add such hook using ```this.hook```, then we will be happy to accept that.

kamsi:
@travist maybe I don't see clearly how hooks work, but from that I see, I could of course add a hook like

```
hooks: { renderEditgrid: function () {
   // Do whatever with plenty of arguments
}
```

But in that case I don't need a PR as I am left with overriding the whole rendering in scope of my hook.
On the other hand if I were to go for checkboxes solution, then I don't see how I achieve this with hooks. What I could try to do instead is to add the checkbox into the `rowTemplate`. But then I would be refactoring the default templates and adding some methods to handle clicks there.
So, perhaps, you can reconsider and take the PR as is (it really is quite invisible to users who do not care about this functionality) or guide me a bit more where to go with this (like: modify rowTemplate to add checkbox)?

travist:
So maybe I am not understanding what this is doing... is all this is doing is adding a "click" event handler to when you click on a row of an editgrid, and then just emit an event?  No apparent changes from the end user?  If that is the case, then we can probably accept it. I was thinking you were adding an additional ui element or something.

kamsi:
@travist you got that right! It simply adds one css class and emits that click, so if no one is interested in consuming this, they shall not see anything.

travist:
Ok, then we will take this.  We are a little backed up on our builds so I will pull this in once we are able to start staging the 4.13.x release.

travist:
This is not letting me merge it. Maybe it needs to be rebased, and also make the pull request against our master branch so that it will give me a merge button in the pull request.

kamsi:
@travist is there any ETA for version 4.13.x?

wag110894:
Helo @kamsi,

Our 4.13x renderer is available in an rc version. We are working hard to get a stable build of 4.13x.
https://github.com/formio/formio.js/blob/master/Changelog.md

alexandraRamanenka:
@kamsi this  PR https://github.com/formio/formio.js/pull/4041 will change it  a bit. Instead of saving the selected state of rows in the dataValue (which represents the current user's input to the editGrid), now it will be saved within the editRows property of the editGrid (which represents current rows and their state). It should work the same, but just keep it in mind and recheck your use cases.

PR #3468, opened 2020-11-09
FJS-1426: Fixes an issue where NestedForm's data is not shown when it is inside DataGrid

alexandraRamanenka:


src/components/form/Form.js
```diff
@@ -536,7 +536,7 @@ export default class FormComponent extends Component {
   }
 
   isEmpty(value = this.dataValue) {
-    return value === null || _.isEqual(value, this.emptyValue) || this.areAllComponentsEmpty(value.data);
+    return value === null || _.isEqual(value, this.emptyValue) || (this.areAllComponentsEmpty(value.data) && !value._id);
```
travist, line 539: Does this change also work when "Save as reference" is turned off?

alexandraRamanenka: Of course, I tested it. If it is not a reference and doesn't have an id, but has a non-empty data property, isEmpty returns false. If data is empty, it returns false. For NestedForms saved as references it returns true if only the data property is empty and value doesn't contain an _id.

PR #3481, opened 2020-11-12
[Feat]: Adds ability to disallow caching Select requests 

alexandraRamanenka:
Required for cases when remote data is changing quite often and required to be up-to-date


src/components/select/editForm/Select.edit.data.js
```diff
@@ -648,4 +648,19 @@ export default [
     tooltip: 'A raw JSON object to use as options for the Select component (Choices JS).',
     defaultValue: {},
   },
+  {
+    type: 'checkbox',
+    input: true,
+    weight: 29,
+    key: 'allowCaching',
```
travist, line 655: This is not reverse compatible. We need to make this falsey by default. I propose to call it ```ignoreCache``` and then make the setting called the same. This way you could just set the following and it will be reverse compatible.

```
options.ignoreCache = this.component.ignoreCache;
```

travist (changes requested):

src/components/select/editForm/Select.edit.data.js
alexandraRamanenka: I set it to true because by default we are always using cache for Select components and if  set it to false, all the Selects will 'disallowCaching'.
But your suggestion is much clear

travist: The problem arises with forms that were created before this change and do not have the ```allowCaching``` property set. Those forms will now start ignoring cache after the update which will break reverse compatibility.

travist (changes requested):
Also please rebase.

PR #3485, opened 2020-11-13
VPAT-431 Fixed issue when CKEditor text size styles were missing in f…

antonSoftensity:
…orms (content components)


antonSoftensity:
[VPAT-431 | GSA | Content | All browsers | Text with any font size defined (like Big or Small) is rendered incorrectly in USWDS](https://trello.com/c/6KSgIRXo/431-vpat-431-gsa-content-all-browsers-text-with-any-font-size-defined-like-big-or-small-is-rendered-incorrectly-in-uswds)

PR #3504, opened 2020-11-19
FJS-1380: Adds a validation which checks if the value is available within component's values setting

alexandraRamanenka:


AlexeyDavyd (approved):

MaksimFalei (approved):

src/i18n.js
```diff
@@ -34,6 +34,7 @@ export default {
         invalid_date: '{{field}} is not a valid date.', // eslint-disable-line camelcase
         invalid_day: '{{field}} is not a valid day.', // eslint-disable-line camelcase
         mask: '{{field}} does not match the mask.',
+        valueIsNotAvailable: '{{field}} conatains unavailable value',
```
travist, line 37: Can we change this to say ```{{ field }} is an invalid value.```

travist (changes requested):

MaksimFalei (approved):

PR #3507, opened 2020-11-19
FJS-1433: remove duplicate label in the list of errors

DmitryNistratov:


alexandraRamanenka (approved):

TanyaGashtold (approved):

src/validator/Validator.js
```diff
@@ -601,6 +617,7 @@ class ValidationChecker {
       },
       custom: {
         key: 'validate.custom',
+        hasLabel: true,
```
travist, line 620: We had someone complain about Custom including the labels as well. If they are adding a custom error message, wouldn't they also be able to add the label if they want it there?

travist (changes requested):

src/validator/Validator.js
DmitryNistratov: I didn't take into account the logic of custom validations. If I understood correctly, now everything should work correctly.

PR #3511, opened 2020-11-20
Fix/uip 260 silent cancel submission

mikekotikov:


src/Webform.js
```diff
@@ -1280,7 +1280,7 @@ export default class Webform extends NestedDataComponent {
 
     // Allow for silent cancellations (no error message, no submit button error state)
     if (error && error.silent) {
-      this.emit('change', { isValid: true });
+      this.emit('change', { isValid: true }, { silent: true });
```
travist, line 1283: Shouldn't this say

```
{isValid: true, silent: true}
```

travist (changes requested):

src/Webform.js
mikekotikov: Can do it this way.
The only difference is now it is recieved as `flags.silent`, and then will be `value.silent`.

travist: Oh, yeah, you're right.  thanks.

AlexeyDavyd (approved):

PR #3512, opened 2020-11-20
Select: add use exact search option

mikekotikov:


mikekotikov:
@travist is the tooltip text ok?

PR #3519, opened 2020-11-24
FJS-1470: fixed an issue where wizard with dataGrid inside creates infinite loop

TanyaGashtold:


AlexeyDavyd (approved):

alexandraRamanenka:
This completely breaks editing components inside DataGrid.  Changes are not saved and components cannot be delated

TanyaGashtold:
Retested, it really breaks dataGrid components logic inside builder

PR #3522, opened 2020-11-25
FJS-1470: fixed an issue where infinite loop is created in wizard aft…

TanyaGashtold:
…er internal form schema change

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -733,10 +733,14 @@ export default class Wizard extends Webform {
     if (!form) {
       return;
     }
-    this.wizard = form;
-    this.component.components = form.components || [];
+
+    const wizardForm = _.cloneDeep(form);
+
+    this.wizard = wizardForm;
+    this.component.components = wizardForm.components || [];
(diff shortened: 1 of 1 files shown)


AlexeyDavyd (approved):

DmitryNistratov (approved):

alexandraRamanenka (approved):

src/Wizard.js
```diff
@@ -733,10 +733,14 @@ export default class Wizard extends Webform {
     if (!form) {
       return;
     }
-    this.wizard = form;
```
travist, line 736: My only issue is that this will now perform a double clone deep since it passes this into the webform (which also performs a cloneDeep). For large wizards, we will see a performance penalty for this. Can we pass a flag into ```setForm``` called "reference" to tell the setForm method to not perform the cloneDeep? Like this?

```
return super.setForm(wizardForm, {...flags, reference: true})
```

travist (changes requested):

PR #3534, opened 2020-11-30
Tabs: indicate some of components on tab is invalid

mikekotikov:
@travist for now made this way, it looks more distinguish to me.
When the tab with error gets active, the color becomes more red.
image

But can make it the way you proposed, with an asterisk.
image


travist:
I prefer the asterisk...

travist (changes requested):
Please turn into asterisks. This will work with other CSS frameworks.

PR #3539, opened 2020-12-01
FJS-1411: fix displaying server errors

DmitryNistratov:


TanyaGashtold (approved):

src/Webform.js
```diff
@@ -1474,11 +1486,24 @@ export default class Webform extends NestedDataComponent {
   executeSubmit(options) {
     this.submitted = true;
     this.submitting = true;
+    this.clearServerErrors();
```
alexandraRamanenka, line 1489: Why did you place clearServerErrors in executeSubmit and not inside submit? Isn't it possible that beforeSubmit() throws an Error and then executeSubmit won't be executed?

alexandraRamanenka (approved):

src/Webform.js
DmitryNistratov: If I understand correctly, server errors should only be cleaned up when we have sent submission on the server. We will then receive a response that will either return updated errors or no errors at all. In either case, you need to clean up old mistakes

travist:
Needs to be rebased.

src/Webform.js
```diff
@@ -1266,6 +1274,10 @@ export default class Webform extends NestedDataComponent {
       if ('details' in error) {
         error = error.details;
       }
+      this.serverErrors = error.map((err) => {
+        err.fromServer = true;
```
travist, line 1278: why are you setting ALL the errors to be fromServer here?  How do we know they are from the server?

travist (changes requested):

PR #3540, opened 2020-12-01
FOR-2874: Added recaptcha async validation.

roflanKisel:


roflanKisel:
Relates to https://github.com/formio/formio/pull/1157

PR #3551, opened 2020-12-03
Fixed an issue where forms which have some logic get stuck in infinite loop

alexandraRamanenka:
Found by myself while testing PAP app


AlexeyDavyd (approved):

parasgera:
I think this is something related to #3549 which I had raised. If we merge this and the issue is fixed, we can close #3549.

alexandraRamanenka:
> I think this is something related to #3549 which I had raised. If we merge this and the issue is fixed, we can close #3549.

Yep, that is exactly the issue which this PR fixes

PR #3558, opened 2020-12-04
implement translations for static templates

mikekotikov:


src/Element.js
```diff
@@ -527,6 +527,9 @@ export default class Element {
    * @return {XML|string|*|void}
    */
   interpolate(string, data) {
+    if (typeof string !== 'function') {
+      string = FormioUtils.translateHTMLTemplate(String(string), (value) => this.t(value));
```
travist, line 531: I am concerned about performance here. Can we only apply this HTML template if this is the "content" component since that is really the only time it is needed?

travist (changes requested):

src/Element.js
mikekotikov: fixed

travist: what does this do?

```
(this.component.templates && Object.values(this.component.templates).some(template => !!template))
```

is this applying this for EditGrid?  Do we really need this?  Since they can override these templates, can we just add the translations using ```{{ ctx.t('Text') }}```

travist (changes requested):

src/components/_classes/component/Component.js
```diff
@@ -871,7 +871,7 @@ export default class Component extends Element {
     if (!template) {
       return '';
     }
-
+    data.t = this.t.bind(this);
```
travist, line 874: The ```t``` method is already added to the evalContext (I thought)...  as ```ctx.t()```.

The issue here is what happens if they have a field called "T" in their form. Won't this override that data with a function of ```t```?  This needs to change.

travist (changes requested):

src/components/_classes/component/Component.js
mikekotikov: In EditGrid to render the templates is used `renderString` method and it gives the error of `t` not defined, so I assigned it.
It will both work ok if I pass `t` in `evalContext` options.
Should I change it this way?

mikekotikov: Or I can pass it separately from `data` property in places where `renderString` is called in EditGrid component.

mikekotikov: Although, for now it is not actually overriding the data, but can be misleading. Will pass it from EditGrid.
image

src/components/editgrid/EditGrid.js
```diff
@@ -64,7 +64,7 @@ export default class EditGridComponent extends NestedArrayComponent {
       {% util.eachComponent(components, function(component) { %}
         {% if (!component.hasOwnProperty('tableView') || component.tableView) { %}
           <div class="col-sm-2">
-            {{ getView(component, row[component.key]) }}
+            {{ ctx.t(getView(component, row[component.key])) }}
```
travist, line 67: This is strange that we are calling ```ctx.t``` on the getView method. If I understand correctly, this performs an interpolation on a string that they configure in the row template. So, why can they not just provide the ```ctx.t()``` calls within those templates that they define?  This concerns me regarding performance and bad side effects.

travist (changes requested):

src/components/editgrid/EditGrid.js
mikekotikov: hm, don't actually remember for what reason I've done that. You're right, it can be done through templates configuration. Will get rid of this.

PR #3575, opened 2020-12-10
FJS-1342: added templates for form errors list and alerts

TanyaGashtold:


TanyaGashtold:
semantic: https://github.com/formio/semantic/pull/24
bootstrap3: https://github.com/formio/bootstrap3/pull/34

PR #3577, opened 2020-12-10
Element: convert lodash to object in evalContext

mikekotikov:


MaksimFalei (approved):

src/Element.js
```diff
@@ -505,7 +506,7 @@ export default class Element {
    */
   evalContext(additional) {
     return Object.assign({
-      _,
+      _: lodashOperators.reduce((obj, operator) => _.set(obj, operator, _[operator]), {}),
```
travist, line 509: Can we not execute this every time evalContext is called?  This happens a lot... I would rather add a global variable right below the ```import``` that says.

```js
const lodash = lodashOperators.reduce((obj, operator) => _.set(obj, operator, _[operator]), {}),
```

OR... you can have this exported by the lodash operators.  Then this would change to this.

```js
_: lodash
```

travist (changes requested):

Issue #3565, opened 2020-12-08
[HELP] Overriding URL in staticRequest hook


gs11:
I'm trying to override the URL in a static request which seems to be ignored. Inspecting the code it looks like `requestArgs.url` is simply ignored - https://github.com/formio/formio.js/blob/7c6eed023280e77c7f9e6786039a3a0b5b4829b0/src/Formio.js#L770-L782

Is there a reason to this or could the function call simply be modified to utilize `requestArgs.url` & `requestArgs.method` instead?

travist:
You should be able to do this with using a Fetch plugin, which is similar to having an interceptor.

https://github.com/formio/formio.js/wiki/Fetch-Plugin-API

gs11:
I'm currently using the Fetch plugin API to attach JWT tokens to all select component API calls. But as I'm guessing from the source snippet above, any changes to `url` or `method` are ignored @travist

travist:
Oh I see.... Yeah, we may need to modify this to allow a plugin to modify the URL. Please feel free to submit a pull request that adds this feature, and we will consider accepting it into the next version of the renderer.

gs11:
Thanks for merging! Closing..

PR #3578, opened 2020-12-10
Allow override of url & method in staticRequest hook

gs11:
This PR enables override of URL & Method from the `staticRequest` hook. The very same change could be done to the `request` hook if desired.

See #3565


travist:
Looks good. Thanks.

PR #3586, opened 2020-12-15
Allow multiple plugins to provide global options

### LINKED ISSUE #3514: Only the last loaded plugin will see its options used
Hi !
I would like to create several plugins providing custom builder groups to the Formio Builder.
The problem is, the way this is implemented today, only the last loaded plugin will see its options used.

https://github.com/formio/formio.js/blob/4cd7749ea22a3344b91f48cdce8ba4c14f8fa151/src/formio.form.js#L22-L24

I think each new `options` object should be merged with the previous one instead of overriding it.
Or is this the expected behaviour ?

Baloche:
Closes #3514

### DIFF
--- a/src/formio.form.js
+++ b/src/formio.form.js
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import AllComponents from './components';
 import Builders from './builders/Builders';
 import Components from './components/Components';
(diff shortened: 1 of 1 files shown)


travist:
Looks good!  Thanks!

PR #3589, opened 2020-12-16
4.8.x - FJS-1192: fixed excessive calls of conditional logic on chang…

TanyaGashtold:
…e event


TanyaGashtold:
all tests passed locally

PR #3593, opened 2020-12-17
Feat/UIP 253 abort file upload

mikekotikov:
Closed the previous version of same PR because of incorrect rebase.

### DIFF
--- a/src/Formio.js
+++ b/src/Formio.js
@@ -550,7 +550,7 @@ class Formio {
                 if (uploadStartCallback) {
                   uploadStartCallback();
                 }
-                return provider.uploadFile({ storage, file, fileName, dir, progressCallback, url, options, fileKey, groupPermissions, groupId, abortCallback });
+                return provider.uploadFile(file, fileName, dir, progressCallback, url, options, fileKey, groupPermissions, groupId, abortCallback);
               }
               else {
                 throw ('Storage provider not found');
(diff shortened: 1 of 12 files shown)


src/Formio.js
```diff
@@ -530,7 +530,7 @@ class Formio {
     });
   }
 
-  uploadFile(storage, file, fileName, dir, progressCallback, url, options, fileKey, groupPermissions, groupId, uploadStartCallback) {
```
travist, line 533: We cannot change the signature of these methods without it being considered a "breaking change". Please do this in a way where the method signature is reverse compatible.

travist (changes requested):

src/Formio.js
```diff
@@ -550,7 +550,7 @@ class Formio {
                 if (uploadStartCallback) {
                   uploadStartCallback();
                 }
-                return provider.uploadFile(file, fileName, dir, progressCallback, url, options, fileKey, groupPermissions, groupId);
+                return provider.uploadFile({ storage, file, fileName, dir, progressCallback, url, options, fileKey, groupPermissions, groupId, abortCallback });
```
travist, line 553: Isn't this still changing the function signature? To accept an object instead of arguments?...  If so, then we cannot change function signatures within a major version. ( we can do this in 5.x but not 4.x).

travist (changes requested):

src/Formio.js
mikekotikov: I see
Thought we must keep it untouched only in Formio
Will change it back in provider files too then

PR #3594, opened 2020-12-17
FIO-994: Makes onlyAvailableItems validation optional

alexandraRamanenka:


MaksimFalei (approved):

src/components/select/editForm/Select.edit.validation.js
```diff
@@ -2,12 +2,32 @@ export default [
...
     key: 'validate.select',
     input: true,
     conditional: {
       json: { var: 'data.searchField' }
     }
+  },
+  {
+    weight: 52,
+    type: 'checkbox',
+    label: 'Perform server validation of static value',
```
travist, line 16: I am really confused why we need to make a distinction of ```server validation```.  Why can we just not make this a setting that is ```Perform validation of select values``` and is executed on both client and server when checked. Otherwise, it will just skip this validation?  I cannot think of a use case where someone would want to validate it on the server, but not also on the client. They either want it all on or all off. Having these two separate configurations is confusing.

travist (changes requested):

PR #3599, opened 2020-12-19
VPAT-806 Added error list hotkey message for further usage in USWDS

antonSoftensity:


antonSoftensity:
[VPAT-806 | FMG | Chrome/NVDA | Enter Data Form | There is no redirect to the error from the error section after pressing Enter button on the error](https://trello.com/c/i0T8uqJ0/806-vpat-806-fmg-chrome-nvda-enter-data-form-there-is-no-redirect-to-the-error-from-the-error-section-after-pressing-enter-button-on)

alina-softensity (approved):

PR #3600, opened 2020-12-21
VPAT-804 Fixed issue when modal edit window was resetting on change

antonSoftensity:


antonSoftensity:
[VPAT-804 | Regression | FMG | Modal Edit | All browsers | After saving value in Modal Edit view dialogue window of Component 1, focus doesn't go through all components in Modal Edit view dialogue window of Component 2, moves outside dw and doesn't return to dw](https://trello.com/c/FK91XCLb/804-vpat-804-regression-fmg-modal-edit-all-browsers-after-saving-value-in-modal-edit-view-dialogue-window-of-component-1-focus-doesn)

alina-softensity (approved):

PR #3605, opened 2020-12-22
fix 'show label in DataGrid' option

mikekotikov:


src/components/datagrid/DataGrid.js
```diff
@@ -458,7 +458,7 @@ export default class DataGridComponent extends NestedArrayComponent {
       const component = this.createComponent(columnComponent, options, row);
       component.parentDisabled = !!this.disabled;
       component.rowIndex = rowIndex;
-      component.inDataGrid = true;
```
travist, line 461: We should not be mutating the "component" object.  This is what causes infinite loop errors with the Logic system because the "component" changed. Why is this necessary?

travist (changes requested):

src/components/datagrid/DataGrid.js
mikekotikov: The option to 'Show Label in DataGrid' was not available at builder because the custom conditional didn't come true.
`inDataGrid` just wasn't assigned to `editComponent`.
image

mikekotikov: Will find another way to pass it

MaksimFalei (approved):

PR #3606, opened 2020-12-22
Localize selectboxes validation messages

shawnbot:
The selectboxes component has some hard-coded validation messages for `minSelectedCount` and `maxSelectedCount`. This change wraps both the hard-coded messages and the custom validation messages (`minSelectedCountMessage` and `maxSelectedCountMessage`) in `this.t()` so that they're passed through the localization layer. This should make it possible to localize them like so:

```js
Formio.createForm({
  components: [
    {
      type: 'selectboxes',
      values: [
        { label: 'Option 1', value: 1 },
        { label: 'Option 2', value: 2 },
        { label: 'Option 3', value: 3 },
        { label: 'Option 4', value: 4 }
      ],
      validate: { maxItemCount: 2 }
    }
  ]
}, {
  i18n: {
    'You can only select up to {{maxCount}} items.': 'Please select {{maxCount}} items or fewer.'
  }
})
```

I've added tests for both the default messages and the custom ones for both the min and max count scenarios.

### DIFF
--- a/src/components/selectboxes/SelectBoxes.js
+++ b/src/components/selectboxes/SelectBoxes.js
@@ -176,8 +176,8 @@ export default class SelectBoxesComponent extends RadioComponent {
         }
         if (maxCount && count > maxCount) {
           const message = this.component.maxSelectedCountMessage
-            ? this.component.maxSelectedCountMessage
-            : `You can only select up to ${maxCount} items.`;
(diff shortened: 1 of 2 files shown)


travist:
As always, thanks @shawnbot for the solid contribution and unit tests!

PR #3608, opened 2020-12-24
revises Checkbox and SelectBox components on Radio component model to support styling

adventmedia:
The Radio component allows styling for selected and unselected states to allow custom styling. This branch updates the Checkbox and Select Boxes components to support the same styling options, as shown in this screen grab.

screenshot_22


travist:
Thank you for your contribution! Does this PR change the default styling or does this just provide the CSS classes that allow you to do such styling? If we want to accept this, it will need to be reverse compatible and cannot change the default styles for checkboxes.

adventmedia:
This does not change the default styling, it supports custom styling for checkbox and selectbox components in the same manner as the radiobutton component. However I’ve closed the pr because I’ve discovered that the components do not show the checked state when viewed in readonly mode. Hopefully I’ll be able to fix this.

regards,
Andrew



Andrew Leith Macrae
Advent Media
Interactive Application Development
1008-720 Spadina Ave
Toronto ON CA M5S 2T9
416.966.3147
Skype: andrewmacrae
adventmedia.net

PR #3613, opened 2020-12-29
Translation issue Fixed

syed-farjana:
There is translation  issue in edit grid validation. Added the translation code like the other component


travist:
There is now a conflict. Please resolve and we will pull in this change.

syed-farjana:
Resolved the conflict please Review

src/components/editgrid/EditGrid.js
```diff
@@ -937,7 +937,9 @@ export default class EditGridComponent extends NestedArrayComponent {
           const errorContainer = rowContainer.querySelector('.editgrid-row-error');
 
           if (!rowValid) {
-            errorContainer.textContent = this.t('invalidRowError');
```
travist, line 940: So, this should be picked up with the following.

https://github.com/formio/formio.js/blob/master/src/i18n.js#L11

Why is this not registering correctly?  I feel like this is not fixing the actual bug here...

travist (changes requested):

syed-farjana:
I think the above changes are in master branch.  but in 4.12.x the above  translation changes are not there.

travist:
Then we should Port the "master" changes into the 4.12.x branch, and that should be included in this pull request.

syed-farjana:
you mean it de we need to merge master and 4.12.x. I guess I can't do this. Can you suggest what I can do here

travist:
You can branch off of the 4.12.x branch, and then when you submit your pull request, you can change the "base" branch to merge into. It is part of the Github pull request process.

PR #3617, opened 2020-12-29
Update formio.form.scss

kamaloush:
@mikekotikov 
Please review this commit. Your last change made the view a little broken on mobile devices, impossible to scroll to the right when component overflows.

### DIFF
--- a/src/sass/formio.form.scss
+++ b/src/sass/formio.form.scss
@@ -989,10 +989,6 @@ body.formio-dialog-open {
   }
 }
 
-.table-responsive[ref=component] {
-  overflow-x: visible;
-}
-
 .formio-component-textarea {
   .alert .ck-editor__editable {
     color: inherit;


travist:
@mikekotikov please provide description on why this css was added to begin with. I am pretty sure this will break what he fixed so would like his input on what was the original issue fixed here.

kamaloush:
`overflow-x: auto`
or
`overflow-x: scroll`
Maybe it will fix your orignal issue and my issue.

mikekotikov:
@kamaloush thanks for noticing the bug.
This styles actually relate to one problem, so I feel better taking care of it myself. 
Will provide a link to my PR here after it's done.

mikekotikov:
@kamaloush can you please provide the json of the form where you experienced this bug?

kamaloush:
@mikekotikov 
I tried to create a Jsfiddle in order to reproduce the bug :
[http://jsfiddle.net/zyvuex27/](http://jsfiddle.net/zyvuex27/)
But it is not revelant.
It seems to be only a bug in my ionic implementation.
I close the PR.

PR #3641, opened 2021-01-10
Add option for errors to builder

juliapath:
It would be nice to be able to set this in the builder. I decided for a simple JSON input, as I didn't want to hunt for all possible options you can set here and then forget some. Also this won't create any additional noise in the `schema`, whereas a `container` with a `textfield` for every option would.

### DIFF
--- a/src/components/_classes/component/editForm/Component.edit.validation.js
+++ b/src/components/_classes/component/editForm/Component.edit.validation.js
@@ -134,6 +134,71 @@ export default [
         input: true
       }
     ]
+  },
+  {
+    type: 'panel',
+    title: 'Custom Errors',
+    collapsible: true,
+    collapsed: true,
+    key: 'errors',
+    weight: 400,
+    components: [
+      {
+        type: 'textarea',
+        key: 'errors',
+        hideLabel: true,
+        rows: 5,
+        editor: 'ace',
+        as: 'json',
+        input: true
+      },
+      {
+        type: 'htmlelement',
+        tag: 'div',
+        content: `
+          <p>This allows you to set different custom error messages for different errors
+          (in contrast to “Custom Error Message”, which only allows you to set one
+          error message for all errors). E.g.</p>
+
+<pre>{
+  "required": "{<span/>{ field }} is required. Try again.",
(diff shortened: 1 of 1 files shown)


src/components/_classes/component/editForm/Component.edit.validation.js
```diff
@@ -134,6 +134,71 @@ export default [
...
+            <li>i<span/>nvalid_regex</li>
+            <li>m<span/>ask</li>
+            <li>p<span/>attern</li>
+            <li>c<span/>ustom</li>
+          </ul>
+
+          <p>Depending on the error message some of the following template variables can be used in the script:</p>
+          <ul>
+           <li><code>{<span/>{ f<span/>ield }}</code> is replaced with the label of the field.</li>
+           <li><code>{<span/>{ m<span/>in }}</code></li>
```
juliapath, line 189: Not sure if the `<span/>` is the best way to prevent formio from substituting templates and translations.

travist:
I do not believe that ```component.errors``` is a real component property. What happens when this is set?

juliapath:
It's documented here: https://github.com/formio/formio.js/wiki/Components-JSON-Schema

> Allows customizable errors to be displayed for each component when an error occurs. This is an object with the following keys
>
> - required
> - min
> - max
> - minLength
> - maxLength
> - invalid_email
> - invalid_date
> - pattern
> - custom

This list is a non exhaustive though.

`component.errors` is accessed here:

https://github.com/formio/formio.js/blob/de2a7f4c5db3bda4c7766540ae651a5b5cdb6f8d/src/components/_classes/component/Component.js#L1586-L1593

Which in turn is called in Validator.js, for example here:

https://github.com/formio/formio.js/blob/de2a7f4c5db3bda4c7766540ae651a5b5cdb6f8d/src/validator/Validator.js#L65-L74

Setting `errors` is important, as it allows customization of individual error messages for different situations. `customErrorMessage` just allows you to define one single error message, annoyingly even overwriting the error messages you return in a Custom Validation.

travist:
Ok great! Thank you!

PR #3642, opened 2021-01-10
Use schema instead of component for Builder Edit JSON dialog

juliapath:
This changes the the “Edit JSON” dialog so it uses the compact `schema` value
without all the default settings.

### Before
before
### After
after


travist:
I like this change a lot.  Thank you!

Issue #3638, opened 2021-01-07
[BUG] Datagrid columns do not adjust to content size


elibus:
## Environment
Please provide as many details as you can:

- Hosting type
  - [x] Form.io
  - [ ] Local deployment
    - Version: 
- Formio.js version: 4.13.0-rc.5
- Frontend framework:
- Browser:
- Browser version:

## Steps to Reproduce
1. Go here: https://jsfiddle.net/zvn03yq2/
2. Check out section 4.a. The datagrid has equally sized columns. Width do not adjust to content size anymore.

## Expected behavior
Column width adjusts to content size.
We would expect this behavior with "Equal column width" option checked. 

## Observed behavior
`td` HTML elements get a `col-md-2` class which is forcing the content to fit two bootstrap columms.

## Example
https://jsfiddle.net/zvn03yq2/

Introduced with PR #3544

elibus:
Hi @heather-formio I am tagging you so you this is Marco

olgabann:
We're currently addressing a backlog of GitHub issues. Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #3658, opened 2021-01-14
Fix/4.13.0 regression datagrid columns width

mikekotikov:
@travist tried my best to make it look ok both in formio-app and in angular-demo, but wasn't able to achieve it =(
Whether I applied logic with counting and setting col-md-[number] classes to every column or tried to apply different styles and other related bootstrap classes, nothing worked out well to me to be used with both repos.
So, was forced to stop on making DataGrid table layout fixed when there is at least one Select component.

### DIFF
--- a/src/components/datagrid/DataGrid.js
+++ b/src/components/datagrid/DataGrid.js
@@ -236,7 +236,7 @@ export default class DataGridComponent extends NestedArrayComponent {
 
   render() {
     const columns = this.getColumns();
-    const colWidth = Math.floor(12 / (columns.length + 1)).toString();
+    const layoutFixed = columns.some(col => col.type === 'select') || this.component.layoutFixed;
     return super.render(this.renderTemplate('datagrid', {
       rows: this.getRows(),
       columns: columns,
@@ -259,7 +259,7 @@ export default class DataGridComponent extends NestedArrayComponent {
       placeholder: this.renderTemplate('builderPlaceholder', {
         position: this.componentComponents.length,
       }),
-      colWidth
+      layoutFixed
     }));
   }
 
--- a/src/templates/bootstrap/table/form.ejs
+++ b/src/templates/bootstrap/table/form.ejs
@@ -6,16 +6,16 @@
   ">
   {% if (ctx.component.header && ctx.component.header.length > 0) { %}
   <thead>
-    <tr>
+    <tr class="row">
       {% ctx.component.header.forEach(function(header) { %}
-      <th>{{ctx.t(header)}}</th>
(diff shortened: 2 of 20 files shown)


AlexeyDavyd (approved):

elibus:
Fix #3638

src/templates/bootstrap/table/form.ejs
```diff
@@ -6,16 +6,16 @@
   ">
   {% if (ctx.component.header && ctx.component.header.length > 0) { %}
   <thead>
-    <tr>
+    <tr class="row">
       {% ctx.component.header.forEach(function(header) { %}
-      <th>{{ctx.t(header)}}</th>
+      <th class="col-md-{{ctx.colWidth}}">{{ctx.t(header)}}</th>
```
paulfalgout, line 11: what?  `colWidth` was removed

test/renders/form-semantic-formWithCustomFormatDate.html
```diff
@@ -24,7 +24,7 @@
         </thead>
         <tbody ref="datagrid-dataGrid-tbody">
           <tr ref="datagrid-dataGrid-row">
-            <td ref="datagrid-dataGrid" class="col-md-3">
+            <td ref="datagrid-dataGrid" class="col-md-">
```
paulfalgout, line 27: Is `col-md-` a class?  If not, isn't this testing that this is just broken?  This was added in a v4.13 rc..

src/components/datagrid/DataGrid.js
```diff
@@ -236,7 +236,7 @@ export default class DataGridComponent extends NestedArrayComponent {
 
   render() {
     const columns = this.getColumns();
-    const colWidth = Math.floor(12 / (columns.length + 1)).toString();
+    const layoutFixed = columns.some(col => col.type === 'select') || this.component.layoutFixed;
```
paulfalgout, line 239: So `select`s certainly _can_ be added to a table without fixed layout without being broken.  This seems like either a fix to a particular use case or further breaking which started in https://github.com/formio/formio.js/pull/3544 but this breaks perfectly working forms from 4.12

travist: I agree that we should look into this. We will investigate.

parasgera: @mikekotikov @travist 
This resulted in a prod issue for us :(
If someone has a select component inside a datagrid, the fixed layout is making the datagrid look weird. 
image

travist:
I believe we will need to revert this pull request. It is causing many more problems than it solves. @mikekotikov please find another way to solve this problem.

PR #3688, opened 2021-01-22
VPAT-870 Changed attachComponent function for further extension in VPAT

antonSoftensity:


antonSoftensity:
[VPAT-870 | FMG | Create/Edit Form | Actions buttons of Layout/Data components aren't hidden when another component in Layout/Data component is focusable and focus moves from component actions buttons to layout/data component actions buttons after using Up/Down arrow keys](https://trello.com/c/hlM97HOv/870-vpat-870-fmg-create-edit-form-actions-buttons-of-layout-component-arent-hidden-when-another-component-in-layout-component-is-foc)

alina-softensity (approved):

PR #3691, opened 2021-01-25
FIO-246 (FJS-1473): removed fixedSize setting from PDF overlay settings

TanyaGashtold:


TanyaGashtold:
PR for formio-pdf: https://github.com/formio/formio-pdf/pull/81

PR #3693, opened 2021-01-25
FIO-1217: Fixes masks caret behavior when it is used inside shadow DOM

alexandraRamanenka:
Also required changes in text-mask lib. I have already created a PR https://github.com/text-mask/text-mask/pull/1034


src/Element.js
```diff
@@ -372,10 +372,13 @@ export default class Element {
         if (input.mask) {
           input.mask.destroy();
         }
+        // eslint-disable-next-line no-debugger
+        debugger;
```
travist, line 376: Please remove the debugger statement.

travist:
We will need to refer to our fork until they pull this in.

PR #3698, opened 2021-01-26
FIO-3530: update input attribute when value changes

### LINKED ISSUE #3530: [BUG] Radio buttons are not re-rendered when selected value changes
## Environment

- Hosting type
  - [x] Form.io
  - [ ] Local deployment
    - Version: 
- Formio.js version: 4.12.1
- Frontend framework: Bootstrap 4
- Browser: Chrome
- Browser version: 86

## Steps to Reproduce
1. open the form with browser's devtools and inspect the elements
2. select an option in radio
3. observe the HTML changes

## Expected behavior

According to the `template` file ( https://github.com/formio/formio.js/blob/ac0966b2e2154c03c795f3df08c1bb06f3699e25/src/templates/bootstrap/radio/form.ejs#L14-L16), the `checked` attr should be set according to the `ctx.value`.

## Observed behavior

However, when I change the selected option, the `checked` attr didn't change.

The only HTML change is the `class` on the `<div ref="wrapper">`. While debugging it with custom template for `radio`, I found that the template is not called when changing selected option.

I am not relying on the `checked` attr, but I want to re-render the radio with template when selected option changes. The reason behind it is that I want to use `ctx.component.properties` in template to extend the display and behaviour of `radio`. If the `radio` is not re-rendered, I am not able to display my extended version correctly.

## Example

demo: https://pro.formview.io/#/ehdkompeqarzcuk/radiobug?header=1&reset=1

image

zzdjk6:
Fix #3530 

- Checkbox
- Radio

### DIFF
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -180,4 +180,19 @@ export default class CheckBoxComponent extends Field {
   getValueAsString(value) {
     return value ? 'Yes' : 'No';
   }
+
+  updateValue(value, flags) {
+    const changed = super.updateValue(value, flags);
+
+    // Update attributes of the input element
+    if (changed && this.input) {
+      if (this.input.checked) {
+        this.input.setAttribute("checked", "true");
+      } else {
+        this.input.removeAttribute("checked");
+      }
+    }
+
+    return changed;
+  }
 }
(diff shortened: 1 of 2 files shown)


travist:
Thank you!

PR #3699, opened 2021-01-26
Add .offOne method to remove one listener for a certain event

zzdjk6:
Method`.off()` of `Element` is removing all listeners for a single event, and we need a way to remove a single listener for a single event as the exact opposite to `.on()`, so that we added `.offOne()`.

The reason behind it is that we want to attach a listener to a single event, and remove it when cleanup of one part without removing all other listeners.

This is particularly important when using `form.io` with `React` -- we want to attach and detach event listeners in `useEffect()` according to the changes of the dependency array.

The example code looks like below:

```jsx
const MyComponent = ({ form, prop }) => {
  useEffect(() => {
    const doOnChange = () => {
        // Do something according to prop
    };

    form.on("change", doOnChange);

    // Clean up
    return () => {
      form.offOne("change", doOnChange);
    };
  }, [form, prop]);
};
```

If we don't remove the listener attached using `.on()`, we have memory leak;
If we remove the listener using `.off()`, it will affect all other listeners for the same event.

In this case, `.offOne()` is the answer:

```js
  /**
   * Remove one listener for a certain event. This is the exact opposite to `.on()`
   * 
   * @param {string} event - The event you wish to register the handler for.
   * @param {function} cb - The callback handler to handle this event.
   */
  offOne(event, cb) {
    if (!this.events) {
      return;
    }
    const type = `${this.options.namespace}.${event}`;

    const listener = this.events
      .listeners(type)
      .find(item => item && item === cb && item.id === this.id);
    if (!listener) {
      return;
    }

    this.events.off(type, listener);
  }
```

### DIFF
--- a/src/Element.js
+++ b/src/Element.js
@@ -151,6 +151,28 @@ export default class Element {
   }
 
   /**
+   * Remove one listener for a certain event. This is the exact opposite to `.on()`
+   * 
+   * @param {string} event - The event you wish to register the handler for.
+   * @param {function} cb - The callback handler to handle this event.
+   */
+  offOne(event, cb) {
+    if (!this.events) {
+      return;
+    }
+    const type = `${this.options.namespace}.${event}`;
+
+    const listener = this.events
+      .listeners(type)
+      .find(item => item && item === cb && item.id === this.id);
+    if (!listener) {
+      return;
+    }
+
+    this.events.off(type, listener);
+  }
+
+  /**
    * Emit a new event.
    *
    * @param {string} event - The event to emit.
--- a/src/Element.js
+++ b/src/Element.js
@@ -130,46 +130,37 @@ export default class Element {
   }
 
   /**
-   * Removes all listeners for a certain event.
+   * Removes a listener for a certain event. Not passing the 2nd arg will remove all listeners for that event.
    *
-   * @param event
-   */
-  off(event) {
-    if (!this.events) {
-      return;
-    }
-    const type = `${this.options.namespace}.${event}`;
-
-    // Iterate through all the internal events.
-    _.each(this.events.listeners(type), (listener) => {
-      // Ensure this event is for this component.
-      if (listener && (listener.id === this.id)) {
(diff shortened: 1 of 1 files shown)


travist:
I would rather just add an additional argument to the ```off``` method where you can pass an optional callback. If you do, then it will remove just the specific event.

zzdjk6:
@travist Sounds good, will update the PR.

zzdjk6:
@travist I have updated PR as suggested, please have a look

travist:
This code iterates through the listeners 3 different times. If there are a lot of listeners, then this is 3x slower than it needs to be. Does the following work?

```js
off(event, cb) {
    if (!this.events) {
      return;
    }
    const type = `${this.options.namespace}.${event}`;
    this.events.listeners(type).forEach((listener) => {
      if (listener.id === this.id && (!cb || cb === listener)) {
          this.events.off(type, listener);
      }
    });
  }
```

zzdjk6:
@travist Thanks for the feedback, and it works.

Most of time, I would prefer clarity over performance, and that's why I was trying to implement that using two `filter` functions and one `forEach` at the end, so that each step is simple and clear.

Saying that, performance might be more important from a library perspective, and I would slightly change the proposed solution as below:

```js
  /**
   * Removes a listener for a certain event. Not passing the 2nd arg will remove all listeners for that event.
   *
   * @param {string} event - The event you wish to register the handler for.
   * @param {function|undefined} cb - The callback handler to handle this event.
   */
  off(event, cb) {
    if (!this.events) {
      return;
    }

    const type = `${this.options.namespace}.${event}`;

    this.events.listeners(type).forEach((listener) => {
      // Ensure the listener is for this element
      if (!listener || listener.id !== this.id) {
        return;
      }

      // If there is a given callback, only deal with the match
      if (cb && cb !== listener) {
        return;
      }

      this.events.off(type, listener);
    });
  }
```

I believe in this way, it is both performant and clear 😄

travist:
Looks good. Thanks!

PR #3701, opened 2021-01-26
FIO-1125: Any value used in Translate list is returning contains an invalid selection

AlehKatsiubaFormIO:


AlehKatsiubaFormIO:
Related to:
https://github.com/formio/bootstrap3/pull/38

travist:
Needs to be rebased.

PR #3705, opened 2021-01-27
Fixed the translation issues while saving edit grid

syed-farjana:
I have created new branch and created pull request to 4.12.x. please review the changes. This change is a fix to the translation issue when when saving the edit grid


travist:
Looks good. Thanks!

syed-farjana:
I can see this change in 4.12.x. In our application we are using 4.12.1-rc.16. I think the change is still not there in 4.12.1-rc.16. Can we have this change in specified branch or tag

travist:
I have not created the next 4.12.x version of our renderer but may do that today...

PR #3711, opened 2021-01-27
support for advanced styling

adventmedia:
based on current usage for the Radio button component, which adds/removes a class to a wrapper div based on the selected state of the button; applied to Checkbox component as well.
Default styling is not affected.

### DIFF
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -180,11 +180,13 @@ export default class CheckBoxComponent extends Field {
   updateValue(value, flags) {
     const checkBoxChecked = 'checkbox-checked';
     const val = this.getValue();
-    if (val) {
-      this.addClass(this.refs.formCheck, checkBoxChecked);
-    }
-    else {
-      this.removeClass(this.refs.formCheck, checkBoxChecked);
+    if (this.refs.formCheck) {
+      if (val) {
+        this.addClass(this.refs.formCheck, checkBoxChecked);
+      }
+      else {
+        this.removeClass(this.refs.formCheck, checkBoxChecked);
+      }
     }
     super.updateValue(value, flags);
   }
--- a/src/components/checkbox/Checkbox.js
+++ b/src/components/checkbox/Checkbox.js
@@ -80,7 +80,7 @@ export default class CheckBoxComponent extends Field {
   }
 
   attach(element) {
-    this.loadRefs(element, { input: 'multiple' });
+    this.loadRefs(element, { input: 'multiple', formCheck: 'single' });
     this.input = this.refs.input[0];
     if (this.refs.input) {
       this.addEventListener(this.input, this.inputInfo.changeEvent, () => this.updateValue(null, {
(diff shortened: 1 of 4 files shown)


adventmedia:
@travist wondering what sort of timeline we can hope for this to be merged and released?

src/components/checkbox/Checkbox.js
```diff
@@ -177,6 +177,18 @@ export default class CheckBoxComponent extends Field {
     return false;
   }
 
+  updateValue(value, flags) {
+    const checkBoxChecked = 'checkbox-checked';
+    const val = this.getValue();
+    if (val) {
+      this.addClass(this.refs.formCheck, checkBoxChecked);
```
travist, line 184: Need to protect against templates that do not have this reference. It will crash otherwise.

src/components/radio/Radio.js
```diff
@@ -158,12 +158,26 @@ export default class RadioComponent extends Field {
...
+          case 'radio' :
         if (input && input.value.toString() === value.toString()) {
           //add class to container when selected
           this.addClass(wrapper, optionSelectedClass);
         }
         else {
           this.removeClass(wrapper, optionSelectedClass);
+        }
+            break;
+          case 'checkbox' :
```
travist, line 171: When would a "radio" ever be a checkbox?  I believe this should be using the Select Boxes component instead.

src/components/radio/Radio.js
```diff
@@ -178,7 +192,7 @@ export default class RadioComponent extends Field {
       && this.previousValue === this.currentValue;
     if (shouldResetValue) {
       this.resetValue();
-      this.triggerChange(flags);
+      this.triggerChange();
```
travist, line 195: Why was this changed? This will certainly introduce a bug somewhere if we accept this. We should be passing along the flags since they are commonly used to flag behaviors during the onChange event loop.

src/templates/bootstrap/radio/form.ejs
```diff
@@ -1,6 +1,8 @@
 <div class="form-radio radio">
   {% ctx.values.forEach(function(item) { %}
-  <div class="form-check{{ctx.inline ? '-inline' : ''}}" ref="wrapper">
+  <div class="form-check{{ctx.inline ? '-inline' : ''}} {% if ((ctx.disabled && ctx.value) && (ctx.value === item.value || (typeof ctx.value === 'object' && ctx.value.hasOwnProperty(item.value) && ctx.value[item.value]))) { %}
```
travist, line 3: This much logic should be placed within the class and then passed into the template as a ctx variable.

travist (changes requested):
I feel a little weary about pulling this in. I gave some general comments so please respond to them so that I can understand your motivation for this change.

src/components/radio/Radio.js
adventmedia: this is to support SelectBoxes component
`export default class SelectBoxesComponent extends RadioComponent`

src/components/radio/Radio.js
adventmedia: fixed (working from 4.12.4)

src/templates/bootstrap/radio/form.ejs
adventmedia: this is done

adventmedia:
these changes are to support CSS restyling of the input components (esp. for mobile devices). It is necessary to add or remove a class from the "wrapper" div based on whether the input is selected or not. Radio component already supports this behaviour.
screenshot_134

PR #3715, opened 2021-01-28
[3.x] FIO-950: Add usage of image processing

RomanLetsuk:


src/components/file/File.js
```diff
@@ -97,9 +97,22 @@ export default class FileComponent extends BaseComponent {
...
+      .then(result => result.url);
+
+    if (!this.component.imageProcessing || !this.component.imageProcessing.trim().length) {
+      return imageUrlPromise;
+    }
+
+    return imageUrlPromise
+      .then((result) => {
+        return Formio
+          .makeStaticRequest(`${Formio.baseUrl}/image-processing`, 'POST', {
```
travist, line 110: There are a few changes I want to make here.

1. I do not want to make the URL go to the API server but be generic.
2. This should not be a component configuration, but rather a configuration passed into the renderer.

Here is what I suggest.  We will make a renderer option as follows.

```js
Formio.createForm(document.getElementById('formio'), 'https://yourproject.form.io/imageupload', {
  imageProcessor: {
    url: 'https://someotherdomain....',
    options: {
       .... whatever you were putting in "component.imageProcessing" ...
    }
  }
});
```

travist (changes requested):

PR #3746, opened 2021-02-05
FIO-1273: disabled possibility to drop components in dataMap and fixed…

TanyaGashtold:
… an issue where dataMap doesnot save changes for value component

### DIFF
--- a/src/WebformBuilder.js
+++ b/src/WebformBuilder.js
@@ -188,7 +188,7 @@ export default class WebformBuilder extends Component {
       containerElement.formioComponent = component;
 
       // Add container to draggable list.
-      if (this.dragula && this.allowDrop(element)) {
+      if (this.dragula && this.allowDrop(element) && component.type !== 'datamap') {
         this.dragula.containers.push(containerElement);
       }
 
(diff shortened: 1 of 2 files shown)


TanyaGashtold:
All tests passed locally. Updated renders for failed tests are in PR: 3738

TanyaGashtold:
Commit with updated test renders: https://github.com/formio/formio.js/pull/3738/commits/50774421076210211b7aa80b847ee7e49b43e241

src/WebformBuilder.js
```diff
@@ -188,7 +188,7 @@ export default class WebformBuilder extends Component {
       containerElement.formioComponent = component;
 
       // Add container to draggable list.
-      if (this.dragula && this.allowDrop(element)) {
+      if (this.dragula && this.allowDrop(element) && component.type !== 'datamap') {
```
travist, line 191: I don't like that we are hardcoding the component key type here. I would rather that this is injected into this method by the datamap class.

Also, they should be able to change out the value component, just not the key component. This change looks like it may remove the ability to change the value component.

travist (changes requested):

PR #3759, opened 2021-02-10
get rid of placeholder set automatically in CalendarWidget

mikekotikov:


travist:
The problem with this is that this would break if they want to provide a placeholder. We should only fix it so that it will not show the "default" placeholder which is the input mask.

mikekotikov:
> The problem with this is that this would break if they want to provide a placeholder. We should only fix it so that it will not show the "default" placeholder which is the input mask.

The placeholders functionality works fine as it's done without using removed method.

PR #3762, opened 2021-02-11
fix resource forms didn't transfer settings to pdf

mikekotikov:


MaksimFalei (approved):

src/PDFBuilder.js
```diff
@@ -450,9 +450,15 @@ export default class PDFBuilder extends WebformBuilder {
 
     const element = e.target;
     const type = element.getAttribute('data-type');
-
+    const key = element.getAttribute('data-key');
+    const group = element.getAttribute('data-group');
     const schema = fastCloneDeep(this.schemas[type]);
 
+    if (key && group) {
+      const info = this.getComponentInfo(key, group);
```
travist, line 458: This fix kind of seems like a hack to me. Can we discuss this fix in developer support meeting?

travist (changes requested):

travist:
We discussed, and it is fine. This will be included in next build.

MaksimFalei (approved):

PR #3766, opened 2021-02-12
FIO-1068: Fixes an issue where revision of the Nested Form loaded on the front-end side differs from the one which loaded on the server-side

alexandraRamanenka:
That issue appears in cases when form was imported or deployed from the older version of builder where revision of the Nested Form was specified using 'formRevision' property. 
Forms that has 'formRevision' property set to any number and 'revision' set to '' (which is default and means to load the current revision), the current version of the form is loaded on the front-end side and the one which 'formRevision' property refers to is loaded on the server-side (eg, for validation)

### DIFF
--- a/src/components/form/editForm/Form.edit.form.js
+++ b/src/components/form/editForm/Form.edit.form.js
@@ -46,6 +46,22 @@ export default [
     weight: 11,
   },
   {
+    type: 'textfield',
+    input: true,
+    label: 'Form Revision (Deprecated)',
+    description: `
+      <div class="alert alert-danger">
+        Having 'formRevision' property means that you form was configured in the older version of builder.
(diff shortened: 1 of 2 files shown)


AlexeyDavyd (approved):

src/components/form/editForm/Form.edit.form.js
```diff
@@ -45,6 +45,22 @@ export default [
     key: 'revision',
     weight: 11,
   },
+  {
+    type: 'textfield',
+    input: true,
+    label: 'Form Revision (Deprecated)',
```
travist, line 51: Need to remove the edit form and just make it have an intelligent automatic fallback.

travist (changes requested):

PR #3775, opened 2021-02-15
VPAT-893 Added attachTooltip function for further VPAT override

antonSoftensity:


antonSoftensity:
[VPAT-893 | FMG | Regression | Create/Edit Form | There is no padding between tooltip and focus outline of action button, when action button is in focus in Form Builder](https://trello.com/c/emgzgySA/893-vpat-893-fmg-regression-create-edit-form-there-is-no-padding-between-tooltip-and-focus-outline-of-action-button-when-action-butt)

alina-softensity (approved):

PR #3777, opened 2021-02-16
add global option to show full component JSON schema

mikekotikov:


mikekotikov:
@travist should I document somewhere the `showFullJsonSchema` builder option made here?

travist:
Yes, the docs for these are @ https://github.com/formio/formio.js/wiki/Form-Renderer#options

PR #3779, opened 2021-02-16
FIO-1352: fix validation for the case when the day is set to dayFirst

DmitryNistratov:


src/validator/Validator.js
```diff
@@ -668,7 +668,12 @@ class ValidationChecker {
           if (component.isPartialDay && component.isPartialDay(value)) {
             return true;
           }
-          const date = moment(value);
+          let date = moment(value);
```
travist, line 671: Why are we initializing this twice?  Can we just make this the following?

```js
const date = component.getValidationFormat ? moment(value, component.getValidationFormat()) : moment(value);
```

travist (changes requested):

PR #3796, opened 2021-02-19
custom styling support for checkbox and select box

adventmedia:
applies pattern for Radio component to Checkbox and Select Box components.


src/components/radio/Radio.js
```diff
@@ -161,24 +161,52 @@ export default class RadioComponent extends Field {
...
-        if (input && input.value.toString() === value.toString()) {
-          //add class to container when selected
-          this.addClass(wrapper, optionSelectedClass);
-          //change "checked" attribute
-          input.setAttribute('checked', 'true');
-        }
-        else {
-          this.removeClass(wrapper, optionSelectedClass);
-          input.removeAttribute('checked');
+        switch (input.type) {
```
travist, line 184: This seems like the same code but just checks for different "values". Why not just change the code to say this?

```js
const checked = (input.type === 'checkbox') ? value[input.value] : (input.value.toString() === value.toString());
if (checked) {
  this.addClass(wrapper, optionSelectedClass);
  input.setAttribute('checked', 'true');
}
else {
  this.removeClass(wrapper, optionSelectedClass);
  input.removeAttribute('checked');
}
```

src/components/radio/Radio.js
```diff
@@ -161,24 +161,52 @@ export default class RadioComponent extends Field {
   }
 
   setSelectedClasses() {
+    switch (this.inputInfo.attr.type) {
```
travist, line 164: This can just be changed to...

```js
this.component.values.forEach(item => {
  item.selected = this.inputInfo.attr.type === 'checkbox' ? this.dataValue[item.value] : (item.value === this.dataValue);
});
```

src/components/checkbox/Checkbox.js
```diff
@@ -182,6 +183,7 @@ export default class CheckBoxComponent extends Field {
   }
 
   updateValue(value, flags) {
+    this.setChecked();
```
travist, line 186: Here you are setting the checked state before the value is updated... you may want to do this after the ```super.updateValue``` call.

travist (changes requested):

travist:
This pull request broke our tests. Please ensure that all tests pass before submitting a pull request.

PR #3797, opened 2021-02-19
Added data to default values component evalContent

ccontour:


travist:
Tests are failing. Please investigate.

ccontour:
Merged develop and it went away, must have branched from a bad point. Looks fine now

PR #3800, opened 2021-02-22
Fix: fix rendering of nested wizards

DmitryNistratov:


src/Wizard.js
```diff
@@ -392,7 +392,7 @@ export default class Wizard extends Webform {
       const nestedPages = [];
       const currentComponents = nestedComp?.subForm ? this.getSortedComponents(nestedComp.subForm) : nestedComp?.components || [];
       const visibleComponents = currentComponents.filter(comp => comp._visible);
-      const additionalComponents = visibleComponents.filter(comp => !comp.subForm);
+      const additionalComponents = visibleComponents.filter(comp => !comp.subForm?.subWizards);
```
travist, line 395: Why is ```subWizards``` in the context of our core renderer? Shouldn't this be controlled by the Dynamic Wizard component to keep this code agnostic to that component?

travist (changes requested):

src/Wizard.js
DmitryNistratov: subWizards only relates to formio.js wizard. This code is needed to split the nested wizards and nested forms.

DmitryNistratov: But I remembered how else I can determine that this is a wizard and changed the condition

PR #3809, opened 2021-02-24
FIO-1422: Add 'Navigate Pages & Save on Enter' option for wizard

RomanLetsuk:


travist:
Please write some tests.

src/components/panel/editForm/Panel.edit.display.js
```diff
@@ -135,6 +135,16 @@ export default [
           next: true
         },
       },
+      {
+        weight: 55,
+        label: 'Navigate Pages & Save on Enter',
```
travist, line 140: We already have an option to "Press Enter to Submit"... We should just make this option "Navigate Wizard on Enter". If they wish to navigate and submit on Enter, then they should have both of those options checked.

travist (changes requested):

PR #3839, opened 2021-03-05
FIO-1207: fixed an issue where selected value always presents in drop…

TanyaGashtold:
…down with found search results in select resource


TanyaGashtold:
All tests passed locally

PR #3850, opened 2021-03-10
FIO-1317: Implements capability to use NextGen Rules Engine from premium in 4x.

MaksimFalei:


travist:
I would prefer to use the "hook" system instead of having special chunks of code included (using the isNextGen). Then the rules plugin can just implement the "hooks" to inject its logic where needed.

MaksimFalei:
> I would prefer to use the "hook" system instead of having special chunks of code included (using the isNextGen). Then the rules plugin can just implement the "hooks" to inject its logic where needed.

Fixed as mentioned.

travist:
Needs to be rebased.

MaksimFalei:
Solved conflicts

PR #3854, opened 2021-03-11
FIO-1488: Fixes an issue where unordered list created in Quill is rendered as an ordered one in read-only mode

alexandraRamanenka:
Styles fix, no changes needed

### DIFF
--- a/src/sass/formio.form.scss
+++ b/src/sass/formio.form.scss
@@ -1235,6 +1235,32 @@ div[data-oembed-url] {
   img {
     max-width: 100%;
   }
+
+  li[data-list="bullet"] {
+    list-style-type: none;
+
+    .ql-ui{
+      padding-right: 0.5rem;
+
+      &::before {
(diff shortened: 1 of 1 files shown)


src/sass/formio.form.scss
```diff
@@ -1235,6 +1235,32 @@ div[data-oembed-url] {
   img {
     max-width: 100%;
   }
+
+  li[data-list="bullet"] {
```
travist, line 1239: Isn't this CSS included in Quill?  This seems REALLY strange we have to include these styles in our renderer. The renderer should only be concerned with styles related to the rendering of forms.

travist (changes requested):

src/sass/formio.form.scss
alexandraRamanenka: In read-only mode we do not load quill or any other editor, moreover, we render the value using Well component, that's why styling differs

PR #3855, opened 2021-03-11
FJS-1272: Form crashes using underscore as a mask

AlehKatsiubaFormIO:


travist:
This kind of seems like a hack band-aid for the real issue. Is there not another way to solve the crashing issues?

PR #3863, opened 2021-03-15
fix values not set correctly

mikekotikov:


src/components/select/Select.js
```diff
@@ -1112,7 +1112,7 @@ export default class SelectComponent extends Field {
     }
     const notFoundValuesToAdd = [];
     const added = values.reduce((defaultAdded, value) => {
-      if (this.isEmpty(value)) {
+      if (!value || _.isEmpty(value)) {
```
travist, line 1115: We should instead be changing the Select component ```isEmpty``` method to return the lodash _.isEmpty check.

mikekotikov: The problem was that if false or 0 received in this check, there will be new select option created for this value and it will prevent setting the default value or submission right.
So, this condition is meant to pass for falsy values. And if I change to calling component's isEmpty method it won't pass.

PR #3867, opened 2021-03-15
create option to remove errors on new file upload

mikekotikov:


src/components/file/editForm/File.edit.file.js
```diff
@@ -170,6 +170,14 @@ export default [
     tooltip: 'This will allow using an attached camera to directly take a picture instead of uploading an existing file.',
     weight: 32
   },
+  {
```
travist, line 173: Shouldn't this be just the default behavior? Why are we adding a configuration for this?

mikekotikov: Okay, made this behavior default.

PR #3881, opened 2021-03-19
FIO-582: Updated way of setting the file links

RomanLetsuk:


travist:
Please write some tests.

PR #3882, opened 2021-03-19
FIO-1443: No dark overlay when PDF form is loading

AlehKatsiubaFormIO:


travist:
Why was this done in the "embed" css? This is purposefully fully transparent.

travist (changes requested):
This is not the place to make this change. Needs to be in another file other than the embed.scss file.

PR #3887, opened 2021-03-22
Fixed an issue where public config is not available for the form and …

TanyaGashtold:
…its components in builer mode


TanyaGashtold:
related to: https://github.com/formio/formio-server/pull/806

PR #3892, opened 2021-03-23
Added additional check for disabled property

AlexeyDavyd:
Copy of https://github.com/formio/formio.js/pull/3742/files for 4.12.7.x


travist:
Included in 4.13.x so closing.

PR #3898, opened 2021-03-24
FIO-1537: fixed submiting issue for wizards recaptcha component.

Maria-Golomb:


src/components/button/Button.js
```diff
@@ -470,11 +470,27 @@ export default class ButtonComponent extends Field {
     if (!this.root) {
       return;
     }
-    const recaptchaComponent = this.root.components.find((component) => {
-      return component.component.type === 'recaptcha' &&
-        component.component.eventType === 'buttonClick' &&
-        component.component.buttonKey === this.component.key;
-    });
+
+    const getRecaptchaComponent = (components, key) => {
```
travist, line 474: Should be using ```everyComponent``` here instead of this.

PR #3916, opened 2021-03-31
prevent deleting pdf component id

mikekotikov:


mikekotikov:
This one fixes the issue https://github.com/formio/formio.js/pull/3914

PR #3917, opened 2021-04-01
FIO-1269: dynamic wizard json template minifying fix

Maria-Golomb:


src/components/_classes/component/Component.js
```diff
@@ -717,7 +718,7 @@ export default class Component extends Element {
    * Returns the JSON schema for this component.
    */
   get schema() {
-    return fastCloneDeep(this.getModifiedSchema(_.omit(this.component, 'id'), this.defaultSchema));
+    return fastCloneDeep(this.getModifiedSchema(_.omit(this.component, 'id'), this.component.type==='dynamicWizard'? Components.components[this.component.type].builderInfo.schema : this.defaultSchema));
```
travist, line 721: This seems like a hack. Why doesn't the "defaultSchema" work for "dynamicWizard". Seems we should fix it there instead.

travist (changes requested):

Issue #3911, opened 2021-03-30
[BUG] Translations do not work for any languge when default language (en) translations are provided in Formio.createForm options.


llaurentiu:
FormioJs version: >=  4.13.rc7
Browser: any browser
--------------------------------------------------------------------------------- 
If we render a form that requires translations, the form starts with language = 'en' and the translations are provided in the Formio.createForm options, the translations will never work.

```
 Formio.createForm(document.getElementById('formio'), 'https://examples.form.io/translations', {
    readOnly: false,
    language: 'en',
    i18n: {
      'en': {
        // provide an initial translation for a label. This will never get translated on initial render
        //or language change for formiojs >= 4.13.rc7
        'First Name': 'THE FIRST NAME'
      }
    }
  })
```
Js fiddle here: https://jsfiddle.net/darkdante/a76u5svo/7/

The bug was introduced with this commit: https://github.com/formio/formio.js/commit/a405438058c4c59f5c74e0b9d43397b134040638 in the Component.js , the translate function : `if (text in i18nConfig.resources.en.translation && params._userInput) {return text; }`

I see what was the intention with `_userInput = true` parameter of the translate function, but I think that the assumption that the labels in the form definition are by default translated in english might not be correct plus it introduces side effects.

For instance we use placeholders instead of meaningful labels.

Translations are working fine with lower versions.

wag110894:
Hello @llaurentiu, 

Would you be able to create a PR to fix this issue?

llaurentiu:
Fixed in https://github.com/formio/formio.js/pull/3918 so closing it.

PR #3918, opened 2021-04-01
FIO-2668: fixed an issue where translation does not work for any language if en translation is provided for this value…

TanyaGashtold:


elibus:
@llaurentiu this should be linked to #3935

llaurentiu:
@elibus they are related but not the same issue. I disabled #3935 and created https://github.com/formio/formio.js/pull/3954 that is linked to a defect I created (https://github.com/formio/formio.js/issues/3955)

elibus:
@llaurentiu yep, you are right, it should be #3911 instead. Correct?

llaurentiu:
@elibus yes. didn't check that they created another defect for the same issue. I closed mine.

PR #3922, opened 2021-04-02
FJS | 4.13.0 Regression | Address: Address that has Modal Edit checked is not shown in error list. Error message doesn't disappear in modal window

yavorovsky:


yavorovsky:
Tests were rewritten according to new logic and fixes. No tests needed

src/components/tabs/Tabs.unit.js
```diff
@@ -15,7 +15,7 @@ describe('Tabs Component', () => {
       form.checkValidity(data, true, data);
 
       setTimeout(() => {
-        const openModalWrapper = form.element.querySelector('[ref="openModalWrapper"]');
+        const openModalWrapper = form.element.querySelector('[aria-label="modal-wrapper"]');
```
travist, line 18: why did we get rid of "refs" here?

travist (changes requested):
I am concerned that we had to change so much here to just fix "Address" component. Why do we need to change the base component class vs. just fix the Address component?

travist:
We need to find a less intrusive way to solve this problem.

src/components/tabs/Tabs.unit.js
yavorovsky: That's because I've changed the logic in a case when we have a parent component as a modal. I added the component's key of a modal's ref. In the test, I was not sure about the exact 'key' and just added as a selector 'aria-label'

yavorovsky:
> I am concerned that we had to change so much here to just fix "Address" component. Why do we need to change the base component class vs. just fix the Address component?

The problem is not only in 'Address' component. We couldn't validate undefined 'modal'. We used to set only 'modal' word as a ref but it's not triggering errors for any component that has modal as a parent.

PR #3924, opened 2021-04-05
FJS - 1294 Text Area - When Text area is empty but has a default row greater than 1 , text row shows as one line when submission is viewed

yavorovsky:


yavorovsky:
changed view styles for the textarea's submission view. I don't think it has to be tested by automation tests

src/components/textarea/TextArea.js
```diff
@@ -333,9 +333,28 @@ export default class TextAreaComponent extends TextFieldComponent {
 
   setReadOnlyValue(value, index) {
     index = index || 0;
+    const brake = '<br>';
```
travist, line 336: It should be "break" not "brake".

Also, this seems very heavy handed to solve this minor issue. There is not an easier way to solve this?

travist (changes requested):

travist:
We need a more simple way to solve this problem.

PR #3926, opened 2021-04-05
FIO-2667: Fixes performance regression issues comparing with formiojs 4.11.x /4.10.x

alexandraRamanenka:


test/renders/component-bootstrap-textarea-readOnly-value1.html
```diff
@@ -5,9 +5,9 @@
   <div ref="element">
     <div class="card card-body bg-light">
       <div ref="textArea">
-        <div ref="input" class="formio-editor-read-only-content"></div>
-      </div>
-    </div>
+        <div ref="input" class="formio-editor-read-only-content" '></div>
```
travist, line 8: Why is this injecting a quote mark?

travist (changes requested):

PR #3929, opened 2021-04-06
Fixes an issue where got an infinite loop when changing the default values of the Data Grid components in the Data Grid.

MaksimFalei:


src/WebformBuilder.js
```diff
@@ -1117,8 +1117,12 @@ export default class WebformBuilder extends Component {
         }
       }
       else {
-        this.preview._data[changed.instance._data.key] = changed.value;
-        this.webform._data[changed.instance._data.key] = changed.value;
+        const dataPath = _.get(defaultValueComponent, 'component.type') === 'datagrid'
```
travist, line 1120: Does this also not affect EditGrid? Seems strange to have a specific case for just datagrid here.

travist (changes requested):

src/WebformBuilder.js
MaksimFalei: No, because we don't have default value settings for the EditGrid

PR #3935, opened 2021-04-06
[Ignore] Public translation function in interpolation context

llaurentiu:
Add a translation function that is available in the interpolation context of every component.
This is useful when we want to translate strings/labels in component templates (especially edit grid where the column names are not translated), error messages etc.
Instead of using the internal function 't' by calling instance.t({text}),  we should have a function that indicates what it does.

Would be a good idea that the default edit grid template translate the column labels by calling this function. Currently they are not translated.


llaurentiu:
Duplicate of https://github.com/formio/formio.js/pull/3954

PR #3936, opened 2021-04-06
FJS | Conditionally showing columns are not auto-adjusting when auto adjust column is set to true

yavorovsky:


travist:
Please write some tests for this.

yavorovsky:
@travist done

Issue #3932, opened 2021-04-06
[BUG] setSelectionRange() for a number input causes a DOMException.


DaveBagler:
## Environment
Please provide as many details as you can:

- Hosting type
  - [ ] Form.io
  - [x] Local deployment
- Formio.js version: I'm using `4.12.7-patch.2`, but the problem still exists on the master branch.
- Frontend framework: VueJS
- Browsers: Chromium Edge 89.0.774.57, Chrome 89.0.4389.72, Firefox 86.0 

## Steps to Reproduce
1. Open the form in a browser that follows the spec for `setSelectionRange()` for number inputs.
2. Open the browser's console.
2. Do something on a form with a number field that triggers its focus call, for example clicking a error message for the field.
3. See the exception in console.

## Expected behavior

The browser should move the focus to the number field and scroll the field into view.

## Observed behavior

The browser does move the focus to the number field, but then it throws a DOMException because of the `setSelectionRange()` call on a number input field. The browser does not scroll the field into view in Edge/Chrome, but does in Firefox.

**Edge/Chrome:**  
<img width="715" alt="ChromeError" src="https://user-images.githubusercontent.com/3266985/113719565-be425100-96bb-11eb-8cb2-02c2f7ff7f1e.PNG">

**Firefox:**  
<img width="437" alt="FirefoxError" src="https://user-images.githubusercontent.com/3266985/113719936-22fdab80-96bc-11eb-8820-8f7b8b9d90a8.PNG">

## Fix

The relevant line of code is here: https://github.com/formio/formio.js/blob/master/src/components/number/Number.js#L181

I think the easiest solution that will provide the most consistent behavior is to remove the call to `setSelectionRange()`.

An alternative that keeps the functionality for browsers that don't follow the spec (for example IE11) could be leaving the call there but wrapping it in a try/catch block and simply swallowing the exception.

Tatsiana8:
Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #3941, opened 2021-04-07
#3932, remove setSelectionRange for number inputs

DaveBagler:
See this issue for more information: https://github.com/formio/formio.js/issues/3932

### DIFF
--- a/src/components/number/Number.js
+++ b/src/components/number/Number.js
@@ -178,6 +178,16 @@ export default class NumberComponent extends Input {
     const input = this.refs.input[0];
     if (input) {
       super.focus.call(this);
+      try {
+        input.setSelectionRange(0, input.value.length);
+      } catch(exception) {
+        if (exception instanceof DOMException === false) {
+          throw(exception);
+        }
+        /* Standard compliant web browsers don't support setSelectionRange()
+           on number input types and throw a DOMException. This try/catch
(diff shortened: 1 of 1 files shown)


src/components/number/Number.js
```diff
@@ -178,7 +178,6 @@ export default class NumberComponent extends Input {
     const input = this.refs.input[0];
     if (input) {
       super.focus.call(this);
-      input.setSelectionRange(0, input.value.length);
```
travist, line 181: We would prefer that you just add a try {} catch statement around this so it is still here but does not crash IE11.

travist (changes requested):

src/components/number/Number.js
DaveBagler: I can do that, but just to be clear, the issue isn't IE11 crashing. Older browsers seemed to run the method without issue, it's the newer browsers that are throwing a DOMException.

DaveBagler: See update in new commit.

travist:
I believe this PR will solve this problem. https://github.com/formio/formio.js/pull/3966

DaveBagler:
@travist, PR #3966 updates component.js and day.js. It doesn't remove the setSelectionRange() call from number.js.

DaveBagler:
@travist, ping on this PR.

PR #3948, opened 2021-04-08
FIO-2695: fix the display of a wizard inside components in which data…

DmitryNistratov:
… is stored as an array

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -393,7 +393,7 @@ export default class Wizard extends Webform {
     // Get all components including all nested components and line up in the correct order
     const getAllComponents = (nestedComp, compsArr, pushAllowed = true) => {
       const nestedPages = [];
-      const dataArrayComponents = ['datagrid', 'container', 'editgrid', 'dynamicWizard'];
+      const dataArrayComponents = ['datagrid', 'editgrid', 'dynamicWizard'];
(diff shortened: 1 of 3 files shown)


src/components/form/Form.js
```diff
@@ -261,8 +261,8 @@ export default class FormComponent extends Component {
           this.setContent(element, this.render());
           if (this.subForm) {
             this.subForm.attach(element);
-            if (this.subForm._form.display === 'wizard' && this.root._form.display === 'wizard') {
-              this.root.redraw();
+            if (this.subForm._form.display === 'wizard' && this.parent.parent?._form?.display === 'wizard') {
```
travist, line 264: this.parent.parent is super fragile and opinionated. What happens if this nested form is inside a panel that is inside a wizard?

travist (changes requested):

src/Wizard.js
```diff
@@ -393,12 +393,14 @@ export default class Wizard extends Webform {
     // Get all components including all nested components and line up in the correct order
     const getAllComponents = (nestedComp, compsArr, pushAllowed = true) => {
       const nestedPages = [];
+      const dataArrayComponents = ['datagrid', 'container', 'editgrid', 'dynamicWizard'];
```
travist, line 396: container does not have data stored as an array.

PR #3954, opened 2021-04-11
Fix edit grid not translated column labels. Bind translation function to any interpolated string context.

llaurentiu:
Fix edit grid not translated column labels. Bind translation function to any interpolated string.

### DIFF
--- a/src/components/editgrid/EditGrid.js
+++ b/src/components/editgrid/EditGrid.js
@@ -53,7 +53,7 @@ export default class EditGridComponent extends NestedArrayComponent {
     return `<div class="row">
       {% util.eachComponent(components, function(component) { %}
         {% if (displayValue(component)) { %}
-          <div class="col-sm-2">{{ component.label }}</div>
+          <div class="col-sm-2">{{ translate(component.label) }}</div>
         {% } %}
       {% }) %}
     </div>`;
--- a/src/components/editgrid/EditGrid.js
+++ b/src/components/editgrid/EditGrid.js
@@ -53,7 +53,7 @@ export default class EditGridComponent extends NestedArrayComponent {
     return `<div class="row">
       {% util.eachComponent(components, function(component) { %}
         {% if (displayValue(component)) { %}
-          <div class="col-sm-2">{{ translate(component.label) }}</div>
+          <div class="col-sm-2">{{ t(component.label) }}</div>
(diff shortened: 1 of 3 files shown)


elibus:
@wag110894 @heather-formio @travist  This is ECB. We would appreciate id this could be considered for 4.13.0

src/components/editgrid/EditGrid.js
```diff
@@ -53,7 +53,7 @@ export default class EditGridComponent extends NestedArrayComponent {
     return `<div class="row">
       {% util.eachComponent(components, function(component) { %}
         {% if (displayValue(component)) { %}
-          <div class="col-sm-2">{{ component.label }}</div>
```
travist, line 56: I would prefer that this is just the text "t".  Like so...

```
{{ t(component.label) }}
```

travist (changes requested):

src/components/editgrid/EditGrid.js
llaurentiu: @travist Isn't this the template that is prefilled in the text area of the edit grid header template (when we drag and drop the editGrid component on the form) ? I used `translate` because it has meaning for the user creating the form (user that might not know that much about formiojs internals)

 the `f` function is not passed to directly to the interpolation context (it is via the instance and it can be accessed like `instance.t`). I passed it in the `translate` also for the purpose of being meaningful when we draft a form and make use of interpolation (eg: you can use it to translate a custom error message).
 
 I can name it `t`. Let me know if you would still prefer to change it.

travist: I only prefer it to be consistent. All other variables are called the same thing... this would be the ONLY case where we change the variable name from "t" to "translate". I understand how it provides more meaning, but it would cause a surprise to someone who is already used to the evalContext variables to now have to know that "I use t() everywhere EXCEPT Edit Grid templates". I would just prefer to have it consistently be "t" everywhere. I vote to change it to "t"

llaurentiu: @travist agree with the consistency. I was making the differentiation between what the developer knows about the code and what the designer of the form sees in the template. For a portal user that is not technical and does not know the formio code, seeing there `t(component.label)` might not mean that much :) 

I made the changes and renamed the evalContext function to t.

PR #3960, opened 2021-04-12
FIO-2492: Fixes an issue when there is a component in one wizard page that has the same API key as the one on another wizard page where validation isn't met for some reasons, there will be misredirection from such an error in the Error list

MaksimFalei:


travist:
Please write a test.

MaksimFalei:
Added tests

PR #3978, opened 2021-04-14
FIO-2740: fix triggerChange call for cases when nested form component…

DmitryNistratov:
… changes

### DIFF
--- a/src/components/editgrid/EditGrid.js
+++ b/src/components/editgrid/EditGrid.js
@@ -833,7 +833,12 @@ export default class EditGridComponent extends NestedArrayComponent {
       options.name += `[${rowIndex}]`;
       options.row = `${rowIndex}-${colIndex}`;
       options.onChange = (flags = {}, changed, modified) => {
-        this.triggerRootChange({ ...flags, noValidate: true }, changed, modified);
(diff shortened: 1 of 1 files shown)


src/components/editgrid/EditGrid.js
```diff
@@ -833,7 +833,12 @@ export default class EditGridComponent extends NestedArrayComponent {
       options.name += `[${rowIndex}]`;
       options.row = `${rowIndex}-${colIndex}`;
       options.onChange = (flags = {}, changed, modified) => {
-        this.triggerRootChange({ ...flags, noValidate: true }, changed, modified);
+        if (changed.instance.root?.id && (this.root?.id !== changed.instance.root.id)) {
```
travist, line 836: Isn't this the same thing as ```this.triggerRootChange```?

travist (changes requested):

src/components/editgrid/EditGrid.js
DmitryNistratov: Not really. options.onChange is passed to all child components of the editGrid, but if there is a nested form inside EditGrid, then when calling onChange of a nested form component, the context will be EditGrid (root is the parent form), but first we need to call root.triggerChange of the nested form. I found this solution to be safe, but maybe I might have overlooked something.

PR #3980, opened 2021-04-15
FJS-1212 | FIO-175: Adds Display Mask option

alexandraRamanenka:
Also requires: https://github.com/formio/bootstrap3/pull/46

I formatted an Input template since it was pretty inconvenient to edit it. 
The only real change in the Input template is **lines # 25-27**


travist:
Please rebase with master branch.

PR #3982, opened 2021-04-16
FIO-1434: Fixes an issue when Default value settings did not work for Tree component. Removed  Multiple values settings for Tree component.

MaksimFalei:


travist:
Please fix conflicts.

PR #3983, opened 2021-04-16
fix submission not set on pdf for components with calculated value

mikekotikov:


mikekotikov:
tests pass at my local.

PR #3988, opened 2021-04-16
FIO-2747: change checking way for component pagination

DmitryNistratov:


DmitryNistratov:
Related to: [PR](https://github.com/formio/premium/pull/115)

src/Wizard.js
```diff
@@ -440,7 +440,8 @@ export default class Wizard extends Webform {
       getAllComponents(component, allComponents);
     }, []);
 
-    if (!this.root?.parent) {
+    // recalculate pages only for root wizards, including the situation when the wizard is in a wrapper
+    if (this.parent?.parent?.parent?._form?.display !== 'wizard') {
```
travist, line 444: This is so fragile. This is assuming that the root is 4 parents up.  We cannot make this assumption.

travist (changes requested):

PR #3997, opened 2021-04-20
Save & Delete Event for Edit Grid

edwinanciani:


rrg92:
Hi guys!

I noted that this change is not present in version published on https://cdn.form.io/formiojs/formio.full.min.js

There are some special reason?

In this version, only the rowIndex is passed.

PR #4009, opened 2021-04-22
FIO-2780: Makes Sanitize a configurable option

alexandraRamanenka:


travist:
We will need tests for this change.

travist:
Couple of questions.

 1.) If you do not provide the "sanitize" option, does it behave like it does today (and sanitize everything?)
 2.) If you provide sanitize "false" does it still sanitize the values? It should always sanitize the values even if sanitize option is false.

PR #4010, opened 2021-04-22
FIO-2622: Fixes an issue when you have a Nested Form component with clearOnHide set, the data becomes detached from the subForm component when the value is reset when cleared.

MaksimFalei:


travist:
Please write a test.

MaksimFalei:
Tests were added

PR #4020, opened 2021-04-26
FIO-1525: Fixes an issue where DataMap value is always shown as empty in the DataTable

alexandraRamanenka:


travist:
Please write a new test for this.

alexandraRamanenka:
> Please write a new test for this.

There is already a test. But it had an empty value to compare with before, so I just updated that value. I can add a new test, but it will do the same thing.

PR #4021, opened 2021-04-26
FIO-2538: Fixes an issue when Tree component has components with 'required' or 'unique' validation, and the Tree data entry is not saved, the validation is not stopping the form from being submitted

MaksimFalei:


travist:
Please write some tests for this.

PR #4041, opened 2021-04-28
FIO-2848, FIO-2831: Fixes some issues with setting error classes inside EditGrid

alexandraRamanenka:
Waiting for response from the customer to confirm how it should behave in few cases to finish this and fix tests

### DIFF
--- a/src/components/panel/Panel.js
+++ b/src/components/panel/Panel.js
@@ -33,20 +33,20 @@ export default class PanelComponent extends NestedComponent {
   get defaultSchema() {
     return PanelComponent.schema();
   }
-
-  checkValidity(data, dirty, row, silentCheck) {
-    if (!this.checkCondition(row, data)) {
-      this.setCustomValidity('');
-      return true;
-    }
-
-    return this.getComponents().reduce(
-      (check, comp) => {
-        return comp.checkValidity(data, dirty, row, silentCheck) && check;
-      },
-      super.checkValidity(data, dirty, row, silentCheck)
-    );
-  }
+  //
+  // checkValidity(data, dirty, row, silentCheck) {
+  //   if (!this.checkCondition(row, data)) {
+  //     this.setCustomValidity('');
+  //     return true;
+  //   }
+  //
+  //   return this.getComponents().reduce(
+  //     (check, comp) => {
+  //       return comp.checkValidity(data, dirty, row, silentCheck) && check;
+  //     },
+  //     super.checkValidity(data, dirty, row, silentCheck)
+  //   );
+  // }
 
   get templateName() {
     return 'panel';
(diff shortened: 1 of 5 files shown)


elibus:
It would be great if this could make it in the next 4.13.1 @heather-formio

alexandraRamanenka:
> Please write some tests.

Sure, I am just waiting for the response of the customer to confirm some aspects of the component's behavior to finish this PR.

src/components/panel/Panel.js
```diff
@@ -33,20 +33,20 @@ export default class PanelComponent extends NestedComponent {
...
-    }
-
-    return this.getComponents().reduce(
-      (check, comp) => {
-        return comp.checkValidity(data, dirty, row, silentCheck) && check;
-      },
-      super.checkValidity(data, dirty, row, silentCheck)
-    );
-  }
+  //
```
travist, line 36: I would rather just delete this code instead of comment it out. We can use GIT to see what was removed.

travist (changes requested):
Also needs tests.

kamsi:
@alexandraRamanenka thing is that I totally need to reflect that selection on a data level. My use case involves saving that choice for later for example. I guess that with your change we move the logic to the purely component render level and so the choice is not persisted, right (or maybe I do not get how editRow works exactly)? So, the way I see it it breaks my use case.

alexandraRamanenka:
So then, you need something like a hidden checkbox with 'isRowSelected' key and add events for handling when the row is selected/deselected. Using Advanced logic you can set the checkbox to change it value once it receives one of those events. 
Because "silently" changing the dataValue and adding/removing new keys to it is not we want to do, it could lead to very unexpected and undesired behavior (I have already found few cases when it totally breaks EditGrid)

kamsi:
@alexandraRamanenka OK, thanks for the guidance, I will look into event based handling with advanced logic + hidden checkbox. And thank you for notifying me explicitly in the first place about the change.

alexandraRamanenka:
> @alexandraRamanenka OK, thanks for the guidance, I will look into event based handling with advanced logic + hidden checkbox. And thank you for notifying me explicitly in the first place about the change.

No problem. If you have any questions or need an example, you can ask me.

elibus:
Thanks @alexandraRamanenka 
@travist can we have a rc with this please?

heather-formio:
@elibus This pull request has been merged and is part of Formio.js version 4.14.0-RC.6. Let us know if you have any questions.

elibus:
Thanks @heather-formio we saw also rc.6 includes the other patch in 4.13, it seems to work fine. Thanks to @travist and @alexandraRamanenka as well 👍

PR #4047, opened 2021-04-30
Inherit default (en) translations (validations, buttons) for custom languages

llaurentiu:
Right now, if the validation labels (eg: required, invalid_email etc.) are not in the dictionary of translated labels, the error label itself is shown if the field is invalid.

This PR adds formio default translations to every language used in the form.

If errors are in the translations dictionary provided to the form, they take precedence over the defaults.

This is useful in case we consider 'en' the system language (and therefore error messages are system messages) and we only want to translate the form content.


elibus:
@heather-formio  this is ECB

PR #4051, opened 2021-05-04
FIO-2784: Fixes an issue where labels that contain HTML tags are not rendered properly within the errors alert

alexandraRamanenka:


alexandraRamanenka:
https://github.com/formio/formio.js/pull/4060

PR #4054, opened 2021-05-05
FIO-2936: replace finally method for better compatibility

DmitryNistratov:


src/Webform.js
```diff
@@ -1526,8 +1526,11 @@ export default class Webform extends NestedDataComponent {
     this.submitting = true;
     return this.submitForm(options)
       .then(({ submission, saved }) => this.onSubmit(submission, saved))
-      .catch((err) => NativePromise.reject(this.onSubmissionError(err)))
-      .finally(() => this.submissionInProcess = false);
+      .then(() => this.submissionInProcess = false)
```
travist, line 1529: Won't this change the response from the promise?  Instead of returning the results of "onSubmit" it will now return "false".  I am sure this will break the renderer.

travist (changes requested):

PR #4062, opened 2021-05-06
FIO-2554: add properties to the default schemas

DmitryNistratov:
@travist FYU I changed the getModifiedSchema method so that it works also for objects inside arrays. Therefore, now the columns field in the Columns component is also removed from the minified schema, if the initial value of the field does not change


travist:
Please write tests.

DmitryNistratov:
@travist, added tests

PR #4075, opened 2021-05-14
FIO-2989: Fixes an issue when for the oAuth redirect URI as host that trigger conflicts with OpenID,

MaksimFalei:


MaksimFalei:
Related PR: https://github.com/formio/formio-server/pull/836

PR #4086, opened 2021-05-21
FIO-2503: Feat: adds Truncate Multiple Spaces option

alexandraRamanenka:


src/components/textfield/editForm/TextField.edit.data.js
```diff
@@ -46,4 +46,11 @@ export default [
       }
     ]
   },
+  {
+    weight: 205,
```
travist, line 50: This also needs to be added to the default schema or it will show up in the minified schema.

travist (changes requested):

PR #4108, opened 2021-05-31
Call onChange on resetValue in Webform

juliapath:
#3666 made `resetValue` call `rebuild` in `Webform.js`, however this is still not enough as a rebuild does not cause elements with calculated values to be recalculated. In the worst case if the user clicks on a reset button and then immediately submits the form this will result in fields which are always supposed to be calculated being submitted uncalculated. In other cases this may still leave the form in an inconsistent state after a reset. Introducing a `onChange` here fixes this and intuitively makes sense, since a reset does indeed cause a change in the form.

It should be noted however, that `resetValue` in `Component.js` explicitly does not cause a change event:

https://github.com/formio/formio.js/blob/f6d2e2177992b6e6ee8561ea75c7efc8b3f6416e/src/components/_classes/component/Component.js#L2564-L2574

The purpose of this is not entirely clear to me, but I'm assuming it is to prevent a change event being fired for every single component in the form. If that's the case this PR would not contradict that purpose as it only causes one change event at the end.

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -1051,7 +1051,7 @@ export default class Webform extends NestedDataComponent {
   resetValue() {
     _.each(this.getComponents(), (comp) => (comp.resetValue()));
     this.setPristine(true);
-    this.rebuild();
+    this.onChange();
   }
 
   /**


juliapath:
Sorry for the indecisiveness, as to `onChange` vs `triggerChange`.

travist:
This looks good to me. Thanks!

PR #4120, opened 2021-06-05
Fix HTML renderMode for the SelectBoxes component

zoafile:
Looks like the fix for #4079 can also be applied to the SelectBoxes component.


travist:
Please rebase webform.unit.js and we will accept this pull request. Thanks!

zoafile:
@travist sounds good, rebased just now

travist:
Thanks for your effort @zoafile but I believe this was resolved with https://github.com/formio/formio.js/pull/4203

PR #4124, opened 2021-06-07
FIO-3143: fix overriding the value with the label of Select/Select Boxes/Radio components

RamanHryhohyk:


RamanHryhohyk:
was fixed in https://github.com/formio/formio.js/pull/4133

PR #4131, opened 2021-06-08
Add aria-invalid attribute, improve test ergonomics

shawnbot:
This PR improves the accessibility of _most_ form inputs by updating the `aria-invalid` attribute to `true` or `false` whenever the `is-invalid` class is added or removed. This called for a new Component method, `setElementInvalid()`, which adds or removes the `is-invalid` class and sets `aria-invalid` in one go.

The only mention of the `is-invalid` class I could find in the tests was for the calendar widget, so I've updated those tests to also assert the correct value of `aria-invalid`. Ideally there would be invalid state tests for every component, but I think it's safe to say that if `is-invalid` is working as a style hook for invalid elements (and I've fixed this in all of the right places) all of the cases should be covered.

I've also introduced a Mocha config file in `.mocharc.js` and moved the CLI options from `package.json` into it so that it's easier to run one-off tests with `npx mocha lib/some/file.unit.js`. Eslint ignores files that begin with `.`, so I've negated it from the eslint config's `ignorePatterns` with `!.*rc.js`. I'm happy to break this out into a separate PR if you'd rather review something smaller. 🍻

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,6 +11,9 @@
     "node": true,
     "mocha": true
   },
+  "ignorePatterns": [
+    "!.*rc.js"
+  ],
   "rules": {
     "max-len": [
       "off",
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,0 +1,10 @@
+process.env.TZ = 'UTC';
+
+module.exports = {
+  bail: true,
+  timeout: 20000,
+  require: [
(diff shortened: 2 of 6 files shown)


travist:
As always, this looks good @shawnbot. We will get this into one of the 4.14.x releases. Most likely will be the 4.14.1 release because we are trying to button up the 4.14.0 release now.

PR #4143, opened 2021-06-15
FIO-560: Fixes an issue where CKEditor media is not shown in the view mode

alexandraRamanenka:


src/components/_classes/component/Component.js
```diff
@@ -928,12 +928,16 @@ export default class Component extends Element {
...
+  sanitize(dirty, options) {
     // No need to sanitize when generating PDF'S since no users interact with the form.
     if (this.options.pdf) {
       return dirty;
     }
-    return FormioUtils.sanitize(dirty, this.options);
+    return FormioUtils.sanitize(
+      dirty,
+      {
+        sanitizeConfig: _.merge(this.options?.sanitizeConfig || {}, options || {}),
```
travist, line 939: This will mutate the global this.options with the configuration for a specific component. We should change it to this instead.

```
sanitizeConfig: _.merge({}, this.options?.sanitizeConfig || {}, options || {}),
```

travist (changes requested):

src/components/_classes/component/Component.js
travist: Actually, reading the docs makes me believe that all options to the left of other options will get overridden.

https://lodash.com/docs/4.17.15#merge

Maybe we should fastCloneDeep the sanitizeConfig?

PR #4146, opened 2021-06-15
Add aria labels to edit/remove row buttons.

azizk:
Add aria-label attributes to make validators happy.

I'm not sure if this is the only place where the template needs to be modified.
Also is interpolating with `${this.t('Edit row')}` correct, or should it rather be `{{ t('Edit row') }}`?


travist:
Thanks!

PR #4151, opened 2021-06-17
FIO-3199: fix triggering time component custom validation

RamanHryhohyk:


src/validator/Validator.js
```diff
@@ -683,6 +683,11 @@ class ValidationChecker {
           if (!setting) {
             return true;
           }
+
+          if (component.type === 'time') {
```
travist, line 687: I would rather figure out why the time component is not sending the updated value to the validator instead of hacking the validation routine.

mikekotikov:
Will redo this

PR #4152, opened 2021-06-17
FIO-3109 Implemented feature to set focus on new row when adding one …

antonSoftensity:
…in Data Grid


AlexeyDavyd (approved):

src/components/datagrid/DataGrid.js
```diff
@@ -409,6 +409,17 @@ export default class DataGridComponent extends NestedArrayComponent {
     this.rebuild();
   }
 
+  focusOnNewRowElement(row) {
+    Object.keys(row).find((key) => {
+      const focusableElements = getFocusableElements(row[key].element);
```
travist, line 414: The inputs should be within a "refs.input" variable so this is not necessary and way less performant than it should be.

travist (changes requested):

PR #4158, opened 2021-06-22
Fixes an issue where the change event for the first character in Textarea with wysiwyg is triggered without modified flag

alexandraRamanenka:


travist:
I cherry-picked this into the master branch. If we need a special build with this, then we can create a ticket to cherry-pick this commit into that other branch, but for now, this exists in the 4.14.x release.

PR #4162, opened 2021-06-23
FIO-2688: fix component layout setting 

RamanHryhohyk:
When the Page setting is changed, the changes do not display until saving the form

### DIFF
--- a/src/PDFBuilder.js
+++ b/src/PDFBuilder.js
@@ -186,6 +186,7 @@ export default class PDFBuilder extends WebformBuilder {
   afterAttach() {
     this.on('saveComponent', (component) => {
       this.webform.postMessage({ name: 'updateElement', data: component });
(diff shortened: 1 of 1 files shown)


src/PDFBuilder.js
```diff
@@ -186,6 +186,7 @@ export default class PDFBuilder extends WebformBuilder {
   afterAttach() {
     this.on('saveComponent', (component) => {
       this.webform.postMessage({ name: 'updateElement', data: component });
+      this.webform.postMessage({ name: 'form', data: this.webform.form });
```
travist, line 189: This is a very drastic change to fix a such a small issue. This makes me believe that all components have this problem. I would rather determine why component layout setting is not setting correctly without having to reset the whole form object. I am sure this will cause some other regressions, so this needs to be revisited.

travist (changes requested):
Please resolve comments.

PR #4170, opened 2021-06-25
FIO-3239: fix displaying custom validation on the inital component with default value

RamanHryhohyk:


src/components/_classes/component/Component.js
```diff
@@ -2804,7 +2804,7 @@ export default class Component extends Element {
 
   setComponentValidity(messages, dirty, silentCheck) {
     const hasErrors = !!messages.filter(message => message.level === 'error' && !message.fromServer).length;
-    if (messages.length && (!silentCheck || this.error) && (dirty || !this.pristine)) {
+    if (messages.length && (!silentCheck || this.error) && (!this.isEmpty(this.defaultValue) || dirty || !this.pristine)) {
```
travist, line 2807: This seems like the wrong place to have this... this is saying that components cannot have custom validity if they have a default value set. This just seems strange and not correct. I don't believe this is the fix in the right place.

travist (changes requested):

src/components/_classes/component/Component.js
antonSoftensity: @travist this is the change I mentioned on dev support today. Do you still think that it should not be fixed there?

PR #4174, opened 2021-06-29
FIO-1310: fixed an issue where select search not always shows all available options

TanyaGashtold:
…

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -38,7 +38,7 @@ export default class SelectComponent extends Field {
       ignoreCache: false,
       template: '<span>{{ item.label }}</span>',
       selectFields: '',
-      searchThreshold: 0.3,
+      selectThreshold: 0.3,
       uniqueOptions: false,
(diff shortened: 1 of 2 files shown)


src/components/select/Select.js
```diff
@@ -855,13 +855,16 @@ export default class SelectComponent extends Field {
...
+        ? {
+          tokenize: true,
+          matchAllTokens: true,
+        }
         : Object.assign(
         {},
         _.get(this, 'component.fuseOptions', {}),
         {
           include: 'score',
-          threshold: _.get(this, 'component.searchThreshold', 0.3),
```
travist, line 864: So, the configuration has always been "selectThreshold" but we use "searchThreshold" in the code?

TanyaGashtold: yes, that is true, the key of the 'search threshold' option in our builder has always been "selectThreshold", but in the code we used "searchThreshold" instead and this option did not work.

PR #4175, opened 2021-06-29
WIP Fix/UIP 295 datagrid logic loop

mikekotikov:


travist:
Wouldn't this break value logic statements on Hidden component types?

src/components/_classes/component/Component.js
```diff
@@ -1871,7 +1871,7 @@ export default class Component extends Element {
             'value',
           );
 
-          if (!_.isEqual(oldValue, newValue)) {
```
travist, line 1874: This does not make sense that this would only set the values if the component is visible. What is the bug this is trying to resolve? This seems like it is not a resolution of the core issue but just a bandaid on top of it.

travist (changes requested):

src/components/_classes/component/Component.js
mikekotikov: I just wasn't been able to understand any use case for changing the value of a hidden component in logic. It's neither present in the submission or data, nor the dataValue for this component is changing.
And about the root of the problem you are right, it appeared after [this PR](https://github.com/formio/formio.js/pull/3655) and I tried some possible fixes within those lines, but the tests failed, so I stopped on this.
If there is some application of the ability to set values in logic for hidden components, I can continue investigating on other solution.

travist: Yes, we will need to further investigate this issue to find root cause. We should be able to apply logic changes to hidden components without issue.

PR #4188, opened 2021-07-07
fix for tests: updated wizard renders

TanyaGashtold:


travist:
I will do this when I build it.

PR #4194, opened 2021-07-08
Fix html render mode for the Day component

zoafile:
This fix provides an html template so that the Day component shows the individual date part values vs rendering the full date multiple times. I hope I'm not misunderstanding the way the Day component should work, let me know if that is the case.

**Current html rendering**
image

**Rendering with this fix**
image


travist:
I appreciate your efforts @zoafile but we fixed this issue with https://github.com/formio/formio.js/pull/4207

PR #4195, opened 2021-07-09
Add margin bottom to all components in builder mode

AlexeyDavyd:


travist:
Wouldn't this make the builder form not resemble the rendered form? I am very cautious about accepting this due to making the builder form look much different. What are some before / after screenshots of this change?

AlexeyDavyd:
@travist 

Here's an example.
Before:
image

After:
image

PR #4199, opened 2021-07-12
FIO-3227: added an option that allows to handle asynchronous code in select custom values

TanyaGashtold:


src/components/select/editForm/Select.edit.data.js
```diff
@@ -434,6 +454,18 @@ export default [
       },
     },
   },
+  {
+    key: 'data.async',
```
travist, line 458: Why does this need to be an option?... Why can't we just look at the object that is returned from the execution and if a Promise is returned, then it becomes "async".

travist (changes requested):

PR #4210, opened 2021-07-16
Improve datetime expected flatpickr locale: Convert possible IETF language tag to 2 chars.

bobslee:
Examples (from IETF to expected flatpicker locale):
- en-US: en
- en-GB: en
- zh-CN: zh
- zh-HK: zh


bobslee:
This still needs to be improved.

First check whether the locale exists in `flatpickr.l10ns`.
If not exist, then apply yhe `substr()`.

travist (changes requested):
Thanks @bobslee I agree, there still needs to be some work with this. I will set it as "Request Changes" until you make those improvements.

bobslee:
@travist Is the `flatpickr` module/object present in global scope or in the component?

travist:
We use ```Formio.requireLibrary``` to bring in the Flatpickr dependency. This is a lazy-loading method which will first check for the global flatpickr and if not found dynamically load it and inject it into the CalendarWidget.

bobslee:
I shall proceed this in a different (specific) branch.

PR #4212, opened 2021-07-16
Add option for authorization header in s3 requests

VApostolovIntracol:


travist:
Looks good to me! Thanks!

Issue #4213, opened 2021-07-16
[BUG] Components inside layout components create elements with non-unique id


airarrazaval:
## Environment
Please provide as many details as you can:

- Hosting type
  - [x] Form.io
  - [ ] Local deployment
    - Version: 
- Formio.js version: `>= 4.12.0-rc.1`
- Frontend framework: `any`
- Browser: `any`
- Browser version: `any`

## Steps to Reproduce
1. Create a form with a Data Grid
2. Insert a Columns component inside the Data Grid
3. Insert any input component inside the Columns component
4. Render the form
5. Add a second row

## Expected behavior
All components should be created with a unique HTML id.

## Observed behavior
Components inside a layout component (which in turn is inside a Data Grid, Edit Grid, etc) retain their unique id and are not updated with the parent Nested Component's row index as it should.

![ezgif com-gif-maker (2)](https://user-images.githubusercontent.com/9391489/125981584-d5fe6b56-2f3b-492a-80e8-a4f3871d0440.gif)

This causes the browser to trigger a warning for detecting non-unique HTML ids.

The bug was introduced in `formiojs@4.12.0-rc.1` and is still present.

## Example

https://jsfiddle.net/airarrazaval/b10pda43/

airarrazaval:
@wag110894  

Related to issue https://github.com/formio/formio.js/issues/3973

airarrazaval:
Bug introduced in PR https://github.com/formio/formio.js/pull/3076. Possible solution would require layout components to check if they are inside a row, and if so, pass on that to their child components so they can build up a unique id.

airarrazaval:
Created PR https://github.com/formio/formio.js/pull/4214.

The issue was caused by NestedComponents not making shallow copies of its components.  This caused the NestedComponent components to overwrite their component id when the NestedComponent was rendered inside a NestedArrayComponent (eg: TextfieldComponent inside a ColumnsComponent inside a DataGridComponent).

jtamm-red:
I try fix it.

I made my PR: https://github.com/formio/formio.js/pull/4281

olgabann:
We're currently addressing a backlog of GitHub issues. Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #4214, opened 2021-07-16
Fixes inputs non-unique id

airarrazaval:
This is a fix for issue https://github.com/formio/formio.js/issues/4213.

### DIFF
--- a/src/components/_classes/nested/NestedComponent.js
+++ b/src/components/_classes/nested/NestedComponent.js
@@ -326,7 +326,7 @@ export default class NestedComponent extends Field {
     if (!this.isInputComponent && this.component.shouldIncludeSubFormPath) {
       component.shouldIncludeSubFormPath = true;
     }
-    const comp = Components.create(component, options, data, true);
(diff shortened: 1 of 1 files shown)


airarrazaval:
Another solution would be to force shallow copies (or even fastCloneDeep) at Components.createComponent in order to prevent future mutation problems, although I don't know what would be the impact in performance (specially for fastCloneDeep). 

You could avoid unnecessary cloning by performing the following check:

```js
let schema;

if (!!component.columns && !component.components) {
  schema = fastCloneDeep(schema);
} else {
  schema = { ...component };
}

return Components.createComponent(schema, ...);
```

I imagine the only case where deep cloning would be necessary is if somewhere there are mutations for object properties (such as validations, etc), which I think that is not yet implemented, but nevertheless is good to have it under consideration.

travist:
@airarrazaval I had to revert this because it was causing test failures. Please resolve these failures and resubmit the pull request.

airarrazaval:
Ok I will have a look at it.

Issue #4164, opened 2021-06-23
[BUG] FormBuilder: misspelled option "isUseWorkerDisabled"


mm-dsibinski:
## Environment
Please provide as many details as you can:

- Hosting type
  - [] Form.io
  - [x] Local deployment
    - Version: n/a 
- Formio.js version: 4.13.2
- Frontend framework: react
- Browser: Google Chrome
- Browser version: Version 91.0.4472.114 (Official Build) (64-bit)

## Steps to Reproduce
1. Go to FomBuilder demo page: https://formio.github.io/formio.js/app/builder
2. Open Developer Console (F12)
3. Drag and drop a `Checkbox` control to the form
4. See the error in the console:
```
ace.js:1 misspelled option "isUseWorkerDisabled"
o   @   ace.js:1
setOption   @   ace.js:1
(anonymous) @   ace.js:1
setOptions  @   ace.js:1
eval    @   Component.js:2381
Promise.then (async)        
addAce  @   Component.js:2378
```

## Expected behavior
No error in the console

## Observed behavior
Error in the console

upexcel:
I am facing similar issue with angular.

"angular-formio": "^4.4.5",
"formiojs": "^4.13.2",

These issues are blocker - (Missing project Id and isUseWorkerDisabled).
My angular app slows down firefox because of these issues.

airarrazaval:
This is an issue caused by a custom setting used by form.io which is passed to the Ace Editor.  Checking the Ace Editor repo you can see that if an unknown key is passed to the editors. settings object, the warning message is displayed.

https://github.com/ajaxorg/ace/blob/94422a4a892495564c56089af85019a8f8f24673/lib/ace/lib/app_config.js#L74-L82

You can check form.io's implementation here:

https://github.com/formio/formio.js/blob/a56cc08744ece4f8c1e443c01a53d07475086192/src/components/_classes/component/Component.js#L2248-L2260

Created PR https://github.com/formio/formio.js/pull/4215 which fixes the warning message.

olgabann:
We're currently addressing a backlog of GitHub issues. Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

ParhamShD:
The issue is still present. Same steps and expected behavior as in the original ticket:
> ## Steps to Reproduce
> 1. Go to FomBuilder demo page: https://formio.github.io/formio.js/app/builder
> 2. Open Developer Console (F12)
[...]

PR #4215, opened 2021-07-16
Fixed Ace Editor unknown option warning

airarrazaval:
Fix for issue https://github.com/formio/formio.js/issues/4164

### DIFF
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -2249,7 +2249,7 @@ export default class Component extends Element {
       .then((editor) => {
         editor = editor.edit(element);
         editor.removeAllListeners('change');
-        editor.setOptions(settings);
+        editor.setOptions(_.omit(settings, 'isUseWorkerDisabled'));
         editor.getSession().setMode(settings.mode);
         editor.on('change', () => onChange(editor.getValue()));
         if (settings.isUseWorkerDisabled) {
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -2117,7 +2117,56 @@ export default class Component extends Element {
         minLines: 12,
         tabSize: 2,
         mode: 'ace/mode/javascript',
-        placeholder: this.t(this.component.placeholder, { _userInput: true })
+        placeholder: this.t(this.component.placeholder, { _userInput: true }),
+        fontFamily: 'Consolas, Menlo, Monaco, "Courier New", monospace',
+        // default editor options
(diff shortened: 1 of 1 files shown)


src/components/_classes/component/Component.js
```diff
@@ -2249,7 +2249,7 @@ export default class Component extends Element {
       .then((editor) => {
         editor = editor.edit(element);
         editor.removeAllListeners('change');
-        editor.setOptions(settings);
+        editor.setOptions(_.omit(settings, 'isUseWorkerDisabled'));
```
travist, line 2252: I would rather use ```_.pick``` so we won't have to keep introducing new "omits" as we add other settings elsewhere.

travist (changes requested):

airarrazaval:
@travist that's a pretty big list:

https://github.com/ajaxorg/ace/wiki/Configuring-Ace

Some options are missing but are defined in the [types declaration file](https://github.com/securingsincity/react-ace/blob/main/types.d.ts#L41-L98) and in their respectove controller files (see below).

I will set the default options in the `wysiwygDefault.ace` object and then do something like 

```js
editor.setOptions(_.pick(settings, Object.keys(this.wysiwygDefault.ace)));
```

I took the default values from the following ace options definitions:

* [Editor Options](https://github.com/ajaxorg/ace/blob/35ddb50f353f14ba56872b5921d95d81e1ad7f08/lib/ace/editor.js#L2861)
* [Renderer Options](https://github.com/ajaxorg/ace/blob/35ddb50f353f14ba56872b5921d95d81e1ad7f08/lib/ace/virtual_renderer.js#L1742)
* [MouseHandler Options](https://github.com/ajaxorg/ace/blob/145328b96083c6b7c306d3dcf841948f73486919/lib/ace/mouse/mouse_handler.js#L214)
* [Session Options](https://github.com/ajaxorg/ace/blob/145328b96083c6b7c306d3dcf841948f73486919/lib/ace/edit_session.js#L2460)

travist:
Oh wow... I didn't realize how big this list was. Let me discuss this internally and see which approach we like better. Thanks for your efforts on this.

travist:
So don't hate me, but I think I like the other way better now. This does not really save us any more maintenance and arguably would add more (where we would need to add a new property as they are added to ace). I would recommend reverting it back to the way it was and we will commit it.

heather-formio:
PR is out of date. Closing, please re-open new PR if still needed. Thank you for your contribution.

PR #4223, opened 2021-07-20
FIO-3465: Errors embedding FJS 4.14.0-rc.13 inside application

AlehKatsiubaFormIO:


travist:
Fixed with https://github.com/formio/formio.js/pull/4217

Issue #4048, opened 2021-04-30
[BUG] FormBuilder error: Uncaught (in promise) Missing projectId


mm-dsibinski:
## Environment
Please provide as many details as you can:

- Hosting type
  - [] Form.io
  - [x ] Local deployment
    - Version: N/A, custom backend
- Formio.js version: 4.13
- Frontend framework: React or vanilla JavaScript
- Browser: Google Chrome
- Browser version: Version 90.0.4430.93 (Official Build) (64-bit)

## Steps to Reproduce
1. Go to https://formio.github.io/formio.js/app/builder
2.  Open Google Chrome console

## Expected behavior
No error in the console

## Observed behavior
Error in the console:
```
builder:1 Uncaught (in promise) Missing projectId
Promise.then (async)
WebformBuilder @ WebformBuilder.js:382
create @ FormBuilder.js:82
setForm @ Form.js:253
Form @ Form.js:122
_createSuperInternal @ FormBuilder.js:50
FormBuilder @ FormBuilder.js:72
(anonymous) @ builder:137
```

We are getting the same error when rendering `FormBuilder` from FormIO 4.13 in our app.
Before 4.13 version it was ok.

## Example
Example is already present in the official FormBuilder demo page: https://formio.github.io/formio.js/app/builder

Is there any workaround for this error to not appear? Maybe some option to set, so the `projectId` is not missing anymore?

dwaynedsouza:
I have the same issue. Any fix?

nazmiwafiyy:
Same issue here.

ErosII:
Mee too

mm-dsibinski:
Hi guys, any info here? @travist , @wag110894 ?

upexcel:
I am facing similar issue with angular.

> "angular-formio": "^4.4.5",
> "formiojs": "^4.13.2",

`zone.js:690 Unhandled Promise rejection: Missing projectId ; Zone: <root> ; Task: Promise.then ; Value: Missing projectId undefined`

can anyone help please?

wag110894:
Apologies for the delay. 

We have created a ticket FIO-3391. We will let you know once this issue has been resolved.

sehsanm:
Any updates regarding this bug?  It is blocker issue

ensemblebd:
[...]

PR #4227, opened 2021-07-23
Fixed builder missing projectId promise rejection

airarrazaval:
Fix for issue https://github.com/formio/formio.js/issues/4048

### DIFF
--- a/src/WebformBuilder.js
+++ b/src/WebformBuilder.js
@@ -229,18 +229,20 @@ export default class WebformBuilder extends Component {
     const isResourcesDisabled = this.options.builder && this.options.builder.resource === false;
 
     formio.loadProject().then((project) => {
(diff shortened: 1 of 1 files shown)


src/WebformBuilder.js
```diff
@@ -229,18 +229,20 @@ export default class WebformBuilder extends Component {
...
-          _.set(this, pathToFormConfig, config);
+          if (this.webform  && !webformConfig) {
+            _.set(this, pathToFormConfig, config);
+          }
         }
-      }
-    });
+      }).catch((err) => {
+        console.warn(`Could not load project settings: ${err.message || err}`);
+      });
```
0liver, line 245: @airarrazaval Thanks for fixing this annoyance! I think you should restore the previous indentation level, though 😉

airarrazaval: fixed.

Issue #4225, opened 2021-07-22
[BUG] textareas with inputFormat plain is rendering HTML


ddurham2:
## Environment
Please provide as many details as you can:

- Hosting type
  - [ ] Form.io
  - [x] Local deployment
    - Version: 5996593db6dbe7bd4f2929f94cc987500769de96
- Formio.js version: 4.13.4
- Frontend framework: Angular-12
- Browser: chromium
- Browser version: v91

## Steps to Reproduce
1. create a form with a textarea with input format set to plain
2. create an object, add add some html to the text area (e.g.` <b> ... </b>`)
3. view the object with formio-js in view-only mode

## Expected behavior
With input format set to plain, I would expect to see the html tags rendered literally and not interpretted

## Observed behavior
The content of the view-only element is rendering interpreted html (e.g. `<b>` renders bold text).
This is concerning as it would possibly indicate that untrusted users who can enter data that is viewed by other users could be subject to scripting attacks.  I observed also that it will render <button> tags as buttons.  Though, I could not immediately get it to execute js.  Nonetheless, it should be a plaintext rendering.

## Example
![image](https://user-images.githubusercontent.com/4805134/126587432-286cc2db-bead-492f-a50d-ec54586e330b.png)
![image](https://user-images.githubusercontent.com/4805134/126586684-20fba57c-4747-42d7-a387-90c2b2157aa4.png)

Where I think this is occurring: 
   -> https://github.com/formio/formio.js/blob/master/src/components/textarea/TextArea.js#L67
     -> https://github.com/formio/formio.js/blob/master/src/templates/bootstrap/well/form.ejs

Is it trying to safely render html, but stripped of potentially harmful elements?  What package is doing that?  I've known that to be risky business.  But still, plain, I would think, should render plainly.

TIA

airarrazaval:
@ddurham2 unfortunately the client that gets downloaded by default with the formio API server (https://github.com/formio/formio-app-formio) is very outdated and uses an old version of the formio.js SDK (if I'm not wrong it is formiojs 2.x or 3.x).  This issue is not present in the latest stable version of formiojs.  In fact you can try the same in https://portal.form.io and it won't happen.

There is a bigger flaw with the `formio-app-formio` client.  You can trigger a custom javascript using the `onload` event in images (for example). 

1. Create a new form using the `formio-app-formio` client
2. Add a TextArea component with `plain` text data type and `tableView = true`
3. Create a new submission with the following value for the TextArea component

```html
[...]

PR #4233, opened 2021-07-26
Fixed HTML being rendered when inputFormat is plain

airarrazaval:
Fix for issue https://github.com/formio/formio.js/issues/4225


ddurham2:
FTR, I approve.  Thanks @airarrazaval !

travist:
@airarrazaval I had to revert this because it was causing a test failure. Please run tests and fix the failures and resubmit when they are fixed.

PR #4234, opened 2021-07-28
FIO-3435: form definition downloading according query parameter

Maria-Golomb:


src/Formio.js
```diff
@@ -325,37 +331,7 @@ class Formio {
   }
 
   loadForm(query, opts) {
-    return this.load('form', query, opts)
```
travist, line 328: This looks like it will break reverse compatibility where it would no longer work if they upgrade renderer pointed to a older server. Is there a way we can do this so that it will not break reverse compatibility?  Maybe go ahead and fetch the form with the URL, and then if it returns the correct schema, then don't load the other requests?

travist (changes requested):

PR #4242, opened 2021-08-02
FIO-3562: separated timezone logic to be able to redefine it in typescript

TanyaGashtold:


TanyaGashtold:
related PR in formio-pdf: https://github.com/formio/formio-pdf/pull/116

PR #4246, opened 2021-08-03
FIO-3488: fixed an issue where values are not recalculated when submission is being set in edit mode

TanyaGashtold:


travist:
I would like to get @alexandraRamanenka input on this since she was the one who did this commit https://github.com/formio/formio.js/commit/509039524af05a11a572aee4324274301392528d which was just a couple of months ago. I am fearful that this would break the case she was trying to fix.

alexandraRamanenka:
@travist no, it is okay, my fix was for values which were manually overridden. And this case still works with her fix.

PR #4264, opened 2021-08-09
 FIO-2764: Fixes an issue when some of the fields were stilling set a default value with the noDefaults option.

MaksimFalei:


MaksimFalei:
Related PR: https://github.com/formio/formio/pull/1306

PR #4266, opened 2021-08-09
Select component - debounce option for server side filtering

llaurentiu:
When querying endpoints for select options and we have server side filtering enabled (searchField is defined), the requests are not debounced which results in spamming the endpoints.

This PR introduces a new builder option to set a debounce timeout value, between 0 and 1s.

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -31,6 +31,7 @@ export default class SelectComponent extends Field {
       lazyLoad: true,
       filter: '',
       searchEnabled: true,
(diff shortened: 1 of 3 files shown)


llaurentiu:
Hi @travist. 
This is ECB.
We created this small PR for a feature that we need. In hope it will get merged soon, please let us know if you are ok with the changes.

cc @elibus

heather-formio:
Hi @elibus This is merged and will be part of 4.14.0, currently in 4.14.0-rc.19. Thank you!

Issue #1773, opened 2019-09-27
Tab conditions


krixerx:
It is not a bug.
We would like to add conditions to tab also like we can add conditions to components.

For example, in first tab we have some questions and depending of these questions, different tabs are visible.

nathanaelbice:
Is there any further information on how to do this, or if this will be supported at some point? This is functionality that I am interested in as well, and I'm a bit surprised that the Tabs component was not implemented as a wrapper around a tab component.

wag110894:
@nathanaelbice, You can use our wizards to accomplish this.  Here is an example of how you can accomplish this with our Wizard: https://formio.github.io/formio.js/app/examples/conditionalwizard.html

nathanaelbice:
@wag110894 Thanks for the suggestion, but I am actually trying to leverage this functionality within the Edit Component modal, with minimally intrusive modifications. Because of this, I believe I will need to use Tabs. I believe that I can modify the tabs component to support this, but I wanted to check for an already supported method first.

randallknutson:
We do have this on our backlog as a feature we would like to add. If you would like to either contribute the functionality in a PR or sponsor the development of it, we'd love to get it done.

heather-formio:
We reviewed this and are not prioritizing on our roadmap, but would welcome and contributions for this request.

PR #4267, opened 2021-08-11
feat: add tab item conditional #1773

jtamm-red:
https://github.com/formio/formio.js/issues/1773


travist:
What makes this one hard is that it will require changes to all the templates we have including semantic, uswds, bootstrap3, etc. Because of this, we will need to hold off on this until we can coordinate that release.

travist:
This PR would need to be re-implemented unfortunately. We are moving all of our templates to their own repo, so the change here would need to be made directly to @formio/bootstrap.

PR #4272, opened 2021-08-13
FIO-3519 Made 'scroll on top' option in wizard go to breadcrumbs if t…

antonSoftensity:
…hey are visible


travist:
We will also need to change the following templates.

PR #4273, opened 2021-08-14
Fix form embedding in async DOM (replace document.write() by DOM insert)

bobslee:
When a document is asynchronously rendered, a document.write() from a different source (eg a script) is ignored and results into following warning. 
Also the embedding of the form fails with an error `Uncaught TypeError: p is null`.

Warning: `A call to document.write() from an asynchronously-loaded external script was ignored.`

### DIFF
--- a/src/Embed.js
+++ b/src/Embed.js
@@ -93,15 +93,19 @@ export function embed(config = {}) {
       return element;
     };
 
-    debug('Embeding Configuration', config);
+    debug('Embedding Configuration', config);
 
     // The id for this embedded form.
     const id = `formio-${Math.random().toString(36).substring(7)}`;
     config.id = id;
 
     debug('Creating form wrapper');
-    document.write(`<div id="${id}-wrapper"></div>`);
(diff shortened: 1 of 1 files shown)


src/Embed.js
```diff
@@ -93,15 +93,19 @@ export function embed(config = {}) {
...
+    debug('Embedding Configuration', config);
 
     // The id for this embedded form.
     const id = `formio-${Math.random().toString(36).substring(7)}`;
     config.id = id;
 
     debug('Creating form wrapper');
-    document.write(`<div id="${id}-wrapper"></div>`);
-    let wrapper = document.getElementById(`${id}-wrapper`);
+    let wrapper = createElement('div', {
```
travist, line 103: This needs to be ```document.createElement```.  Also, the options to this method will not work with IE, so this should probably be the following.

```
let wrapper = document.createElement('div');
wrapper.id = `${id}-wrapper`;
```

travist (changes requested):

Issue #4277, opened 2021-08-18
[BUG] select did not render template on component 'Edit Data'


jtamm-red:
## Environment
- Formio.js version:4.x

## Description
Select component shows id value on component Edit grid. Customer does not understand, what is this value. Should render template as label name!

I choose 'Aspargus' and click button 'Save'.
![image](https://user-images.githubusercontent.com/53046846/129855351-090c74a5-65e1-47d9-9537-ba9423c8c2f6.png)

Hex value appears as id.
![image](https://user-images.githubusercontent.com/53046846/129856252-50d35c6b-df87-44f3-b37c-297bbd24f1d9.png)

## Example
Formio schema
```json
{
  "display": "form",
  "components": [
    {
      "label": "Edit Grid",
      "tableView": false,
      "rowDrafts": false,
      "key": "editGrid",
      "type": "editgrid",
      "input": true,
      "components": [
        {
          "label": "Select",
          "widget": "choicesjs",
          "tableView": true,
          "dataSrc": "json",
          "data": {
            "json": [
              {
                "hex": "#EFDECD",
                "name": "Almond",
                "rgb": "(239, 222, 205)"
              },
              {
                "hex": "#CD9575",
                "name": "Antique Brass",
                "rgb": "(205, 149, 117)"
              },
              {
                "hex": "#FDD9B5",
                "name": "Apricot",
                "rgb": "(253, 217, 181)"
              },
              {
                "hex": "#78DBE2",
                "name": "Aquamarine",
                "rgb": "(120, 219, 226)"
              },
              {
                "hex": "#87A96B",
                "name": "Asparagus",
                "rgb": "(135, 169, 107)"
              },
              {
                "hex": "#FFA474",
                "name": "Atomic Tangerine",
                "rgb": "(255, 164, 116)"
              },
              {
                "hex": "#FAE7B5",
                "name": "Banana Mania",
                "rgb": "(250, 231, 181)"
              },
              {
                "hex": "#9F8170",
                "name": "Beaver",
                "rgb": "(159, 129, 112)"
              },
              {
                "hex": "#FD7C6E",
                "name": "Bittersweet",
                "rgb": "(253, 124, 110)"
              },
              {
                "hex": "#000000",
                "name": "Black",
                "rgb": "(0,0,0)"
              }
            ]
          },
          "idPath": "hex",
          "valueProperty": "hex",
          "template": "<span>{{ item.name }}</span>",
          "key": "select",
          "type": "select",
          "input": true
        }
[...]

PR #4278, opened 2021-08-18
fix: selected option label should appear on edit grid #4277

jtamm-red:


travist:
I am curious about the case where this is broken?  It would be very helpful to provide the JSON of a form that breaks, and then illustrate how this is fixed with this pull request. On that note, please add some automated tests to show how this fix solves a problem.

PR #4281, opened 2021-08-19
fix: element id should be unique

jtamm-red:


travist:
What bug is this fixing? I feel like this will cause some regressions since this is a pretty core place in the code where every component uses the "id". Why is the "id" not unique? This should be a random string.

jtamm-red:
Nested compnent (ex Data Edit) will copy component without random generating of id, when I add new row. I saw that component (not class) id is duplicated. Id is generated on initial as well, but component id append to component's (class) id. You can test code in my PR.


You can check sandbox id: 6123ae72e982096e5fa2c124
You have to try click button 'Add Another' several times. You see console log, where error comes.
image

travist:
I would rather solve this by ensuring that the "id" of the component is removed before the createChildComponent is called from the NestedComponent class. This way it will work better with reverse compatibility.

travist:
Closing due to introduction of performance regressions.

PR #4286, opened 2021-08-23
Fixed an issue with each component recursion.

travist:


sp193:
May I ask, what exactly was fixed? With these commits, it looks like the behaviour of eachComponent() gets changed to only recursing once.

travist:
Just now getting to this and I agree.  This is getting removed with the 5.x renderer and back to the way it was. We are now using the @formio/core version of this method as well.

PR #4302, opened 2021-09-01
FIO-3753: Conditionally shown edit grid with "Open First Row when Empty" enabled doesn't save value

AlehKatsiubaFormIO:


src/components/editgrid/EditGrid.js
```diff
@@ -263,7 +263,7 @@ export default class EditGridComponent extends NestedArrayComponent {
     const data = this.dataValue;
 
     (this.editRows || []).forEach((row, index) => {
-      const rowData = data[index];
+      const rowData = data[index] || {};
```
travist, line 266: This seems like it will detach this new row's data.  Why not something like this?

```
if (!data[index]) {
  this.dataValue[index] = {};
  data = this.dataValue;
}
```

PR #4303, opened 2021-09-01
FIX: Editgrid showing error div

edwinanciani:


src/components/editgrid/EditGrid.js
```diff
@@ -1143,6 +1143,7 @@ export default class EditGridComponent extends NestedArrayComponent {
           const errorContainer = rowContainer.querySelector('.editgrid-row-error');
 
           if (!rowValid && errorContainer && (!this.component.rowDrafts || this.shouldValidateDraft(editRow))) {
+            this.addClass(errorContainer,  'help-block' );
```
travist, line 1146: What if the error is cleared?  Where are we removing the class?

travist (changes requested):

src/components/editgrid/EditGrid.js
edwinanciani: This error only shows when you submit the form and there are not saved rows. and right now it doesn't clear the message in prod until you save the row. @travist

PR #4310, opened 2021-09-06
FIO-3798: fix(Component): calculated value is set to the component on load even if it is hidden and set to clear on hide

alexandraRamanenka:


travist:
Please write a test in a future commit.

PR #4311, opened 2021-09-06
FIO-3752: Check parent component conditions

AlehKatsiubaFormIO:


travist:
The tests look like this causes an infinite loop.

AlehKatsiubaFormIO:
> The tests look like this causes an infinite loop.

@travist  Fixed

PR #4313, opened 2021-09-07
FIO-3758: Duplication of radio ids inside nested forms

AlehKatsiubaFormIO:


travist:
Will also need a fix in:
 - https://github.com/formio/uswds

PR #4315, opened 2021-09-07
FIO-3616: next generation rules updates

TanyaGashtold:


TanyaGashtold:
Related to:
https://github.com/formio/premium/pull/148
https://github.com/formio/formio-app/pull/960

PR #4316, opened 2021-09-07
FIO-3760: Custom Component сtx.component.fields is undefined

AlehKatsiubaFormIO:


travist:
Will also need a fix in:
 - https://github.com/formio/uswds

PR #4317, opened 2021-09-07
API | Receiving Validation Errors when rendering a form with a submission in draft-state

yavorovsky:
…setValue" had been calling for "content" and "htmlelement"

### DIFF
--- a/src/components/content/Content.js
+++ b/src/components/content/Content.js
@@ -29,6 +29,11 @@ export default class ContentComponent extends Component {
     return ContentComponent.schema();
   }
 
+  constructor(component, options, data) {
+    super(component, options, data);
+    this.options.noDefaults = true;
+  }
+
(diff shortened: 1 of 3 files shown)


src/components/content/Content.js
```diff
@@ -29,6 +29,11 @@ export default class ContentComponent extends Component {
     return ContentComponent.schema();
   }
 
+  constructor(component, options, data) {
+    super(component, options, data);
+    this.options.noDefaults = true;
```
travist, line 34: ```this.options``` is a globally shared options object. What this means is that when you set it here, it will set it for ALL components. Is this intended?

travist (changes requested):

travist:
Ok, we will need to revert it then. We should be able to do it for the next RC release.

PR #4345, opened 2021-09-20
FIO-3744 fixed an issue where unique validation is not honored for re…

AlexanderYakubovskiy:
…source component on submit


travist:
Wouldn't this be considered an incorrectly configured component?  Also, would this not break other templates we have out there where they do provide ```item.data.firstName``` in the template?

AlexanderYakubovskiy:
Yes, I will continue investigating on other solution.

PR #4349, opened 2021-09-22
FIO-3970: fixed an issue where no table headers display when 'Open First Row when Empty' setting is enabled for Edit Grid

TanyaGashtold:


TanyaGashtold:
all tests passed locally

PR #4353, opened 2021-09-23
FIO-3984: First click on a date does not select it

AlehKatsiubaFormIO:


elibus:
@heather-formio we would like to see this merged in the next rc if possible.

PR #4368, opened 2021-09-29
FJS | 4.x Regression | Button: Button Custom logic does not work

yavorovsky:


yavorovsky:
The failed test isn't related to the changes.

Issue #3687, opened 2021-01-22
Create a form with a submission


paulfalgout:
We're programmatically creating forms and passing the json objects to them.  When running `createForm` it would be preferred if both the definition and submission could be instantiated simultaneously.

Currently changing the submission in the `then` allows for the possibility of some custom javascript on a component acting based on no submission data, where it's already done its action by the time the submission data is immediately applied.

I'd like:
```js
Formio.createForm(document.getElementById('formio'), formDef, { submission })
```

Instead of
```js
Formio.createForm(document.getElementById('formio'), formDef).then(form => form.submission = submission);
```

travist:
I agree with this. I actually would not be that difficult of an addition. If anyone submits a PR for it, we would be happy to accept it.

olgabann:
We're currently addressing a backlog of GitHub issues. Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #4393, opened 2021-10-14
feat: create form with submission #3687

jtamm-red:
https://github.com/formio/formio.js/issues/3687


travist:
It really is not that hard to do the following.

```
Formio.createForm(document.getElementById('formio'), formJSON).then((form) => {
  form.submission = subJSON;
});
```

Because of this, we will not include this change.

PR #4398, opened 2021-10-15
Add capture option to file fields

bmcbride:
This adds a new `capture` option to file fields, which allows a mobile device to open the camera or microphone directly in [capture mode](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture). This offers more flexibility over the current `webcam` setting as it lets the user configure which device to launch (camera, video, microphone- based on `filePattern`) and specify front or rear device, on supported platforms (iOS). Capture can be used alongside webcam and is ignored on desktop.

While working on this, I also tried to tidy up the file input a bit, removing the "Drop files to attach" label on mobile and adding some icons to the link buttons.

### Launch rear-facing camera
```js
{
  type: "file",
  label: "My Photos",
  key: "photos",
  storage: "base64",
  image: true,
  capture: "environment",
  filePattern: "image/*",
  multiple: true,
  imageSize: "250"
}
```

### Launch front-facing video camera
```js
{
  type: "file",
  label: "My Videos",
  key: "videos",
  storage: "base64",
  capture: "user",
  filePattern: "video/*",
  multiple: true,
  imageSize: "250"
}
```

### Launch microphone
```js
{
  type: "file",
  label: "My Audio",
  key: "audio",
  storage: "base64",
  capture: "environment",
  filePattern: "audio/*",
  multiple: true,
  imageSize: "250"
}
```

image

### DIFF
--- a/src/components/file/File.js
+++ b/src/components/file/File.js
@@ -284,9 +284,12 @@ export default class FileComponent extends Field {
     if (this.component.multiple) {
       options.multiple = true;
     }
+    if (this.component.capture) {
+      options.capture = this.component.capture;
+    }
     //use "accept" attribute only for desktop devices because of its limited support by mobile browsers
+    const filePattern = this.component.filePattern.trim() || '';
     if (!this.isMobile.any) {
-      const filePattern = this.component.filePattern.trim() || '';
       const imagesPattern = 'image/*';
 
       if (this.imageUpload && (!filePattern || filePattern === '*')) {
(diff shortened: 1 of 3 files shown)


kamel-msaoubi:
Will we have a chance to see this PR merged in 5.x.x ?
Looking for this feature for resolving the issue: when we are taking picture from a mobile, we cannot switch from front camera (selfie) to back camera, on Android.

PR #4415, opened 2021-10-25
FIO-3528 Changed wizardnav button classes for better action indcitaion

antonSoftensity:


src/templates/bootstrap/wizardNav/form.ejs
```diff
@@ -2,12 +2,12 @@
   {% ctx.buttonOrder.forEach(function(button) { %}
     {% if (button === 'cancel' && ctx.buttons.cancel) { %}
     <li>
-      <button class="btn btn-secondary btn-wizard-nav-cancel" ref="{{ctx.wizardKey}}-cancel" aria-label="{{ctx.t('cancel')}} button. Click to reset the form">{{ctx.t('cancel')}}</button>
```
travist, line 5: I don't want to use the danger class here.  I am not a fan of a bright red button on the forms. I would rather just make the cancel button not on by default or maybe make it a small link or something... we need to discuss this one.

travist (changes requested):

lane-formio:
Closing per comments left on https://formio.atlassian.net/browse/AI-1792

PR #4429, opened 2021-11-02
FIO-3969: fixed an issue where lodash is still executing functions wh…

AlexanderYakubovskiy:
…en protected-eval is being used

### DIFF
--- a/src/formio.form.js
+++ b/src/formio.form.js
@@ -92,6 +92,11 @@ const registerPlugin = (plugin) => {
       case 'evaluator':
         Formio.Evaluator.registerEvaluator(plugin.evaluator);
         break;
+      case 'default':
+        if (plugin.default.evaluator) {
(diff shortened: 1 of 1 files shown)


src/formio.form.js
```diff
@@ -92,6 +92,11 @@ const registerPlugin = (plugin) => {
       case 'evaluator':
         Formio.Evaluator.registerEvaluator(plugin.evaluator);
         break;
+      case 'default':
```
travist, line 95: Why is the plugin name called "default"?  This seems really strange like we should ix the place where "evaluator" is not being used.

travist (changes requested):

src/formio.form.js
AlexanderYakubovskiy: I understand. Going to investigate this issue some more.

PR #4436, opened 2021-11-03
FIO-3814: do not fetch Select options again if there was an error

mikekotikov:


mikekotikov:
After running build script the tests passed at my local

PR #4447, opened 2021-11-09
FIO-4241: fixed an issue where calculation validation not showing unt…

AlexanderYakubovskiy:
…il submission when numbers are inside datagrid component


AlexanderYakubovskiy:
This is test for FIO-4241 which reproducible in 4.13.7 but fixed in master by:
https://github.com/formio/formio.js/pull/4323
https://github.com/formio/formio.js/pull/4049

PR #4453, opened 2021-11-15
Fix: components which could not trigger focus/blur events

rthjy:
1. add focus/blur events to Radio, Checkbox, SelectBoxes
2. add focus/blur events to the user input box of datetime component

### DIFF
--- a/src/components/_classes/field/Field.js
+++ b/src/components/_classes/field/Field.js
@@ -1,3 +1,4 @@
+import { delay } from '../../../utils/utils';
 import Component from '../component/Component';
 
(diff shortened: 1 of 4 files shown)


travist:
I am fearful that this is going to cause serious regressions because it adds 2 event listeners for every component on the form. I think a better approach is to just make these kinds of event listeners outside of the rendered form if you need them.

```
Formio.createForm(document.getElementById('formio'), ....).then((form) => {
  const blurComp = form.getComponent('datetime');
  blurComp.input.addEventListener('blur', ....);
});
```

PR #4454, opened 2021-11-15
fix issue: when submission.data doesn't exist, the submission.metadata is removed

rthjy:
fix issue: when submission.data doesn't exist, the submission.metadata is removed


travist:
I like it. Thank you!

PR #4456, opened 2021-11-16
Feat/fio 3144 parent form data

mikekotikov:


src/components/_classes/component/Component.js
```diff
@@ -2315,6 +2316,25 @@ export default class Component extends Element {
     return this.root ? this.root.data : this.data;
   }
 
+  /**
+   * Get the data value from the top-level form (including nested forms).
+   *
+   * @return {*}
+   */
+  get rootData() {
```
travist, line 2324: I don't want to add this. If you wish to reference your parent forms data, you can use the following.

```js
instance.root.parent.root.data;
```

travist (changes requested):

src/components/_classes/component/Component.js
mikekotikov: Yeah, sure. I just thought it to be more convenient as you may also have a couple of levels of nesting. And this way it's easier to go from top level form to child by it's api key.
Nevertheless, there is a possibility to reach the data on parent forms now, so if you don't see this changes necessary I'll close it.

PR #4464, opened 2021-11-23
Evaluate custom condition for checkbox

rajiv-kumar-mck:


travist:
I am pretty sure this will break Checkbox components with Radio set as the input type. Please verify....

travist:
I think you are right.  We will give this a try and see how it does. I do like the idea of removing code that fixes bugs.

PR #4497, opened 2021-12-10
FIO-3884 Fixed issue when conditionally required fields weren't indic…

antonSoftensity:
…ated as required in labels


michealharrington:
I have run into an issue with this fix.
My datagrid's are being redrawn after every change in the form.
Debugging into the code, it is showing that the column compare isEqual check is always returning false, but the only difference in the data is that the auto-generated "id" field is different.

My suggestion for a solution would be to change the isEqual to compare all fields EXCEPT the "id" field.

PR #4503, opened 2021-12-15
FIO-4362/4363: fix checkbox type radio incorrect value

mikekotikov:


mikekotikov:
@alexandraRamanenka took this over

PR #4518, opened 2021-12-29
Fixing issue where styles declaration for components is wrong.

travist:


travist:
This was fixed @ https://github.com/formio/bootstrap

PR #4545, opened 2022-01-20
Bootstrap 5 Support

ajhalls:
Updated the builderSidebar and builderSidebarGroup to support the new Bootstrap 5 style. Preliminary evaluation shows that to be the main roadblock. There may be additional modifications needed for individual items. Cancelled last request to bundle with latest change.


travist:
This is coming in 5.x.  Thanks for your contribution!

PR #4569, opened 2022-02-02
FIO-4558: fixed an issue where html element displays error if contain…

AlexanderYakubovskiy:
…s only word select


travist:
This is not the correct place to fix this. This will affect every translation that is performed (which is many).

Issue #3684, opened 2021-01-22
[BUG] wizard pages caching data overrides set submission


paulfalgout:
Sorry to break with the issue formatting, but I do think I've boiled this head scratcher down to the simplest reproducible issue.

Breaks with 4.12.x +
I think the issue may have been introduced here https://github.com/formio/formio.js/commit/3d7992f9987b79ad0c1930ba7633684d3d4e1a73 but that's a bit of a guess based on git blame.

As best as I can determine this is a complicated set of configuration, but if you have a wizard with two tabs, and the first tab has a datagrid that contains an htmlelement, then it will write over any submission data set after form creation.

Here is the example working in 4.11.0: https://jsfiddle.net/brepxfq7/
Here is the same code broken in 4.12.5: https://jsfiddle.net/mvoeag2p/ (notice the textarea is empty on the second tab)
Here is 4.12.5 working if the datagrid/html tab is the 2nd tab instead of the 1st https://jsfiddle.net/23fjzmob/

Also if this helps narrow it down:  If you duplicate the `form.submission =...` line and do it twice instead of once in the `then` of the form creation it will fix the issue.. granted it will draw the form a bunch.

paulfalgout:
No improvement with 4.13.0

paulfalgout:
This appears to be something to do with attempting to set the default value on each datagrid on every page while prior to the pages getting the data which in some cases reduces the submission passed in.  It's odd that an htmlelement even tries to set a default value, but that's what is happening.

PR #4580, opened 2022-02-07
Add submission as form creation option

paulfalgout:
I couldn't actually get the test to run.  Too many failing tests, but this works in a jsfiddle for at least the use cases I'm aware of.

Solves: https://github.com/formio/formio.js/issues/3687
~~Fixes: https://github.com/formio/formio.js/issues/3684~~. It addresses this issue, but this issue is still broken if you're using the submission in the regular fashion of setting it in the `then`

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -925,7 +925,14 @@ export default class Webform extends NestedDataComponent {
    * Build the form.
    */
   init() {
-    this._submission = this._submission || { data: {} };
+    if (this.options.submission) {
+      const submission = _.extend({}, this.options.submission);
+      this._submission = submission;
+      this._data = submission.data;
+    }
+    else {
+      this._submission = this._submission || { data: {} };
+    }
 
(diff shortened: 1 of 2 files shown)


paulfalgout:
Screen Shot 2022-02-13 at 11 52 35 PM
FWIW on a datagrid if this is set to true it'll clear out the related data on a submission set at instantiation

travist:
I like it!!! We are working on the failing tests but I see no reason to include this in the 4.15.x version

paulfalgout:
@travist been using this for a couple of weeks now by just overriding the prototype and honestly it's cleared up a lot of quirks we were working around.

TriStarGod:
Hi @paulfalgout , could you give me an example of how you are overriding the prototype to get this functionality? There are couple of features I want to use but currently waiting on merges.

travist:
This will be part of the 5.x release!  Thank you for your contribution!

PR #4594, opened 2022-02-15
UIP-317: Fixes protected eval crushing

mikekotikov:


src/components/_classes/component/Component.js
```diff
@@ -1539,7 +1539,7 @@ export default class Component extends Element {
         data: this.rootValue
       }),
       form: this.root ? this.root._form : {},
-      options: this.options,
+      options: FormioUtils.Evaluator.protectedEval ? {} : this.options,
```
travist, line 1542: This change does not make much sense. You do not wish to pass the options if we are using protectedEval plugin?  I do not understand how this would affect the renderer.

travist (changes requested):

src/components/_classes/component/Component.js
mikekotikov: I can actually move this to `protected-eval` exclusions, as [here](https://github.com/formio/protected-eval/blob/master/src/ProtectedEvaluator.ts#L14)
Because many properties in `options` cause stack overflow as they get infinitely converted to js-interpreter objects

travist: Yep that will work. I like that better. Just move it into the protected-eval plugin.

PR #4629, opened 2022-03-03
FIO-4664 Fixed conditional logic based on Address component

antonSoftensity:


src/components/address/Address.js
```diff
@@ -158,7 +158,7 @@ export default class AddressComponent extends ContainerComponent {
         mode: AddressComponentMode.Autocomplete,
         address: {},
       }
-      : {};
+      : '';
```
travist, line 161: I would believe that the empty value for an address, which is usually an object, is an empty object. I do not agree with this change since the data type for the empty value is not compatible with the populated value. I think this needs some more work.

travist (changes requested):

PR #4630, opened 2022-03-04
displayed the value that was actually selected

HannaKurban:


src/components/select/Select.js
```diff
@@ -505,6 +505,10 @@ export default class SelectComponent extends Field {
   /* eslint-enable max-statements */
 
   get defaultValue() {
+    if (this.options.display === 'pdf' && (this.options.readOnly || this.root.formio?.submissionId)) {
+      this.component.defaultValue = '';
```
travist, line 509: This is an unusual change. What is the issue that you are fixing here, and why is it necessary to reset the defaultValue of the component to fix it?  Is there another way?

PR #4645, opened 2022-03-16
fixed Advanced Conditional based on numeric value

HannaKurban:


src/components/number/Number.js
```diff
@@ -98,7 +98,7 @@ export default class NumberComponent extends Input {
   }
 
   get emptyValue() {
-    return '';
+    return '\0';
```
travist, line 101: I would like to understand why this change was made. This seems like it will cause regressions.

travist (changes requested):

PR #4656, opened 2022-03-24
Update translations for aria-labels

Willem-Dekker:
Whole label can now be localized instead of only a small piece


travist:
Thank you!

PR #4658, opened 2022-03-25
Allow customization of edit grid error messages

juliapath:
In #3580 `invalidRowError`, `invalidRowsError` and `unsavedRowsError` have been made translatable, but it should also be possible to customize them on a per component basis.


travist:
Thank you!

PR #4663, opened 2022-03-30
FIO-4851: fixed when 'Multiple Values' setting is enabled, all the selected val…

AlexanderLihodievskiy:
…ues display as the last

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -249,8 +249,8 @@ export default class SelectComponent extends Field {
     }
     if (typeof data === 'string') {
       const selectData = this.selectData;
-      if (selectData) {
-        data = selectData;
+      if (selectData && selectData[value]) {
+        data = selectData[value];
       }
       else {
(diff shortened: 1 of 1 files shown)


src/components/select/Select.js
```diff
@@ -249,8 +249,8 @@ export default class SelectComponent extends Field {
     }
     if (typeof data === 'string') {
       const selectData = this.selectData;
-      if (selectData) {
-        data = selectData;
+      if (selectData && selectData[value]) {
```
travist, line 252: What if this.selectData is a string?

travist (changes requested):

src/components/select/Select.js
AlexanderLihodievskiy: I can't see any cases when this.selectData can be a string. There is only one place in code where selectData changes and it can be only object. 
See please src/components/select/Select.js:1425 
If you know any cases when selectData can be a string, please tell me.

PR #4672, opened 2022-04-06
FIO-4752 Added autoscroll when dragging components inside of formbuilder

antonSoftensity:


travist:
Wouldn't this move the page as you are trying to place items in the builder?  At what point does it autoscroll?

antonSoftensity:
> Wouldn't this move the page as you are trying to place items in the builder? At what point does it autoscroll?

Yeah, it does autoscroll the page when we pick up the item from the sidebar and drag it towards the end of the viewport, in case if the form is too long vertically and we have a scroll

PR #4677, opened 2022-04-18
FIO-4953: Fixed an issue where signature component was not populating…

travist:
… the data properly.


astef:
FYI, still not working here: https://formio.github.io/formio.js/app/builder

PR #4707, opened 2022-05-12
FIO-1425: Fixed submission object empty when using calculated values …

AlexanderLihodievskiy:
…on the server

### DIFF
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -1536,9 +1536,7 @@ export default class Component extends Element {
       // It is useful to translate strings in different scenarions (eg: custom edit grid templates, custom error messages etc.)
       // and desirable to be publicly available rather than calling the internal {instance.t} function in the template string.
       t: this.t.bind(this),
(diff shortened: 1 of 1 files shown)


src/components/_classes/component/Component.js
```diff
@@ -1536,9 +1536,7 @@ export default class Component extends Element {
       // It is useful to translate strings in different scenarions (eg: custom edit grid templates, custom error messages etc.)
       // and desirable to be publicly available rather than calling the internal {instance.t} function in the template string.
       t: this.t.bind(this),
-      submission: (this.root ? this.root._submission : {
-        data: this.rootValue
-      }),
+      submission: { data: this.rootValue },
```
travist, line 1539: Won't this break if we need to get the submission metadata?  This needs to be able to handle the full submission object if it is available.

travist (changes requested):

src/components/_classes/component/Component.js
AlexanderLihodievskiy: Okay, I'll reconsider the decision.

PR #4715, opened 2022-05-19
FIO-3959: fixed an issue where dataTable error list does not redirect…

AlexanderYakubovskiy:
… you to the required component

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -1128,13 +1128,23 @@ export default class Webform extends NestedDataComponent {
       this.refs.errorRef.forEach(el => {
         this.addEventListener(el, 'click', (e) => {
           const key = e.currentTarget.dataset.componentKey;
-          this.focusOnComponent(key);
+          if (!this.focusOnComponent) {
+            this.root?.focusOnComponent(key);
+          }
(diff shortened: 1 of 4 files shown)


src/Webform.js
```diff
@@ -1128,13 +1128,23 @@ export default class Webform extends NestedDataComponent {
       this.refs.errorRef.forEach(el => {
         this.addEventListener(el, 'click', (e) => {
           const key = e.currentTarget.dataset.componentKey;
-          this.focusOnComponent(key);
+          if (!this.focusOnComponent) {
```
travist, line 1131: Why is this method not defined?  Is the "this" pointer not the webform class?  This seems like a strange change to get around a different problem.

travist (changes requested):

src/Webform.js
AlexanderYakubovskiy: Yes, "this" indicates to the component which caused error. Not the webform.

travist: But this is called within an arrow function... wouldn't that inherit context from the bounding context (ie webform instance)?

PR #4750, opened 2022-06-29
FIO-5152: fixed an issue where value of html element displays differe…

AlexanderYakubovskiy:
…ntly on pdf download and form builder pages


src/templates/bootstrap/html/form.ejs
```diff
@@ -1,4 +1,4 @@
-<{{ctx.tag}} class="{{ ctx.component.className }}" ref="html"
```
travist, line 1: Doesn't this mean that if they provide a "className" value in their component, that it will still be broken?  Shouldn't this be the following?

```
<{{ctx.tag}} class="formio-component-htmlelement {{ ctx.component.className }}"
```

travist (changes requested):

PR #4753, opened 2022-06-30
only generate key when it doesn't exist for custom group

claytongulick:


claytongulick:
This is to address the problem with the key field getting stomped on with custom components

travist:
I like this change and am considering it, but this seems to only take into affect if you are using a group called "custom". What I would prefer is to make is so that this works if the component type is not found like so.

```
if (!Components.components.hasOwnProperty(info.type)) {
```

claytongulick:
Thanks @travist !

That PR was issued from when I was working for a company Telos Health Solutions, I'm not with them anymore and don't have access to the repo to modify the PR.

I recall this issue being a pretty big problem when we were mapping custom components to database columns, the ID getting stomped on forced us to fix it locally and operate off a fork.

Wonderful library, very powerful - keep up the amazing work!

travist:
No worries... I can make the change myself if you agree with the change.

travist:
I will make the change I suggested after I merge. Thank you for your contribution!

PR #4754, opened 2022-07-04
FIO-4551 added unique IDs inside the Field Sets

HannaKurban:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-4551

## Description
Added unique IDs to the fields inside the NestedComponents in the Data Grid rows


src/components/fieldset/Fieldset.js
```diff
@@ -40,4 +41,11 @@ export default class FieldsetComponent extends NestedComponent {
     super(...args);
     this.noField = true;
   }
+
+  createComponent(component, options, data, before) {
+    const comp = super.createComponent(component, options, data, before);
+    if (document.querySelectorAll(`[id^=${comp.id}]`).length) {
```
travist, line 47: This seems really strange. Why didn't createComponent add the id in the first place? Why do we need to add a querySelectorAll here?

travist (changes requested):

src/components/fieldset/Fieldset.js
HannaKurban: removed querySelectorAll

HannaKurban:
all tests are passed

src/components/fieldset/Fieldset.js
```diff
@@ -40,4 +41,9 @@ export default class FieldsetComponent extends NestedComponent {
     super(...args);
     this.noField = true;
   }
+
+  createComponent(component, options, data, before) {
+    component.id =  getRandomComponentId();
+    return super.createComponent(component, options, data, before);
```
TanyaGashtold, line 47: I have checked the ticket. Does this issue effect only Fieldset component? I am pretty sure that all nested components (like Panel, Container and other) inside NestedArray components have the same issue with ids. It worth moving the fix to the NestedComponent/NestedDataComponent/NestedArrayComponent level.

TanyaGashtold (changes requested):

src/components/fieldset/Fieldset.js
HannaKurban: added the fix tio the NestedComponent

TanyaGashtold (approved):

TanyaGashtold (approved):

PR #4757, opened 2022-07-06
FIO-5182: Fixes an issue where Form.io logs you out if request to the external resoure uses Form.io Authentication and returns Bad Token response

alexandraRamanenka:


travist:
This was pulled into @formio/core.

PR #4758, opened 2022-07-06
FIO-4782 Ampersand is replaced by %26

HannaKurban:


HannaKurban:
tests pass locally

src/components/select/Select.js
```diff
@@ -595,7 +595,7 @@ export default class SelectComponent extends ListComponent {
 
     // Add filter capability
     if (this.component.filter) {
-      url += (!url.includes('?') ? '?' : '&') + this.interpolate(this.component.filter);
+      url += (!url.includes('?') ? '?' : '&') + this.interpolate(this.component.filter.replace(/&/g, '26amp').replace(/(26amp){2}/g, '&&')).replace(/&/g, '%26').replace(/26amp/g, '&');
```
TanyaGashtold, line 598: This solution does not resolve the whole issue. Here just the &-symbol is replaced. What if some other symbols that are not acceptable in query (like #, +, ? and so on) are used? The filter query will not work.
My thoughts:
1) I am not sure that we need to have this in our code at all. The filter query is something that is defined by the customer in select component settings and the customer should be responsible for query correctness. E.g. encodeURIComponent can be used inside the interpolated string, like this:
`data.tech={{encodeURIComponent(data.main)}}&submitted=true`
2) we need to find a way to parse the filter query string and then pass each parameter through encodeURIComponent. But it may not work correctly as most probably we will parse the query splitting it by & and =, that means those symbols cannot be used in query.
3) as an option, we can allow users to pass filter parameters as an object, e.g. {'par1' : '{{data.test}}'}

TanyaGashtold (changes requested):

PR #4773, opened 2022-07-13
added fallback for timezones loading to load timezones from formio-vi…

AlexeyNikipelau:
…ewer static content


travist:
I would prefer to instead have the zones loaded into the bundle and if it detects that, then it does not need to load the zones at all. We would then do this for the PDF renderer since this is our own renderer bundle.

PR #4775, opened 2022-07-14
FIO-5262: Add option to not make new api requests when scrolling through search results

mikekotikov:


src/components/select/Select.js
```diff
@@ -1121,7 +1121,7 @@ export default class SelectComponent extends Field {
         }
       }
 
-      if (this.itemsFromUrl) {
+      if (this.itemsFromUrl && !this.component.noOnScrollRefresh) {
```
TanyaGashtold, line 1124: 1) Should this option prevent scrolling only when viewing search results? Seems YES based on the PR description. If so, I am afraid this check prevents loading of additional select options at all when 'noOnScrollRefresh' is enabled. Please check this case.
2) It might be better to call the option 'noRefreshOnScroll'. For me it sounds better.

TanyaGashtold (changes requested):

src/components/select/Select.js
TanyaGashtold: Additionally please confirm that all tests are passing at least locally. For some reason CircleCI did not start tests for this PR.

mikekotikov: @TanyaGashtold 
1) You're right, with this option enabled infinite scroll won't work, because it prevents results refetching.
2) Totally agree, renamed.

TanyaGashtold (approved):

PR #4785, opened 2022-07-21
FIO-4705: fixed an issue with bad request error in console when rende…

AlexanderYakubovskiy:
…r pdf form


travist:
This change does not seem right... why did these have to be reversed?

AlexanderYakubovskiy:
My bad. Refactored it.

PR #4799, opened 2022-08-03
FIO-5310: fixewd an issue where date showing incorrectly when manuall…

AlexanderYakubovskiy:
…y typing month


TanyaGashtold (changes requested):
This issue is expected to be resolved by https://github.com/formio/flatpickr/pull/15 and https://github.com/formio/flatpickr/pull/17. I think we can close this PR.

PR #4814, opened 2022-08-11
FIO-4856: Fixed an issue where onBlur validation does not work for Select, Tags and Day components

alexandraRamanenka:
https://formio.atlassian.net/browse/FIO-4856

Select and Day component do not use Input class (they extend Field directly instead), but because those components actually have inputs they should share the same functionality as other input components, like triggering onBlur validation and saving the focused element properly to restore a focus after form redraw. 
Despite Tags component extends Input class, the formiojs input elements is not getting blur event (choices.js input element is getting the event instead), so I added emit to pass the blur event to formio.js input element and fire the listener attached in the Input class


travist:
Breaks tests....  reverting and will need to be resubmitted.

alexandraRamanenka:
> Breaks tests.... reverting and will need to be resubmitted.

https://github.com/formio/formio.js/pull/5294

PR #4815, opened 2022-08-11
FIO-5371: fixed an issue where changes of configuration on settings p…

AlexanderYakubovskiy:
…age do not save

### DIFF
--- a/src/components/_classes/nested/NestedComponent.js
+++ b/src/components/_classes/nested/NestedComponent.js
@@ -728,7 +728,7 @@ export default class NestedComponent extends Field {
     if (component.type === 'button') {
       return false;
     }
-    if (value && component.hasValue(value)) {
+    if (value && component.hasValue(value) && component.component.type !== 'tabs') {
       return component.setValue(_.get(value, component.key), flags);
     }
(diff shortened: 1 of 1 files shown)


src/components/_classes/nested/NestedComponent.js
```diff
@@ -728,7 +728,7 @@ export default class NestedComponent extends Field {
     if (component.type === 'button') {
       return false;
     }
-    if (value && component.hasValue(value)) {
+    if (value && component.hasValue(value) && component.component.type !== 'tabs') {
```
travist, line 731: This code should be within the Tabs component which derives from NestedComponent. You need to create a method in there that can be overridden. Maybe have "hasValue" return false?

travist (changes requested):
Let's discuss.

src/components/_classes/nested/NestedComponent.js
AlexanderYakubovskiy: This issue already fixed by this PR: https://github.com/formio/formio.js/pull/4812.
So we don't need my changes, this PR can be closed.

PR #4818, opened 2022-08-15
Correct link to formio.js

tkalmar:
Casing is hard


travist:
Unfortunately this is auto generated. We will need to figure out a way to fix it with the generated code.

PR #4823, opened 2022-08-18
FIO-5418: Fixes checkbox values in edit grid not translated

mikekotikov:


src/components/checkbox/Checkbox.js
```diff
@@ -183,7 +183,7 @@ export default class CheckBoxComponent extends Field {
   }
 
   getValueAsString(value) {
-    return value ? 'Yes' : 'No';
+    return this.t(value ? 'Yes' : 'No');
```
biftime, line 186: Should these keys be added also in https://github.com/formio/formio.js/blob/master/src/translations/en.js ?

travist: Thank you for your review. I agree, yes it would normally be the case, but for this case "Yes" and "No" can very easily be added to the translations like

```
{
  i18n: {
    es: {
      'Yes': 'Si',
      'No': 'No'
    }
  }
}
```

PR #4829, opened 2022-08-22
FIO-5376: fixed an issue where unique validation errors do not disapp…

AlexanderYakubovskiy:
…ear after changing fields values

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -1356,6 +1356,7 @@ export default class Webform extends NestedDataComponent {
     let isChangeEventEmitted = false;
     // For any change events, clear any custom errors for that component.
     if (changed && changed.component) {
+      this.clearServerErrors();
       this.customErrors = this.customErrors.filter(err => err.component && err.component !== changed.component.key);
     }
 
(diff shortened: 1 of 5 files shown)


src/Webform.js
```diff
@@ -1356,6 +1356,7 @@ export default class Webform extends NestedDataComponent {
     let isChangeEventEmitted = false;
     // For any change events, clear any custom errors for that component.
     if (changed && changed.component) {
+      this.clearServerErrors();
```
TanyaGashtold, line 1359: That means if component A has server error and we change any other component in the form (e.g. component B that is valid), the error from component A disappears. This behavior does not seem to me correct. Only if change happens for the component with server error, the error should disappear and only from the changed component. If there are some other components in the form with server errors, those errors should persist until those components are changed.

TanyaGashtold (changes requested):

src/Webform.js
TanyaGashtold: Additionally please rebase PR

PR #4831, opened 2022-08-22
interpolate redirectURI in button oauth config

devonpis:
interpolate redirectURI in oauth settings so it could construct the redirectURI dynamically.
for example, redirect to different subdomain but having the same folder path in different project stages.
eg. {{window.location.origin}}/test/
another example is constructing the redirectURI using data object.
eg. {{data.myUrl}}


travist:
👍

PR #4835, opened 2022-08-24
FIO-5445: fixed an issue where simple mappings setting was empty on s…

AlexanderYakubovskiy:
…ave submission action

### DIFF
--- a/src/components/_classes/nesteddata/NestedDataComponent.js
+++ b/src/components/_classes/nesteddata/NestedDataComponent.js
@@ -25,6 +25,9 @@ export default class NestedDataComponent extends NestedComponent {
   }
 
   componentContext() {
+    if (this.component.type === 'resourcefields') {
+      return this._data;
+    }
     return this.dataValue;
   }
 


src/components/_classes/nesteddata/NestedDataComponent.js
```diff
@@ -25,6 +25,9 @@ export default class NestedDataComponent extends NestedComponent {
   }
 
   componentContext() {
+    if (this.component.type === 'resourcefields') {
```
travist, line 28: Shouldn't the ResourceFields component just override this method and return ```this._data```?  This seems like a hack that we will regret in the future.

travist (changes requested):

src/components/_classes/nesteddata/NestedDataComponent.js
AlexanderYakubovskiy: Sure. ResourceFields component from formio-app, so I've created PR with componentContext method there.
https://github.com/formio/formio-app/pull/1223
This PR can be closed.

PR #4841, opened 2022-08-28
Fix and cleanup form submission api error handling

hschletz:
When the form submission request results in a 4xx or 5xx response, showErrors() is invoked with a message string. This is no longer supported and results in an exception, causing script abortion without any feedback to the user.

This PR removes the showErrors() call as it seems to be a leftover from a previous implementation that supported string arguments. The message is still displayed via setAlert().

I also cleaned up the error handling code by deduplicating the message composition code.


travist:
👍

PR #4842, opened 2022-08-29
Hide removeItemButton on choicesjs select when the placeholder is selected

paulsukow:
Screen Shot 2022-08-29 at 12 22 17 PM

Fixing a bug where the choices.js's removeItemButton (the x button in the screen shot) is being shown for the placeholder option.

According to the [choices.js readme](https://github.com/Choices-js/Choices#placeholder). The value of the placeholder option should be set to `''`. When the value is set to `''` the remove button is hidden.


travist:
👍

PR #4852, opened 2022-09-08
FIO5253 Changed pdf loading spinner color to green

HannaKurban:


travist:
This should probably not change in the OSS renderer. If we need this to be a different color in the application, then we just need to do a CSS override there.

Issue #4851, opened 2022-09-07
Error when running `Formio.builder(...)` after some delay/events


bilogic:
This is part of my HTML
```HTML
<head>
    ...
    <link rel="stylesheet" href="https://cdn.form.io/formiojs/formio.full.min.css">
    <script src="https://cdn.form.io/formiojs/formio.full.min.js"></script>
</head>
<body>
    ...
    <script>
        Formio.builder(document.getElementById('builder'), {}, {})
            .then(function(builder) {
                builder.on('saveComponent', function() {
                    console.log(builder.schema);
                });
            });

        function render() {

        }
    </script>
</body>
```
![image](https://user-images.githubusercontent.com/946010/188776277-a00608eb-f592-40e8-8f10-de6eb2fc0aea.png)

1. Using the above code, form builder shows up correctly.
2. As I need the Formio code to run only after some events, I moved them into `render()`
3. But running `render()` gets the `Missing projectId` warning + error, and the builder doesn't render to a workable state
4. Any idea how I can troubleshoot?

Thank you.

_Originally posted by @bilogic in https://github.com/formio/formio.js/issues/4048#issuecomment-1238843625_

bilogic:
![image](https://user-images.githubusercontent.com/946010/188786936-47eb8ae4-585f-4333-a449-dd5c360822da.png)

I narrowed it down to `utils.js::bootstrapVersion()`

1. When outside `render()`, it goes to line 47 and `returns 0;`
2. When inside `render()`, it goes to line 44

Any idea what this set of logic is for? Especially the use of `$`.
Sorry I'm not familiar with javascript, but isn't `$` like a global variable or something?

Thank you.

bilogic:
```HTML
<script>
    function render() {
        $ = null; // this works around the issue

        Formio.builder(document.getElementById('builder'), {}, {})
            .then(function(builder) {
                builder.on('saveComponent', function() {
                    console.log(builder.schema);
                });
            });
    }
</script>
```
I worked around the issue by writing the above
But would love to see this issue get resolved "properly". Thank you.

0liver:
I seem to be missing a critical part of your setup, because in your code extract the `render()` method is empty, and it's not clear when and how you call it. Also, the formio initialization code outside of `render()` confuses me.

> 2. As I need the Formio code to run only after some events, I moved them into render()

What do you mean by that?

bilogic:
Hi,

`render()` is empty, that is the exact code I'm using.
[...]

PR #4854, opened 2022-09-08
use jQuery instead of $

bilogic:
prevent clashing of `$` as seen in #4851


bilogic:
@travist possible to have this? Thank you.

travist:
👍

Issue #4862, opened 2022-09-16
[HELP] What is the response format?


bilogic:
I keep getting this exception in console log

![image](https://user-images.githubusercontent.com/946010/190608522-13fe1941-af43-4125-b909-c004e171c23c.png)

# Part 1
https://github.com/formio/formio.js/blob/040e5901d977083d08b540ce27d772bcc704fea0/src/Webform.js#L1632-L1638
# Part 2
https://github.com/formio/formio.js/blob/040e5901d977083d08b540ce27d772bcc704fea0/src/Webform.js#L1246-L1248

If there was an issue with the submission in part 1, what kind of JSON format should be returned?
Because line 1637 calls into line 1247 and results in an exception.

travist:
It is very possible that this is a bug within the showErrors function where it expects to receive a list of errors in the form.io error format which should include an array of "messages" attached to each error object. Looking at the code, I believe that you may be able to make to cooperate by changing the code to look like the following.

```
this.showErrors({messages: [`${e.statusText ? e.statusText : ''} ${e.status ? e.status : e}`]});
```

I have not tried this myself, so let me know if this does not work. Ultimately we may need to make a change to the showErrors function so that it can handle multiple formats.

bilogic:
thanks, let me try. I have another PR, possible to accept? https://github.com/formio/formio.js/pull/4854

daneformio:
Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

Issue #3970, opened 2021-04-14
[BUG] form.showError no longer accepts strings


paulfalgout:
- Formio.js version: 4.13.0

## Steps to Reproduce
1. form.showErrors('error');
2. form.showErrors(['error']);
3. 

## Expected behavior
Previous to https://github.com/formio/formio.js/commit/9d645494d4cb89b8fc5df78ae5a5bed0006504f0#diff-7855a6dd80d9df39fdfc8b5408184edab6a38ecfb49a165c448d5dfab1264a1fR1210 showError accepted a string or an array of strings

## Observed behavior
showError fails with "TypeError: Cannot create property 'formattedKeyOrPath' on string"

ccontour:
@paulfalgout I see you have a commit referenced that seems to provide a fix to this issue. If you are happy with that solution, can you provide that as a PR to this repository and we can try to get it merged in?

wag110894:
Hello @paulfalgout, 
Can you provide an example of the issue taking place?

paulfalgout:
No need for a video.  https://jsfiddle.net/5cgkuj8n/
```js
Formio.createForm(...).then(function(form) {
  // Both of these used to work and now they don't
  form.showErrors('error');
  form.showErrors(['error']);
});
```

bilogic:
@paulfalgout 

Was your issue addressed? The docs say it has to be an object but webform.js itself still uses string. This is getting quite confusing #4862

paulfalgout:
AFAIK this hasn't been addressed @bilogic

Tatsiana8:
Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #4863, opened 2022-09-16
follow the error handling of executeSubmit()

bilogic:
I believe this addresses 

https://github.com/formio/formio.js/issues/4862
https://github.com/formio/formio/issues/1474
https://github.com/formio/formio.js/issues/3970


travist:
👍

PR #4873, opened 2022-09-24
FIO-5430 Fixed datetime not being initialized when there are multiple…

antonSoftensity:
… ones on the form while translations are enabled


src/widgets/CalendarWidget.js
```diff
@@ -208,11 +208,9 @@ export default class CalendarWidget extends InputWidget {
               if (locale && locale.length >= 2 && locale !== 'en') {
                 return Formio.requireLibrary(
                   `flatpickr-${locale}`,
-                  `flatpickr-${locale}`,
+                  `${locale}`,
```
travist, line 211: So when flatpickr is added to the page, it is adding as the object named the locale?  So, to access it, it would be "en" instead of "flatpickr-en"?  If this is the case, is there any way we can change how it is added to the page, since I really believe that "en" would be problematic and prone to collisions.

travist (changes requested):

PR #4876, opened 2022-09-27
FIO-4318: fixed displaying data for select with url data src

Maria-Golomb:


src/components/select/Select.js
```diff
@@ -991,8 +991,9 @@ export default class SelectComponent extends ListComponent {
         if (this.choices && this.choices.input && this.choices.input.element) {
           this.choices.input.element.value = '';
         }
-
-        this.updateItems(null, true);
+        if (!this.itemsFromUrl ) {
+          this.updateItems(null, true);
+        }
```
TanyaGashtold, line 996: I am pretty sure that this solution causes regressions.

1) Always setting search input equal to dataValue means that select dropdown will always show only 1 option, that is equal to dataValue.
2) According to the changes, select Url will never update items on 'hideDropdown', that means that only found by search items will be shown when you open the dropdown again.

Please spend more time to investigate the issue.

TanyaGashtold: Please fix tests.

TanyaGashtold (changes requested):

lane-formio:
Issue no longer reproduceable, closing PR.

PR #4896, opened 2022-10-11
FIO-5674: fixed broken positioning of the data table on the submissio…

AlexanderYakubovskiy:


src/sass/formio.form.builder.scss
```diff
@@ -151,7 +151,7 @@
 .builder-sidebar {
   &_scroll {
     position: sticky;
-    top: 30px;
+    top: 15px;
```
travist, line 154: This is very arbitrary change. We need to discuss how this will affect non-portal versions of the builder.

travist (changes requested):

src/sass/formio.form.builder.scss
AlexanderYakubovskiy: It's just revert of this one https://github.com/formio/formio.js/pull/4855/commits/aac5b6d68cb2ed79d00e43da5149aee7d9c4f170

PR #4943, opened 2022-11-18
FIO-5627: fixed an issue where manual input does not work for date with full month format

TanyaGashtold:


src/widgets/CalendarWidget.js
```diff
@@ -477,7 +477,9 @@ export default class CalendarWidget extends InputWidget {
       }
     });
 
-    if (!this.settings.readOnly) {
+    const excludedFromMaskFormats = ['MMMM'];
```
travist, line 480: This seems like a hack.  I would bet the issue is actually within the "convertFormatToMask" method. Have you looked for the solution there?

travist (changes requested):

src/widgets/CalendarWidget.js
TanyaGashtold: Yes, the issue is that the mask is not defined correctly for the "MMMM" format. The "MMMM" format means the full name of the month, but the months have a different number of letters (May - 3, April - 5, etc.). There is no suitable mask for this case. So I decided to disable the mask for this format.

travist: So flatpickr does not allow for the setting of full month name.  I assumed that the number of letters within the month would not matter and that MMMM always meant the full name regardless of the length of the name. This still seems off here and I bet the fix should be within the convertFormatToMask function to convert the format correctly.

TanyaGashtold: According to the flatpickr documentation, it supports full month format: https://flatpickr.js.org/formatting/. The MMMM format is converted to "F" flatpickr format. When the value is selected from the calendar, it works fine. The issue appears only when entering value manually. The input mask for MMMM format is defined as 2 numbers in our code and when a user enters value (lets say, May), it is cleared immediately . So that it works, input mask is expected to be equal to letters (but what number of letters???). Defining 3 letters, only May will work, entering Apr, value will be cleared on blur, as Apr does not correspond to full month format. So there is no proper mask for full month. Once the mask is disabled, manual input starts working well for all months.

PR #4950, opened 2022-11-23
Fio 5272 use the url data source type for radio buttons and select boxes

HannaKurban:


TanyaGashtold (approved):

src/components/_classes/list/ListComponent.js
```diff
@@ -0,0 +1,226 @@
...
+          this.itemsLoadedResolve();
+          return;
+        }
+
+        let resourceUrl = this.options.formio ? this.options.formio.formsUrl : `${Formio.getProjectUrl()}/form`;
+        resourceUrl += (`/${this.component.data.resource}/submission`);
+
+        if (forceUpdate || this.additionalResourcesAvailable || !this.serverCount) {
+          try {
+            this.loadItems(resourceUrl, searchInput, this.requestHeaders);
```
TanyaGashtold, line 130: ListComponent does not have its own or prototype loadItems method. That is not correct. If someone inherits from ListComponent class and has no idea what methods it is expected to have, it will break everything. We need to add its own loadItems method that will do, at least, nothing and comment it like 'Must be implemented in child classes'.
The same is for setItems method.

TanyaGashtold (changes requested):

src/components/_classes/list/ListComponent.js
HannaKurban: Methods added. All tests passed locally

TanyaGashtold (approved):

PR #4954, opened 2022-11-25
FIO-5943 Fixed modal edit buttons for next portal

antonSoftensity:


travist:
These should not be in the renderer css. There should be no color configurations within this CSS. Please just add these to our portal css.

TanyaGashtold (changes requested):
I agree with Travis, we need to move this styles to next portal css file, this is just a next portal style issue. When we render the formio.js form outside the portal, the modal edit buttons have background color: 
image

antonSoftensity:
PR for next portal - https://github.com/formio/formio-app/pull/1344. Current PR can be closed

Issue #4905, opened 2022-10-17
[BUG] File upload (multi) - aborting upload always cancels the last one in the list.


llaurentiu:
## Environment
Please provide as many details as you can:

- Hosting type
  - [ x] Local deployment
    - Version: Any
- Formio.js version: All
- Frontend framework: All
- Browser: Any
- Browser version: Any

## Steps to Reproduce
1. Have a form with a file component, multiple upload enabled
2. Drag multiple files (eg: 5) in the file upload field. The uploads should take enough time that the user can cancel any of them.
3. Click the [x] of the eg: 3rd uploaded file

## Expected behavior
> 3rd file is removed from the upload list.
> upload of the 3rd file is aborted
## Observed behavior
> last uploaded file is canceled
> 3rd file is uploaded
> last file is canceled but still appears rendered in the list (intermediary state).

## Example
JS fiddle to reproduce [here ](https://jsfiddle.net/llaurentiu/64q7sgf9/45/)

----
Cause of the bug: the File component always registers the last file upload abort callback (overwrites the previous).

Possible solutions: 

1. Register separate abort callbacks, per instance of the upload request (eg: have it as a property of the `fileUpload` objects in the context).
2. Pass the `file, fileName` to the abortCallback. In which case the file service would have to internally keep the state of the uploads and call the right abort callback.

Solution:1 should be the way to do it.

Susccy:
@travist why is the merged fix only included in rc tags for formiojs v5 instead of a patch for v4 ?

Susccy:
@travist bump

travist:
@Susccy we will discuss adding this into the v4 renderer in our next planning meeting.

lane-formio:
Thanks for the contribution!
 This has been included in formiojs@4.18.1 as a part of our formio-enterprise@8.4.1 release.

PR #4957, opened 2022-11-29
Multi upload: abort the correct file when clicking the cancel upload button

llaurentiu:
Fixing BUG opened [here](https://github.com/formio/formio.js/issues/4905)


llaurentiu:
@travist  this is ecb

travist:
This pull request is causing our tests to fail. Please investigate these failures and re-submit your pr when you find the solution. We will need to revert this commit.

Issue #4963, opened 2022-12-02
[BUG] Select does not support non-Latin characters


ousmorez:
## Environment
Formio uses [Choicejs ](https://www.npmjs.com/package/choices.js)as an external library. As you can see the version of [Choicejs that formio uses](https://www.npmjs.com/package/@formio/choices.js) is not up to date.

We recently found the following bug in Formio that we thought was from Choicejs:
https://github.com/Choices-js/Choices/issues/1068
After fixing the bug we were talking about in the pull request below, we found that the problem still persists:
https://github.com/Choices-js/Choices/pull/1072
It looks like Formio needs to update the Choicejs dependency or make the following changes to the ChoicesWrapper.js file:

```javascript
      if (!hasActiveDropdown && !this._isTextElement && /[^\x00-\x1F]/.test(keyString) && event.key.length === 1) {//<---
        var currentValue = this.input.element.value;
        this.input.element.value = currentValue ? "".concat(currentValue).concat(event.key) : event.key;//<---
        this.showDropdown();
      } 
```
- Hosting type
  - [x] Form.io
  - [ ] Local deployment
    - Version: 
- Formio.js version: 4.14.8
- Frontend framework: JS
- Browser: Chrome
- Browser version: Latest

heather-formio:
Thank you for your report. We are reviewing the choices.js library upgrade and when we can prioritize this. Stay tuned to the changelog for the upgrade.

PR #4964, opened 2022-12-02
[BUGFIX] Select does not support non-Latin characters

ousmorez:
This is a pull request to resolve the following bug:

https://github.com/formio/formio.js/issues/4963


travist:
This is great!  I will make sure we get this in a future release.  

Thank you for your contribution!

travist:
This pull request changes the eslint format to require ES5. Please refactor and re-submit changes using ES6.

PR #4977, opened 2022-12-13
FIO-4230 Fixed label display for Button component inside DataGrid

HannaKurban:
https://formio.atlassian.net/browse/FIO-4230


src/templates/bootstrap/datagrid/html.ejs
```diff
@@ -8,7 +8,7 @@
     <tr>
       {% ctx.columns.forEach(function(col) { %}
         <th class="{{col.validate && col.validate.required ? 'field-required' : ''}}">
-          {{ col.hideLabel ? '' : ctx.t(col.label || col.title, { _userInput: true }) }}
+          {{ (col.hideLabel || (col.dataGridLabel===false && col.type === 'button')) ? '' : ctx.t(col.label || col.title, { _userInput: true }) }}
```
TanyaGashtold, line 11: As far as I know 'dataGridLabel' option is responsible for displaying component label inside dataGird rows. Why should this work different for button component? If you do not want the label to display in dataGrid header, you need just to check hideLabel option for the component inside dataGrid. 
I think we need to keep the option functionality consistent for all components. If button does not have hideLabel option, then we need to make it available when button is inside dataGrid. If 'dataGridLabel' option does nothing for button, we need to remove it. 
It might be worth changing the description of the 'Show Label in DataGrid' option specifying that it relates to label display inside each row.

TanyaGashtold (changes requested):

src/templates/bootstrap/datagrid/html.ejs
TanyaGashtold: Please fix tests

HannaKurban:
Removed changes in templates. 
Added 'hideLabel' checkbox for the Button component inside the DataGrid. 
'dataGridLabel' checkbox for the Button component has been removed because this option does nothing. 
The tooltip for 'dataGridLabel' has been changed. 
All tests passed locally

TanyaGashtold (changes requested):
Please resolve merge conflicts

TanyaGashtold (approved):

PR #4980, opened 2022-12-14
FIO-5527: Fixed an issue where value in TextArea with wysiwyg editor is not getting cleared on reset

alexandraRamanenka:


alexandraRamanenka:
image
Tests are passing locally after rebuild

PR #4986, opened 2022-12-19
FIO-5685 Data in Select when set Value as Entire Object to be set as …

HannaKurban:
…an object


src/components/select/Select.js
```diff
@@ -1672,4 +1685,11 @@ export default class SelectComponent extends ListComponent {
       super.setErrorClasses([this.refs.selectContainer], dirty, hasError, hasMessages, element);
     }
   }
+
+  beforeSubmit() {
+    if (this.isEntireObjectDisplay() && typeof this.dataValue === 'string') {
+      this.dataValue = JSON.parse(this.dataValue);
```
TanyaGashtold, line 1691: It seems to me not correct to change dataValue type (from sting to object) before submission. The value type must be constant otherwise we will not be able to use it in our conditional logic or calculated value.
From my prospective, for 'entire object' value property, the label should be a stringified object, but the value - always an object.

TanyaGashtold (changes requested):

src/components/select/Select.js
TanyaGashtold: Additionally, please fix failing tests

Issue #4494, opened 2021-12-09
[BUG] Uncaught RangeError: Invalid currency code :


naomiaro:
## Environment
Please provide as many details as you can:

![Screenshot 2021-12-08 at 4 01 34 PM](https://user-images.githubusercontent.com/35253/145310415-eb637783-a726-4c1c-850c-b9892c1bf09e.png)

- Formio.js version: 4.14

## Steps to Reproduce
Removing the currency by hitting the 'x' on the input and saving the form

![Screenshot 2021-12-08 at 3 58 57 PM](https://user-images.githubusercontent.com/35253/145310232-976ab037-c84a-43eb-ac46-5a376de4b482.png)

![Screenshot 2021-12-08 at 3 59 05 PM](https://user-images.githubusercontent.com/35253/145310237-a03532f8-267e-4914-8582-d662d5438d9b.png)

A default currency needs to be set in cases like this?

## Example

https://jsfiddle.net/9cwqetrb/2/

abaltuta:
It seems that [the check](https://github.com/formio/formio.js/blob/master/src/utils/utils.js#LL886C7-L886C7)works only for `undefined`. But if you remove the currency by hitting "x" as shown above the value gets set to an empty string.

daneformio:
Closing this thread as it is outdated. Please re-open if it is still relevant. Thank you for your contribution!

PR #4988, opened 2022-12-20
fix: falsy currency defaults to USD

abaltuta:
Currency can be an empty string value as seen in #4494 
Changed to provide a default value when passed falsy value


travist:
👍

PR #4992, opened 2022-12-22
FIO-5785 Fixed Select component submission showing id instead of temp…

antonSoftensity:
…late for URL data source


antonSoftensity:
Also, in order for Select component to display correct values in submissions with multiple values, [this](https://github.com/formio/formio.js/pull/4663/commits/f53e5e7d7f595757b9af65951ac430a58bf0fe78) commit needs to be cherry-picked to master

TanyaGashtold (approved):

PR #4999, opened 2022-12-27
FIO-6078 Hidden Select Boxes validation do not prevent submission

HannaKurban:


src/components/selectboxes/SelectBoxes.js
```diff
@@ -205,14 +205,17 @@ export default class SelectBoxesComponent extends RadioComponent {
...
       const count = Object.keys(this.validationValue).reduce((total, key) => {
         if (this.validationValue[key]) {
           total++;
         }
         return total;
       }, 0);
 
+      if (minCount && count < minCount) {
+        this.component.validate.required = true;
+      }
```
TanyaGashtold, line 218: I do not think that we need to make the component required automatically when minCount is set for it. The minCount validation sets the rules for the value when component has it. To make component required, the user should  check this option in component settings.

src/components/selectboxes/SelectBoxes.js
```diff
@@ -205,14 +205,17 @@ export default class SelectBoxesComponent extends RadioComponent {
     const maxCount = this.component.validate.maxSelectedCount;
     const isValid = this.isValid(data, dirty);
 
-    if (maxCount || minCount) {
+    if ((maxCount || minCount) && this.visible) {
```
TanyaGashtold, line 208: It might be better to use shouldSkipValidation method from parent class instead of just visibility check..

TanyaGashtold (changes requested):

src/components/selectboxes/SelectBoxes.js
HannaKurban: changed

src/components/selectboxes/SelectBoxes.js
HannaKurban: fixed

TanyaGashtold (approved):

HannaKurban:
all tests are passing locally

PR #5005, opened 2023-01-02
FIO-5511: ensure contextual row data is validated correctly for nested components

brendanbond:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-5511

## Description

**What changed?**

Previously, forms with nested components (in particular, a container component) would conditionally show correctly based on a contextual `row` property but would not validate. This was due to the `row` property not being properly tied to a container component's context. This PR attempts to alleviate this problem by setting `row` to the component's `dataValue` property (or, in other words, keys into the submission data by its own property key) so it can be validated correctly if it has custom conditionals.

Although I'm fairly certain this is a minor change, this is my first renderer ticket, so I'm nervous about "unknown unknowns" here - I've requested a review from @TanyaGashtold to make sure I'm not barking up the wrong tree.

## Dependencies

n/a

I added an automated test that covers the case of a wizard with numerous nested components, one of which is conditionally visible and is required. Previously, nextPage() would not validate the required component even if it was visible and unchecked. This test ensures that validation occurs correctly. 

## Checklist:

### DIFF
--- a/src/components/_classes/nested/NestedComponent.js
+++ b/src/components/_classes/nested/NestedComponent.js
@@ -643,6 +643,7 @@ export default class NestedComponent extends Field {
   }
 
   checkValidity(data, dirty, row, silentCheck) {
+    row = this.dataValue || row;
     if (!this.checkCondition(row, data)) {
       this.setCustomValidity('');
       return true;
--- a/src/components/_classes/nested/NestedComponent.js
+++ b/src/components/_classes/nested/NestedComponent.js
@@ -642,17 +642,20 @@ export default class NestedComponent extends Field {
     );
   }
 
+  checkChildComponentsValidity(data, dirty, row, silentCheck, isParentValid) {
+    return this.getComponents().reduce(
+      (check, comp) => comp.checkValidity(data, dirty, row, silentCheck) && check,
+      isParentValid
+    );
+  }
+
   checkValidity(data, dirty, row, silentCheck) {
-    row = this.dataValue || row;
     if (!this.checkCondition(row, data)) {
       this.setCustomValidity('');
       return true;
     }
 
-    const isValid = this.getComponents().reduce(
-      (check, comp) => comp.checkValidity(data, dirty, row, silentCheck) && check,
-      super.checkValidity(data, dirty, row, silentCheck)
-    );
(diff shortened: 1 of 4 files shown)


src/components/_classes/nested/NestedComponent.js
```diff
@@ -643,6 +643,7 @@ export default class NestedComponent extends Field {
   }
 
   checkValidity(data, dirty, row, silentCheck) {
+    row = this.dataValue || row;
```
TanyaGashtold, line 646: I am afraid, this change can create some regressions in validations of dataGrid/editGrid/containers/custom components with rows where row is not equal to the component dataValue.
Addionally:
1. Row, passed as a parameter, is expected to have first priority. In your case component dataValue is prioritized.
2. logically, row means a context (like environment) where the component, that is being checked, exists. The dataValue is the value of the component itself, not the context.

Checking the 5511 ticket, I see that the issue is in the Container Component. When we are in checkValidity, row value is correct for the component itself but not correct for its children as row for children is expected to be equal to the parent container dataValue. This is very specific to Container and is not related to some other nested components like fieldSet or panel. So, I would prefer to move the nested components validity check to separate method in NestedComponent and then to redefine this method for ContainerComponent and pass container dataValue as a row to it.
Something like this:

**Nested Component**

```
  checkChildComponentsValidity(data, dirty, row, silentCheck, isParentValid) {
    return this.getComponents().reduce(
      (check, comp) => comp.checkValidity(data, dirty, row, silentCheck) && check,
      isParentValid
    );
  }
```

```
  checkValidity(data, dirty, row, silentCheck) {
    if (!this.checkCondition(row, data)) {
      this.setCustomValidity('');
      return true;
    }
    const isValid = this.checkChildComponentsValidity(data, dirty, row, silentCheck, super.checkValidity(data, dirty, row, silentCheck));
    this.checkModal(isValid, dirty);
    return isValid;
  }
```

**ContainerComponent**

```
  checkChildComponentsValidity(data, dirty, row, silentCheck, isParentValid) {
    return super.checkChildComponentsValidity(data, dirty, this.dataValue, silentCheck, isParentValid);
  }
```

TanyaGashtold (changes requested):

brendanbond:
Awesome, thanks for the feedback! I'll make these changes now

TanyaGashtold (approved):

PR #5017, opened 2023-01-16
Should allow quotes in options

iconnor:
Initial draft PR to demonstrate the problem by adding failing tests.

## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-XXXX

## Description

**What changed?**

*Previously, formio.js ... This PR replaces this behavior by ...*

*Although there were many potential solutions such as ..., [my solution] was best because ...*

## Dependencies

## Checklist:

### DIFF
--- a/src/components/select/Select.unit.js
+++ b/src/components/select/Select.unit.js
@@ -167,6 +167,29 @@ describe('Select Component', () => {
     });
   });
 
+  it('should allow quotes in options', function(done) {
+    comp5.template = '<span>{{ item.label }}</span>';
+    comp5.uniqueOptions = true;
+    Harness.testCreate(SelectComponent, comp5).then((component) => {
+      component.setItems([{
+        'label': 'Label "1"',
+        'value': 'value1'
+      }, {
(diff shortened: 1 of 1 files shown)


brendanbond:
Your tests fail because your strings don't match the label value after it's made ready for display (i.e. after the template is interpolated with the string value of the label). Try updating them to...
```
assert.equal(component.selectOptions[0].label, '<span>Label "1"</span>');
assert.equal(component.selectOptions[1].label, '<span>Label "2"</span>');
assert.equal(component.selectOptions[2].label, "<span>Label \'3\'</span>");
```
...and they pass. Unless I'm missing something about what your requirements are here?

iconnor:
I might not be testing the right component. I might need to dig into how the options get rendered rather than formed. We can't use values that contain ' unless we hack options to use double quotes (e.g. ’<option value = "Test's Val"...‘) but then we can't use double quotes. I was hoping escapes might solve the problem but they also failed. I added this test to see if I could have escaped quotes and see if that solved our problem as we could escape and quotes needed.

PR #5024, opened 2023-01-19
FIO-5566: Fixes an issue where deprication warning appears in the console when using Day component with dateFirst setting

alexandraRamanenka:
https://formio.atlassian.net/browse/FIO-5566
There was a change done 2 years ago that was rewriting minDate and maxDate settings from the ISO format (yyyy-mm-dd) to another (dd/mm/yyyy) when the dayFirst setting was turned on. But due to the other changes in min/max date validation, this fix caused other issues where a deprication warning was shown and min/max date setting were read wrongly (eg. 2000-03-01 is supposted to be 1 March, 2000, but it was read as 3 January, 2000 instead) that was leading to min/max date validations were not always working properly.

The test written for rewriting min/maxDate setting to other format was replaced with a test checking that ensures min/max date validations works properly


TanyaGashtold (changes requested):
Please fix failing tests

TanyaGashtold (approved):

PR #5049, opened 2023-02-01
FIO-3884 Improved column comparison by excluding the id.

michealharrington:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-3884
https://github.com/formio/formio.js/pull/4497

## Description

**What changed?**

I have updated how we compare columns to determine if the datagrid needs to be updated.
The previous comparison would compare every field including the auto-generated "id" field.
Through debugging, I've found that the datagrid was being re-drawn unnecessarily because the only field in the columns that were different were the "id"s.
I've updated the logic to ignore the id field when comparing.

I kept the majority of the previous solution, simply skipping the comparison of the id to avoid unnecessary rebuilding of the datagrids.

I tested it locally with my use case and this reduced the re-drawing of the grid.

## Checklist:

### DIFF
--- a/src/components/datagrid/DataGrid.js
+++ b/src/components/datagrid/DataGrid.js
@@ -622,7 +622,14 @@ export default class DataGridComponent extends NestedArrayComponent {
 
           if (col.component.logic && firstRowCheck) {
             const compIndex = _.findIndex(this.columns, ['key', key]);
-            if (!_.isEqual(this.columns[compIndex], col.component)) {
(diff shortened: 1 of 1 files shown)


michealharrington:
Is there anything else I need to do to get this bugfix merged? I need this fixed before I can update to the latest form.io version.

brendanbond:
@michealharrington Thanks for submitting this! I'll try and find some time this weekend to review.

michealharrington:
Any update on this getting merged? It is blocking me from updating to the latest version of formio.

travist:
👍 

Will be included in version 5.x.  Thank you!

PR #5084, opened 2023-03-03
FIO-5086: Date/Time component allows to submit any string and the server-side validation is not triggered

alexandraRamanenka:
https://formio.atlassian.net/browse/FIO-5086

Despite the fact we have the 'date' validation, it was never triggered since if we pass any value to the check function except of the string 'Ivalid date', it returns true. And even if the value passed to the function was parsed with new Date(value), it still was returning true because new Date(...) returns an object if can't convert value to a valid date.

### DIFF
--- a/src/validator/Validator.js
+++ b/src/validator/Validator.js
@@ -538,7 +538,16 @@ class ValidationChecker {
           });
         },
         check(component, setting, value) {
-          return (value !== 'Invalid date');
+          if (!value || value instanceof Date) {
+            return true;
+          }
+          if (value === 'Invalid date' || value === 'Invalid Date') {
+            return false;
+          }
+          if (typeof value === 'string') {
+            value = new Date(value);
+          }
+          return value.toString() !== 'Invalid Date';
         }
       },
       day: {


src/validator/Validator.js
```diff
@@ -538,7 +538,16 @@ class ValidationChecker {
           });
         },
         check(component, setting, value) {
-          return (value !== 'Invalid date');
+          if (!value || value instanceof Date) {
```
brendanbond, line 541: This looks good, although it has one problem: because JavaScript is weird, the following code is valid:
```
try {
  const myDate = new Date('hello, world!');
  console.log(myDate instanceof Date); // logs true
  console.log(typeof myDate === 'object'); // logs true
  console.log(myDate); // logs 'Invalid Date'
} catch (err) {
  // this code is unreachable, because the Date constructor will not throw an error
}
```

Because of this behavior, I think we need to move the `instanceof Date` check to after we check the `toString()` value to ensure that we're handling the case of `myDate instanceof Date === true && myDate.toString() === 'Invalid Date'`

brendanbond (changes requested):
See comment re: dating is hard

PR #5090, opened 2023-03-10
FIO-6358-Save Progress as Draft and not properly Re-rendering the form with conditional logic

Alexandr-Lonski:
## Link to Jira Ticket

https://formio.atlassian.net/jira/software/c/projects/FIO/issues/FIO-6358

## Description

Added functionality to wizard for save data before submit as draft.

I added automated tests.

## Checklist:


TanyaGashtold (changes requested):
I do not see a reason why we need to store wizard data in localStorage. It seems to be made not to lose data on page reload, but it is expected that unsaved data is lost. I think FIO-6358 needs more investigation.

Alexandr-Lonski:
> I do not see a reason why we need to store wizard data in localStorage. It seems to be made not to lose data on page reload, but it is expected that unsaved data is lost. I think FIO-6358 needs more investigation.
 
Ok, Thank you! I will continue to investigate and add a few clarifying questions to the ticket for the customer.

PR #5092, opened 2023-03-13
FIO-4809: Fixes an issue where components outside panels are rendered on each Wizard page

alexandraRamanenka:
https://formio.atlassian.net/browse/FIO-4809


alexandraRamanenka:
image
Tests are passign locally

heather-formio:
I am not comfortable with this change as it is common for users to want components and the associated data to reside outside of panels (therefore outside of the wizard UX) but still as part of the Submission JSON.

alexandraRamanenka:
> I am not comfortable with this change as it is common for users to want components and the associated data to reside outside of panels (therefore outside of the wizard UX) but still as part of the Submission JSON.

Not sure that this is connected. Currently, if we create a Wizard form, there is no way to add components that are not inside od the panel (wizard's page). The issue was that if we have a siple Webform with some panels, and then we want to change the Display type to the Wizard, it will make each panel of the form a wizard's page. At the same time, there was a bug that if we have some components outside of any panel, the builder adds it to each wizard's page and those comoponent do not work as expected. And because there is no obviouse solution for how to deal with components that are outside of any panel when converting webform to a wizard, I make the builder remove them, but asking a confirmation first. So if user don't want them to be removed, he can simply move them to any panel he wants.

heather-formio:
> > I am not comfortable with this change as it is common for users to want components and the associated data to reside outside of panels (therefore outside of the wizard UX) but still as part of the Submission JSON.
> 
> Not sure that this is connected. Currently, if we create a Wizard form, there is no way to add components that are not inside od the panel (wizard's page). The issue was that if we have a siple Webform with some panels, and then we want to change the Display type to the Wizard, it will make each panel of the form a wizard's page. At the same time, there was a bug that if we have some components outside of any panel, the builder adds it to each wizard's page and those comoponent do not work as expected. And because there is no obviouse solution for how to deal with components that are outside of any panel when converting webform to a wizard, I make the builder remove them, but asking a confirmation first. So if user don't want them to be removed, he can simply move them to any panel he wants.

The Problem with this is there are strong use cases for retaining components outside of panels, that are intentionally showing on every page as a result of them being outside the panel. Therefore, I don't understand what we are fixing / disagree with that there should even be a workflow where components are removed because of a form type switch. This could result in a lot of lost work if revisions aren't on and a user doesn't realize what the implications are.

PR #5151, opened 2023-05-04
FIO-6760:fixed an issue where in select component when use resource multiple values data isn't rendered correctly

Alexandr-Lonski:
## Link to Jira Ticket

https://formio.atlassian.net/jira/software/c/projects/FIO/issues/FIO-6760

## Description

Fixed issue, when use resource with multiple values, data isn't rendered correctly

I tested locally and added automated tests to cover this case

## Checklist:

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -198,7 +198,7 @@ export default class SelectComponent extends ListComponent {
   }
 
   selectValueAndLabel(data) {
-    const value = this.getOptionValue((this.isEntireObjectDisplay() || !this.itemValue(data)) ? data : this.itemValue(data));
+    const value = this.getOptionValue((this.isEntireObjectDisplay() && !this.itemValue(data)) ? data : this.itemValue(data));
     return {
       value,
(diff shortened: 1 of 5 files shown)


src/components/select/Select.js
```diff
@@ -198,7 +198,7 @@ export default class SelectComponent extends ListComponent {
   }
 
   selectValueAndLabel(data) {
-    const value = this.getOptionValue((this.isEntireObjectDisplay() && !this.itemValue(data)) ? data : this.itemValue(data));
+    const value = this.getOptionValue((this.isEntireObjectDisplay() || !this.itemValue(data)) ? data : this.itemValue(data));
```
TanyaGashtold, line 201: I am afraid that can effect the select with EntireObject property value. Please discuss it with Hanna.

TanyaGashtold (changes requested):

src/components/select/Select.js
Alexandr-Lonski: Thanks for the comment I found another solution for my issue

TanyaGashtold (approved):

PR #5186, opened 2023-05-23
FIO-6632: formiojs-to-run-npm-run-release

ryanformio:
…build is called

## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-6632

## Description

**What changed?**

Added npm run release to the npm run build script.

This should be the easiest way to have the command run with needing changes to the build script to accomodate.

It has not been tested against a build yet but will be tried on this next one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "lint"
   ],
   "scripts": {
-    "build": "esdoc && gulp build",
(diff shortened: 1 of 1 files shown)


travist:
We cannot release when we build.

PR #5221, opened 2023-06-13
FIO-5967: fixed issue with empty data for html 5 select component

RomanLetsuk:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-5967

automated tests

## Checklist:


src/components/select/Select.js
```diff
@@ -421,7 +421,7 @@ export default class SelectComponent extends ListComponent {
     }
 
     // Add the value options.
-    if (!fromSearch) {
+    if (!fromSearch || this.component.widget === 'html5') {
```
travist, line 424: Why is ```fromSearch``` true for html5 widget settings?  Wouldn't that be the the source of the real problem instead of fixing it with a bandaid?

travist (changes requested):

PR #5292, opened 2023-07-28
Refactored the embed script so that it can be imported within an application using createForm.

travist:
## Description

**What changed?**

This change makes the "embed" capability more accessible for libraries and applications. It makes it so that you can import the "embed" feature into your application using the following code.

```js
import { Formio } from '@formio/js/embed';
Formio.createForm(document.getElementById('formio'), 'https://examples.form.io/example');
```

This works just like the other "createForm", but instead of loading the whole renderer codebase, it instead only requires a 7kb embed renderer which places a loader in the provided DOM element while it lazy-loads the whole renderer library. This makes it so that you can build your applications without requiring the import of a 1MB+ renderer as part of your application bundle size, but rather, only import 7kb embed code which will dynamically render the form in your application. This makes the application load immediately and the end user will see a loader where the form will eventually render (hence a much better UX improvement).

Because our renderer size is large, it causes applications built on top of the Form.io renderer library to have a large bundle size. This can cause the "white screen of death" when the application first loads while it downloads all the vendor files. This mitigates this problem by only requiring the use of a 7kb embed file which is responsible for dynamically pulling in the renderer dependencies and render the form once they are loaded.

I added an example to the examples application on how this works.

## Checklist:


brendanbond (approved):
Looks awesome!

PR #5331, opened 2023-08-31
FIO-7082: Moved Wizard Breadcrumbs Type to form settings

alexandraRamanenka:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7082

## Description

Removed Condensed type cause it does nothing from the time when the feature was added. Breadcrumbs type select dropdown is moved to the form settings page instead. For the old forms that have breadcrumb type set per page everything will work the same. 
I also had to pass form schema from the form builder to the evalContext of the editForm to prevent showing settings that should not be available when breadcrumbs type is 'hidden'.

## Dependencies

https://github.com/formio/formio-app/pull/1494

Manually

## Checklist:

### DIFF
--- a/src/components/panel/editForm/Panel.edit.display.js
+++ b/src/components/panel/editForm/Panel.edit.display.js
@@ -76,28 +76,23 @@ export default [
     input: false,
     components: [
       {
-        type: 'select',
-        input: true,
-        label: 'Breadcrumb Type',
-        key: 'breadcrumb',
-        dataSrc: 'values',
-        data: {
-          values: [
-            { label: 'Default', value: 'default' },
-            { label: 'Condensed', value: 'condensed' },
-            { label: 'Hidden', value: 'none' },
-          ]
-        }
-      },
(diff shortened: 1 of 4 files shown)


src/components/panel/editForm/Panel.edit.display.js
```diff
@@ -75,29 +75,24 @@ export default [
...
         }
+        // conditional: {
+        //   json: {
+        //     'and': [
+        //       { '!==': [{ var: 'data.breadcrumb' }, 'none'] },
+        //       { '!==': [{ var: 'buildingForm.settings.wizardBreadcrumbsType' }, 'none'] }
+        //     ]
+        //   }
+        // },
       },
```
TanyaGashtold, line 96: Why should we keep the commented code? Please remove it.

TanyaGashtold (changes requested):

TanyaGashtold (approved):

PR #5337, opened 2023-09-07
FIO-6859: update-s3-to-accept-headers-from-signer-presign

ryanformio:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-6859

## Description

Loops over the headers that have been sent from server at src/storage/s3/aws.js signed.presigned()
image

These headers are coming directly from the `presigned.headers` WITH the presignedUrl from the server that should be necessary to make the request.
image
image

### Pull requests
- https://github.com/formio/formio-server/pull/1336

Refer to https://github.com/formio/formio-server/pull/1336, I had made this small change to the formio.js node_modules in formio-app when running the portal locally.

## Checklist:

### DIFF
--- a/src/providers/storage/s3.js
+++ b/src/providers/storage/s3.js
@@ -5,7 +5,7 @@ const s3 = (formio) => ({
       response.data.fileName = fileName;
       response.data.key = XHR.path([response.data.key, dir, fileName]);
       if (response.signed) {
-        xhr.openAndSetHeaders('PUT', response.signed.url);
+        xhr.openAndSetHeaders('PUT', response.signed);
         Object.keys(response.data.headers).forEach(key => {
           xhr.setRequestHeader(key, response.data.headers[key]);
         });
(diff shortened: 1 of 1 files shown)


src/providers/storage/s3.js
```diff
@@ -5,8 +5,10 @@ const s3 = (formio) => ({
       response.data.fileName = fileName;
       response.data.key = XHR.path([response.data.key, dir, fileName]);
       if (response.signed) {
-        xhr.openAndSetHeaders('PUT', response.signed);
```
travist, line 8: How did this ever work if it used to pass in an OBJECT for the PUT request?

ryanformio: You're correct this needed to be updated based on a PR change request on [formio-server](https://github.com/formio/formio-server/pull/1336#discussion_r1333648380) Brendan and I recently discussed, as a result formio.js needed to be updated. 

It was not updated. Have applied requested change as a result of that discussion.
image

PR #5353, opened 2023-09-26
Fix broken file exclusion patterns

thetrevdev:
## Description

**What changed?**
Support filePatterns that include exclusions such as `'!.exe'`.

Looks like this refactor broke it:
https://github.com/formio/formio.js/commit/1c3fe1f8ee776fec12e89dee563023c646d7eb66#diff-88fcc2d2f31515efb91047d6564cc78e83922d7d795e776c2523d7c94f844e6dL450

The previous refactor added a negation to these results so any exclusion pattern will always fail.

The exclusion regular expressions are a negative lookahead for the exclusion.
Each exclusion pattern matches a string that does not contain the exclusion.

**Screenshot**
Example of it currently broken:
image

## Checklist:

### DIFF
--- a/src/components/file/File.js
+++ b/src/components/file/File.js
@@ -622,8 +622,8 @@ export default class FileComponent extends Field {
     }
     valid = pattern.excludes.reduce((result, excludePattern) => {
       const exclude = new RegExp(excludePattern, 'i');
-      return result && (_.isNil(file.type) || !exclude.test(file.type)) &&
-        (_.isNil(file.name) || !exclude.test(file.name));
+      return result && (_.isNil(file.type) || exclude.test(file.type)) &&
+        (_.isNil(file.name) || exclude.test(file.name));
     }, valid);
     return valid;
   }


travist:
This seems to change the logic of the exclude check. Isn't this allowing files that are in the exclusion list? I think the solution here is to change the exclude pattern to allow the file you wish to upload.

thetrevdev:
> This seems to change the logic of the exclude check. Isn't this allowing files that are in the exclusion list? I think the solution here is to change the exclude pattern to allow the file you wish to upload.

The exclusion patterns are actually patterns that pass the regex check if they do not include the excluded pattern.  
so each exclude pattern is equivalent to `pass if the string doesn't include this text`.  That was how it originally worked in formio and ng-file-upload where the code was copied from.  I linked to the refactor where this broke.  They were just refactoring it to use a forEach instead of a while loop and they added the exclamation then.

before [broken refactor](https://github.com/formio/formio.js/commit/1c3fe1f8ee776fec12e89dee563023c646d7eb66#diff-88fcc2d2f31515efb91047d6564cc78e83922d7d795e776c2523d7c94f844e6dL450)
```  
while (len--) {
      const exclude = new RegExp(pattern.excludes[len], 'i');
      valid = valid && (file.type == null || exclude.test(file.type)) &&
        (file.name == null || exclude.test(file.name));
    }
```

thetrevdev:
Here is an example showing the broken current state and the fix:

https://jsfiddle.net/r5was27e/1/

travist:
Sounds good. I will include this in the 5.x renderer build.

PR #5366, opened 2023-10-10
FIO-7355 Fixed issue with HTML5 select flickering on initial click

antonSoftensity:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7355

## Description

Removed handler that called update() function for HTML5 select on focus, because there is no mechanisms to change Select options data after the element was already initialized, and any custom/url and other dynamic data gets set on init

Tested locally

image

## Checklist:

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -943,6 +943,11 @@ export default class SelectComponent extends ListComponent {
       }
 
       this.focusableElement = input;
+
+      if (this.component.dataSrc === 'custom') {
+        this.addEventListener(input, 'focus', () => this.updateCustomItems());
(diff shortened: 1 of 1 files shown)


TanyaGashtold (changes requested):
I am afraid that this may break select components (html5) with custom data source (e.g. if the options are calculated and depend on the values of other form component). Can you please check this use case.

antonSoftensity:
Good catch, I reworked the fix so that the values are updated when the Select data source is set to 'Custom' without unnecessary update on the first click

TanyaGashtold (approved):

PR #5393, opened 2023-11-03
FIO-7184 Fixed showing incorrect value for DateTime and Time componen…

antonSoftensity:
…ts with multiple value enabled inside of the DataTable

## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7184

## Description

Previously, DateTime and Time components didn't have support for multiple values inside of getValueAsString methods, which caused an error when the value was formatted by moment library. Added a case to format each value one by one.

Tested locally

## Checklist:

### DIFF
--- a/src/components/datetime/DateTime.js
+++ b/src/components/datetime/DateTime.js
@@ -204,9 +204,13 @@ export default class DateTimeComponent extends Input {
     const timezone = this.timezone;
     if (value && !this.attached && timezone) {
       if (Array.isArray(value) && this.component.multiple) {
-        return value.map((item) => _.trim(FormioUtils.momentDate(item, format, timezone).format(format))).join(', ');
+        return value.map(item => _.trim(FormioUtils.momentDate(item, format, timezone).format(format))).join(', ');
       }
       return _.trim(FormioUtils.momentDate(value, format, timezone).format(format));
+    }
+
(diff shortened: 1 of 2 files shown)


src/components/datetime/DateTime.js
```diff
@@ -203,6 +203,9 @@ export default class DateTimeComponent extends Input {
     format += format.match(/z$/) ? '' : ' z';
     const timezone = this.timezone;
     if (value && !this.attached && timezone) {
+      if (Array.isArray(value) && this.component.multiple) {
+        return value.map((item) => _.trim(FormioUtils.momentDate(item, format, timezone).format(format))).join(', ');
+      }
       return _.trim(FormioUtils.momentDate(value, format, timezone).format(format));
     }
     return (value ? _.trim(moment(value).format(format)) : value) || '';
```
TanyaGashtold, line 211: What if timezone is undefined and component is multiple? The line 207 is skipped and we go to line 211 that does not expect array value.
It also would be nice to separate repeated code in some new method. I mean lines 207 and 209. The same is for Time component.

TanyaGashtold (changes requested):

src/components/datetime/DateTime.js
antonSoftensity: I don't think that the repeated code should be separated in that case, because it is specific for that particular function.

PR #5405, opened 2023-11-15
FIO-7481: handle 504 response from service worker in offline mode for file request

roma-formio:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7481

## Description

Added condition to check response status for file upload requests. More information about 504 status from the service worker can be found on angular documentation [Service worker requests when the server can't be reached](https://angular.io/guide/service-worker-devops#service-worker-requests-when-the-server-cant-be-reached).
Although there were many potential solutions:
1. Check by 504 status code
2. Include 'ngsw-bypass' header in a specific request to let the service worker know that he should skip this request and allow to do it directly to the server. (I don't think it's the best idea to include specific header for angular in our core library)
3. Extend our 'fix-sw' gulp task in the form manager to check the request URL and skip service worker safe fetch when our request points to any storage.

I've chosen this solution because we already have a check by 504 status code to perform offline operations.

Manual testing.

## Checklist:

### DIFF
--- a/src/providers/storage/xhr.js
+++ b/src/providers/storage/xhr.js
@@ -49,6 +49,12 @@ const XHR = {
       throw err;
     }
     if (!response.ok) {
+      if (response.status === 504) {
+        const error = new Error('Network request failed');
+        error.networkError = true;
+        throw error;
+      }
+
       const message = await response.text();
       throw new Error(message || 'Unable to sign file.');
     }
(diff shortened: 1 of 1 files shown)


brendanbond (changes requested):
I skimmed the documentation you linked, but I'm not sure I fully understand how re-throwing a 504 enables the file component to do it's work in the offline plugin; can we discuss in dev support?

EDIT: Thanks @roma-formio - I didn't see your comment in the Jira ticket that said this already works as expected except in certain scenarios.

brendanbond (approved):

PR #5418, opened 2023-12-04
FIO-7544: Fixes an issue where scripts inside HTML component will be executed during interpolation

alexandraRamanenka:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7544

## Description

**What changed?**

Previously, formio.js was sanitizing the HTML component's content at the end of get content(), but since interpolate() method uses translateHTMLTemplate utility function that creates a DOM element and sets a content inside it, we need to sanitize the content before it goes there. 

## Dependencies

## Checklist:

### DIFF
--- a/src/components/html/HTML.js
+++ b/src/components/html/HTML.js
@@ -26,14 +26,6 @@ export default class HTMLComponent extends Component {
     };
   }
 
-  constructor(...args) {
-    const alert =  window.alert;
-    window.alert = (msg) => {
-      return alert(msg);
-    };
-    super(...args);
-  }
-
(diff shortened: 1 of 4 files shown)


src/components/html/HTML.js
```diff
@@ -26,6 +26,14 @@ export default class HTMLComponent extends Component {
     };
   }
 
+  constructor(...args) {
+    const alert =  window.alert;
+    window.alert = (msg) => {
+      return alert(msg);
+    };
+    super(...args);
```
TanyaGashtold, line 34: Why do we need this?

Issue #5420, opened 2023-12-05
EditGrid not showing view icons in readOnly mode


llaurentiu:
For edit grids, we are using a custom template for the edit grid rows template so that, if the grid is read only, we show a different icon - iconClass('eye'). Clicking this icon opens the clicked row so that we can view the row details (grid can be complex so not everything is visible in the grid columns). See the row template below.

The html template for the button makes use of the same 'editRow' event that is attached to the button. Using this class is the only way to attach the event.

`...
{ % if (instance.options.readOnly) { % }
    <div class="btn-group pull-right">

      //editRow is the class to which the EditGrid component attaches the click event.
      <div class="btn btn-default btn-light btn-sm editRow">

        <i class="{{ iconClass('eye') }}"></i>

      </div>

    </div>
    ...
`

This was working fine, prior to this change https://github.com/formio/formio.js/pull/4850 - which removes the row events completely if grid is readOnly.

Instead of creating a new event for viewing the data (which is not backward compatible) - we should just amend the new condition to be applied for the PDF converter.

PR #4850, opened 2022-09-06
FIO-5303 fixed Edit and Delete buttons

HannaKurban:
## Link to Jira Ticket (if applicable)

https://formio.atlassian.net/browse/FIO-5303

## Description

*Removed the Edit and Delete Buttons for the EditGrid on PDF download page*

## Dependencies

*no*

## How has this PR been tested?

*all tests pass locally*

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] My changes generate no new warnings
- [ ] My changes include tests that prove my fix is effective (or that my feature works as intended)
- [x] New and existing unit/integration tests pass locally with my changes
- [ ] Any dependent changes have corresponding PRs that are listed above

PR #5421, opened 2023-12-05
Fix edit grid events in read only mode

llaurentiu:
## Link to Jira Ticket

Fixing issue opened here: https://github.com/formio/formio.js/issues/5420

## Description

Revert to adding events in read only mode and only removing them in pdf rendering mode, which seems to have been the intention in the original MR: https://github.com/formio/formio.js/pull/4850 

This solution ensures backward compatibility.

Although there were other solutions, such as:
* Firing a different event for viewing the grid row (eg: viewRow instead of editRow).


travist:
👍

PR #5425, opened 2023-12-06
FIO-7637: add catch block to subform submission loading

brendanbond:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7637

## Description

A customer had an urgent production issue - their servers were crashing with unhandled promise exceptions. Upon investigation, the server-side formio.js renderer is attempting to load subform submissions from a baseUrl that is not set, ending in a "Invalid Alias" error. Since [Formio.request](https://github.com/formio/core/blob/a60896ce96c7a7b4d759a4a3bae5ad28a6b2f003/src/sdk/Formio.ts#L1561) rejects on 4xx responses, the server will crash because NodeJS expects you to handle uncaught Promise exceptions.

An alternative might have been to examine why Formio.request rejects on 4xx responses rather than delegating handling non network errors to the caller (a lá the fetch web api), but due to the urgency of this ticket for the customer in question and the potential disruption of a change to Formio.request, this seems like the most suitable solution for the moment. 

## Dependencies

n/a

n/a

## Checklist:


travist:
👍

PR #5429, opened 2023-12-07
FIO-7514: fixed an isse where new simple conditionals do not work when condition is based on the value of resource select with object value

TanyaGashtold:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7514

## Description

**What changed?**

When the resource select with object value (select with reference, select resource without value property or select resource with 'entire object' value property) is used in the new conditional logic, we suppose that user expects the equality of labels of the value in condition and the actual value of the select component. This PRs makes so the select component label data is saved as a value for the condition and than compared with the label data of the actual select value.

Discussed with Travis as best solution for this use case.

Manually + added tests
All tests passed locally

## Checklist:

### DIFF
--- a/src/Webform.unit.js
+++ b/src/Webform.unit.js
@@ -35,6 +35,7 @@ import {
   formWithCollapsedPanel,
   formWithCustomFormatDate,
   tooltipActivateCheckbox,
+  formWithObjectValueSelect
 } from '../test/formtest';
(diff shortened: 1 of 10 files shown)


lane-formio:
Can you resolve the merge conflict and I believe tests should pass now with master pulled in.

brendanbond (approved):
LGTM

PR #5438, opened 2023-12-22
Pass options to the deleteFile method for file components with URL storage

bmcbride:
**What changed?**

*Previously, formio.js did not pass http request options to the `deleteFile` method for file components with URL storage.  This PR replaces this behavior by passing the options to the `deleteFile` method, allowing users to better secure their files by requiring authorization headers.*


travist:
Thank you for the contribution. I am happy to include your change, but was curious about reverse compatibility?  With this change, if someone is configured before this is made, will it still function the way it did previously?

travist:
I reviewed this and it looks good to me. I can also see this is reverse compatible so we will pull this into the 5.x renderer. Thank you for your contribution.

PR #5463, opened 2024-01-18
Wizard: Keep current page after language is changed

akgagnat:
## Link to Jira Ticket

N/A

## Description

**What changed?**

When rendering a form with display 'wizard' the current page will now be kept after language is changed.

After upgrading to the most recent version of formio.js we saw that when a form was rendered with display 'wizard' the current page was reset to zero after changing the language. Prior to https://github.com/formio/formio.js/pull/5176 (which changed the implementation of 'set language' to invoke rebuild instead of redraw) this did not happen.

This is a breaking change if someone expects the form with display 'wizard' to reset current page to zero when the form is rebuilt.

## Dependencies

N/A

Unit test, and in my team's application using formio.js.

## Checklist:

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -1013,7 +1013,7 @@ export default class Wizard extends Webform {
   rebuild() {
     const currentPage = this.page;
     const setCurrentPage = () => this.setPage(currentPage);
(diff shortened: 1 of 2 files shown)


travist:
Looks great!   Thank you for your contribution!

PR #5488, opened 2024-01-31
FIO-7808: fixed an issue where conditional components inside conditional components lose their value after submission when they are on the same level with editGrid

TanyaGashtold:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7808

## Description

**What changed?**

Added check that allows the conditions check of the parent component in setValue of the editGrid on the server side only in case when the parent component of the parent component is visible. 

When the conditions check for editGrid parent component is run inside setValue and the parent component of the parent component is not visible, it unsets all values of conditional components as they are considered as invisible. So it appears that by the moment when the form conditions check starts, all conditional values on the same level with editGrid are already lost.

Manually + added tests

## Checklist:


travist:
Why is this behavior different on client vs. server?  I am confused why there are checks for server execution in this fix.

TanyaGashtold:
@travist , the issue is reproducible when the server option is passed. So, the values of conditional components disappear when we validate the submission on the server side. This is becouse in the clearOnHide (https://github.com/formio/formio.js/blob/4bcb87e1a1da46b178d32ae690e545e4e9154f66/src/components/_classes/component/Component.js#L2189) we have a check that removes the value from component when the component is not visible and the server option is passed.

HannaKurban (approved):

PR #5535, opened 2024-03-13
FIO-8040: use core utils for evaluation instead of local ones

AlexeyNikipelau:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8040

## Description

Previously, formio.js was using its own util function set for evalutions context. Now it's usng core functions.

## Dependencies

https://github.com/formio/core/pull/54

manually

## Checklist:

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -16,7 +16,7 @@ const { fetch } = fetchPonyfill({
   Promise: Promise
 });
 
-export * from Utils;
+export * from '@formio/core/lib/utils/formUtil';
 
 // Configure JsonLogic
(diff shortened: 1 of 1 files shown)


brendanbond (changes requested):
This doesn't work because @formio/core doesn't expose that path via the "exports" param in its package.json, you can confirm by just running the tests 

my vote is to ditch the "exports" field in both @formio/js and @formio/core, it doesn't do much for us (e.g. doesn't tree shake) and just makes it harder to use

travist:
I did this here. https://github.com/formio/formio.js/pull/5576/commits/0c43678c488435036222eb7f6b38b4b44d9684a5

PR #5551, opened 2024-04-01
FIO-8141 Fixed error for nested forms in deployed in jsfiddle

HannaKurban:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8141

## Description

*Previously, when loading a Nested Form component via jsfiddle, an incorrect request was sent to load the child form. As a result, error 400 occurred and the form was not displayed. This has been fixed by adding an additional check for the formSrc property*

## Dependencies

*no*

*locally with using JSFiddle*

## Checklist:

### DIFF
--- a/src/components/form/Form.js
+++ b/src/components/form/Form.js
@@ -90,7 +90,17 @@ export default class FormComponent extends Component {
     if (!this.formSrc && this.options.formio) {
       const rootSrc = this.options.formio.formsUrl;
       if (this.component.form && isMongoId(this.component.form)) {
-        this.formSrc = `${rootSrc}/${this.component.form}`;
+        if (
+          this.parent.form?.project &&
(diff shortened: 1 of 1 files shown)


src/components/form/Form.js
```diff
@@ -90,7 +90,17 @@ export default class FormComponent extends Component {
     if (!this.formSrc && this.options.formio) {
       const rootSrc = this.options.formio.formsUrl;
       if (this.component.form && isMongoId(this.component.form)) {
-        this.formSrc = `${rootSrc}/${this.component.form}`;
+        if (
```
travist, line 93: This change seems very specific and worries me that this is a band-aid for a problem that would be fixed elsewhere. Let's talk through this in Dev Support.

travist (changes requested):

PR #5612, opened 2024-05-23
FIO-8389: Fix EditGrid with Empty/Is Not Empty conditional logic in there

mikekotikov:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8389

## Description

**What changed?**

Previously, formio.js would fail to call 'isEmpty' method if it doesn't exist. This PR adds a check to make sure it exists, before calling.

## Dependencies

[@formio/core:98](https://github.com/formio/core/pull/98)

I added automated tests to cover conditionals cases in Edit Grid with Text field and Checkbox components, corresponding to the scenario where this bug was uncovered.

## Checklist:


brendanbond (changes requested):
Can we discuss this? This issue was also reported by an additional customer, but it seems to me this doesn't solve the root cause of the problem (because it is unsolvable? how does the conditional logic know which row is empty?).

mikekotikov:
Closing, since [another PR](https://github.com/formio/formio.js/pull/5638) is made to fix this

PR #5613, opened 2024-05-23
FIO-8366: API key is not unique translation

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8366

## Description

**What changed?**

- API key validation errors now are able to be translated using the apiKey property. This is possible now because the apiKey property was added to en.js translation file and the highlightInvalidComponents function which is responsible for setting the apiKey validation error now uses this.t("apiKey") instead of hard coding the error. The apiKey is also passed in the context of what apiKey is causing the error using comp.key. 
- The check to remove "api key is not unique" error message from the form is unnecessary as the check for invalid api keys is made when the form build is saved and the form builder will redraw the components upon saving which removes the apiKey error

manually

## Checklist:


brendanbond (changes requested):
Tests are failing.

travist:
You need to write a test for this.

ZenMasterJacob20011:
Added Tests. Also added some render html files generated from running test:updateRenders because if you add a json file to formtest directory the render tests expect a rendered html file of that json in the renders directory

brendanbond (approved):

PR #5634, opened 2024-06-06
FIO-7206: Fixes an issue where API keys of the removed components are…

alexandraRamanenka:
## Link to Jira Ticket
https://formio.atlassian.net/browse/FIO-7206

## Description

In some cases when Data Grid has a defaultValue with nested components' API keys inside, thoe keys were not removed after removing the child components. 

## Dependencies

## Checklist:

### DIFF
--- a/src/WebformBuilder.js
+++ b/src/WebformBuilder.js
@@ -1132,7 +1132,12 @@ export default class WebformBuilder extends Component {
         parent.formioComponent.removeChildComponent(component);
       }
       if (component.input && componentInstance && componentInstance.parent) {
-        _.unset(componentInstance._data, componentInstance.key);
+        if (Array.isArray(parent.formioComponent.component.defaultValue)) {
+          parent.formioComponent.component.defaultValue.forEach(v => _.unset(v, componentInstance.key));
+        }
+        else if (typeof parent.formioComponent.component.defaultValue === 'object') {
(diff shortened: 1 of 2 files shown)


src/WebformBuilder.js
```diff
@@ -1132,7 +1132,12 @@ export default class WebformBuilder extends Component {
         parent.formioComponent.removeChildComponent(component);
       }
       if (component.input && componentInstance && componentInstance.parent) {
-        _.unset(componentInstance._data, componentInstance.key);
+        if (Array.isArray(parent.formioComponent.component.defaultValue)) {
+          parent.formioComponent.component.defaultValue.forEach(v => _.unset(v, componentInstance.key));
+        }
+        else if (typeof parent.formioComponent.component.defaultValue === 'object') {
+          _.unset(parent.formioComponent.component.defaultValue, componentInstance.key);
+        }
```
TanyaGashtold, line 1140: I believe we should have more reliable check for parent.formioComponent.component.defaultValue. What if there is no parent or  parent.formioComponent. You can use lodash to get the default value: _.get(parent, 'formioComponent.component.defaultValue')

TanyaGashtold (changes requested):

src/WebformBuilder.js
alexandraRamanenka: There is a check for 'parent' at the very top of the method, so I added check for parent.formioComponent and used _.get to get it's default value

PR #5642, opened 2024-06-11
FIO-8420: file component no defaults causes error

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8420

## Description

**What changed?**

Previously, formio.js would error on a file upload if the noDefaults option in Formio.createForm function was set to true. This PR replaces this behavior by validating that this.dataValue exists before calling the .some function on it.

Although there were many potential solutions my solution was best because it does not cause any new errors/bugs to appear in the code and allows the next couple of lines of code to work without modification.

## Dependencies

N/A

## Checklist:

### DIFF
--- a/src/components/file/File.js
+++ b/src/components/file/File.js
@@ -744,8 +744,7 @@ export default class FileComponent extends Field {
     const fileWithSameNameUploading = this.filesToSync.filesToUpload
       .some(fileToSync => fileToSync.file?.name === file.name);
 
-    const fileWithSameNameUploaded = this.dataValue && this.dataValue
-      .some(fileStatus => fileStatus.originalName === file.name);
+    const fileWithSameNameUploaded = _.some(this.dataValue, fileStatus => fileStatus.originalName === file.name);
 
     return fileWithSameNameUploaded || fileWithSameNameUploading
       ? {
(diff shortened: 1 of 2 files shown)


brendanbond (approved):

src/components/file/File.js
```diff
@@ -744,7 +744,7 @@ export default class FileComponent extends Field {
     const fileWithSameNameUploading = this.filesToSync.filesToUpload
       .some(fileToSync => fileToSync.file?.name === file.name);
 
-    const fileWithSameNameUploaded = this.dataValue
+    const fileWithSameNameUploaded = this.dataValue && this.dataValue
       .some(fileStatus => fileStatus.originalName === file.name);
```
TanyaGashtold, line 748: I'd prefer to make this check even more secure by checking if this.dataValue is an array. Or you can just use lodash, like this:
_.some(this.dataValue, fileStatus => fileStatus.originalName === file.name);
Lodash does all the checking on the incoming value on its own and doesn't throw an error even if this.dataValue is undefine/string/number etc.

TanyaGashtold (changes requested):

src/components/file/File.js
ZenMasterJacob20011: Replaced check with lodash.some function

PR #5666, opened 2024-06-28
FIO-6632: Automate @formio/js updates to formiojs.test-form.io

ryanformio:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-6632

## Description

1. Noticed a few things weren't building quite right. The SemVer in the `package.json` was not being updated automatically during build process. Created a script that will run everytime `yarn build` is called to get the tag from git and update pacakge.json. Should ensure correct time every build.
https://github.com/formio/formio.js/actions/runs/9700333835 <- Happened to see that the version was v5.0.0-rc60 but did not see the version on the CDN. Saw v5.0.0-rc.59 and this was because the `package.json` version read 5.0.0-rc.59.
2. Noticed some differences in the CDN files based on looking at the artifact being distributed on [GH Pages](https://github.com/formio/formio.js/actions/runs/9701553308). 
3. Added updates to the `release.yml` to include the `yarn install`, `yarn build`, `yarn release`, and `yarn invalidate` to updates trigger properly in pipeline.

Was asked to provide some testing steps for previous issue FIO-6632 per Dane and noticed some issues in the process.

## Dependencies

n/a

Pushed the updates to test CDN to test commands manually.
https://formiojs.test-form.io/5.0.0-rc.60/

Link to steps documented in Confluence -> https://formio.atlassian.net/wiki/x/AQCkCQ

image

## Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,8 +3,7 @@ name: Release
 on:
   push:
     tags:
-      - '*' # This will make sure tag creations also trigger the workflow.
-
+      - '*'
 env:
   NODE_VERSION: 20.x
   AWS_DEFAULT_REGION: us-west-2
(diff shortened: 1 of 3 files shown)


travist:
I don't think we should be automatically updating the symver automatically. I am fine with DEV tags, but there may be some bad side effects of this releasing the RC versions automatically. It may also cause issues with our build script. Is this really necessary?

PR #5673, opened 2024-07-03
Fixed an issue where the 'root' may be the component that does not have a getComponent method.

travist:
https://formio.atlassian.net/browse/FIO-8802

## Description

The tests in formio CSVExporter are failing because it creates the components with Component.create().  The issue is that this will crash if the component has a simple conditional since the component does not have a method getComponent.

## Checklist:


src/utils/conditionOperators/DateGreaterThan.js
```diff
@@ -26,7 +26,7 @@ export default class DateGeaterThan extends ConditionOperator {
 
         let conditionTriggerComponent = null;
 
-        if (instance && instance.root) {
+        if (instance?.root?.getComponent) {
             conditionTriggerComponent = instance.root.getComponent(conditionComponentPath);
         }
```
johnformio, line 31: You can use optional chaining on [functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) as well.
```
let conditionTriggerComponent = instance?.root?.getComponent?.(conditionalComponentPath)
```

johnformio (approved):
Approved, but made an optional suggestion.

johnformio (approved):

PR #5676, opened 2024-07-05
FIO-8270: panel component closing on logic event trigger

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8270

## Description

**What changed?**

Previously, formio.js would revert back to the original component schema if a logic event action changed the hidden state of the component and any component triggered its change event. This PR replaces this behavior by allowing logic event listeners to update the original component hidden property so that it does not get reverted back when an input is made on the form.

Although there were many potential solutions, my solution was best because it allows all the tests to pass

## Dependencies

N/A

## Checklist:

### DIFF
--- a/src/components/_classes/component/Component.unit.js
+++ b/src/components/_classes/component/Component.unit.js
@@ -5,11 +5,8 @@ import sinon from 'sinon';
 import Component from './Component';
 import Webform from '../../../Webform';
 import Harness from '../../../../test/harness';
-import { comp1, comp6 } from './fixtures';
+import { comp1, comp4, comp3, comp5, comp6, comp7 } from './fixtures';
 import _merge from 'lodash/merge';
-import comp3 from './fixtures/comp3';
-import comp4 from './fixtures/comp4';
-import comp5 from './fixtures/comp5';
 
 describe('Component', () => {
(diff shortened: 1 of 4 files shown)


src/components/_classes/component/Component.unit.js
```diff
@@ -394,4 +391,22 @@ describe('Component', () => {
...
+      showButtonComponent.refs.button.click();
+      setTimeout(()=>{
+        textFieldComponent.refs.input[0].value = "test"
+        textFieldComponent.refs.input[0].dispatchEvent(new Event('input'));
+        setTimeout(()=>{
+          assert.equal(panelComponent.component.hidden, false);
+          assert.equal(panelComponent.visible, true);
+          done();
+        },400);
+      },200);
```
johnformio, line 409: I'm not fond of this pattern.  We should discuss a better pattern for these.

johnformio (approved):

src/components/_classes/component/Component.unit.js
ZenMasterJacob20011: Is it that we have two sources of truth to check if a component is visible or not?

PR #5687, opened 2024-07-10
FIO-8645: day component required validation error not correct

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8645

## Description

**What changed?**

Previously, formio.js did not have the requiredDayField translation key in the en.js file which caused the required validation to not display correctly. This PR replaces this behavior by adding requiredDayField translation key to the translation file

## Dependencies

N/A

## Checklist:

### DIFF
--- a/src/components/day/Day.unit.js
+++ b/src/components/day/Day.unit.js
@@ -233,11 +233,11 @@ describe('Day Component', () => {
         dayComponent.refs.day.dispatchEvent(new Event('input'));
 
         setTimeout(() => {
-          assert(dayComponent._errors.length && dayComponent._errors[0].message === 'requiredDayField', 'Day should be valid while changing');
+          assert(dayComponent._errors.length && dayComponent._errors[0].message === 'Day is required', 'Day should be valid while changing');
           dayComponent.refs.day.dispatchEvent(new Event('blur'));
 
           setTimeout(() => {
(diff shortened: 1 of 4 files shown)


travist (changes requested):
Couldn't we also fix this by just changing "requiredDayField" to just "required" and not introduce a new translation that is exactly the same as "required"?

ZenMasterJacob20011 (commented):
In @formio/core/src/process/validation/rules there is a file called validateRequiredDay.ts. This seems to be added in because you can set the required validation check on the day, month, and year inputs individually. Right now we have three translation keys for the day component and that is requiredDayField, requiredMonthField, and requiredYearField. I can either ...
- change requiredDayField, requiredMonthField, and requiredYearField to just required
- delete the validateRequiredDay.ts file and pass the responsibility back to some function inside the validateRequired.ts file. This would require a check to see if the component is a day component and some extra validation checking due to the fact that you can set required validations for day, month, and year individually.

I've also just noticed that even if you have required validation on for all of them, the validator will only throw an error for one of them in the order of day, month, and year. So if you have all of them set to required only the day will be marked as a required error

brendanbond (changes requested):
left a comment

ZenMasterJacob20011:
Added core tests and translations
https://github.com/formio/core/pull/119

ZenMasterJacob20011 (commented):
reviewed

brendanbond (approved):

PR #5734, opened 2024-08-08
FIO-8783: Fixed the issue when validation messages disappear after collapse and expand the panel

roma-formio:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8783

## Description

**Issue**: All nested components have a similar issue when using the `loadRef` method with `{ messageContainer: 'single' }`. This causes the nested component to attach the message container element from the first visible child component. As a result, during the expansion of the panel component, an additional validation check is triggered. After validating all child components, the nested component validates itself. If there are no error messages to show, the nested component clears its message container with errors. However, this reference is linked to the message container of the first visible child component, causing the child to lose its validation message.

Solution:
The same issue has already been resolved for the [EditGrid](https://github.com/formio/formio.js/pull/4734) and [DataGrid](https://github.com/formio/formio.js/pull/3957) components with different solutions that do not consider other nested components. In this PR, we remove the specific solutions for these components and instead use the concept of a `single-scope` ref for all nested components. This ensures that all nested components refer to their corresponding message containers.

An existing test has been updated to include additional checks to ensure that all messages are correctly placed.

## Checklist:


brendanbond:
fantastic, nice work

brendanbond (approved):

Issue #5735, opened 2024-08-08
ES Module Format Support


manuelroth:
**Problem Description**
I am using [Vite](https://vitejs.dev/) as a bundler. Vite expects ES Module by default. CommonJS Format is automatically transformed to ES Module during development builds. But somehow I can't get it working with formiojs for production builds. It results in `exports is not defined` Error: 
![image](https://github.com/user-attachments/assets/3209cf43-e03f-41e1-a67f-4e7cbd30da08). I guess this error originates in Vite not being able to convert formiojs from CommonJS to ES Module during production build. There for it still contains exports/require after building.

Currently formiojs only supports CommonJS as output format. When are gonna support ES Module output format?

**Solution**
In addition to CommonJS format, also support ES Module format. So Vite could directly use this version instead of having to do the conversion by itself.

manuelroth:
I opened a PR to correct the above issue: https://github.com/formio/formio.js/pull/5736

manuelroth:
In case somebody else encounters this issue using Vite. I was able to work around it by forcing Vite to convert formiojs and its dependencies to ES Module:
```js
import commonjs from '@rollup/plugin-commonjs';

export default defineConfig({
  plugins: [
    commonjs({
      include: [
        /node_modules\/formiojs(\/|\/node_modules\/).*/, // Include formiojs and all sub-dependencies
      ],
    }),
    ...
```
I shouldn't have to this if formiojs would be configured correctly. Please have a look into this. Many people are moving on from Webpack to Vite and will run into this issue. It would be ideal if you provide the library in ES Module format

ZenMasterJacob20011:
@manuelroth 
In order for us to look into this, we kindly ask for a minimally reproducible example of the issue. Unfortunately, we're unable to investigate the problem further without this. Once we receive your example, we'll be happy to take a closer look and help resolve it. We look forward to your response.

brendanbond:
@ZenMasterJacob20011 thanks (and @manuelroth that would be extremely helpful) but I'm going to take another stab at this to make sure we aren't missing anything obvious. @manuelroth if it's possible to provide a minimally reproducible example by all means, and I'll come back to this thread with what I find

manuelroth:
Thats awesome. Thank you for your effort ❤️

PR #5736, opened 2024-08-08
Corrects ES Module exports in package.json

manuelroth:
This PR correct the ES Module exports in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,23 +5,23 @@
   "main": "lib/cjs/index.js",
   "exports": {
     ".": {
-      "import": "./lib/cjs/index.js",
+      "import": "./lib/mjs/index.js",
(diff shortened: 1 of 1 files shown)


manuelroth:
I think its also important to define the `type` and `module` properties. See https://github.com/formio/formio.js/pull/5736/commits/41130970e84c60e571467e3542ae20f0451296c6. This way bundlers will know which entrypoint to use for CommonJS and ES Module.

manuelroth:
Hi @travist , were you able to test the changes downstream?- We still have to keep the workaround described in https://github.com/formio/formio.js/issues/5735#issuecomment-2277265172 in place. I would be glad if we could remove it at some point. Therefore it would be great to know if you have made an progress on this issue. Kind regards, Manuel

PR #5739, opened 2024-08-11
FIO-8375: allow date time to have manual input

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8375

## Description

**What changed?**

Previously when manual input was given to the date/time component the input mask would fill to create a complete time on blur. Now when the input triggers a blur event the manual input will not fill the input mask but instead be left incomplete. I have also made changes so that the validation will trigger as you are typing a manual input.

This solution allows you to use both the flatpickr and manual input without needing to switch one or the other off

This PR should be tested extensively for correct validation, and correct submission data when submitted. To allow for date time input mask to remain incomplete upon clicking outside of the input required a lot of fighting with the Flatpickrs event listeners. 

## Dependencies

N/A

Manually tested and automated test

## Checklist:


src/widgets/CalendarWidget.js
```diff
@@ -142,9 +135,13 @@ export default class CalendarWidget extends InputWidget {
...
       if (this.settings.allowInput && this.settings.enableTime) {
-        this.calendar._input.value = this.settings.manualInputValue || this.calendar._input.value;
-        this.settings.isManuallyOverriddenValue = false;
-        this.emit('update');
+        // We need to make sure that this block of code runs after Flatpickr sets the inputs value so that we can
+        // override the value Flatpickr sets with a manual value if we are in the state of manually overridden value
+        setTimeout(()=> {
+          this.calendar._input.value = this.settings.isManuallyOverriddenValue ? this.settings.manualInputValue : this.calendar.altInput.value;
+          this._input.value = this.settings.isManuallyOverriddenValue ? this.settings.manualInputValue : this.calendar.altInput.value
+          this.emit('update');
```
TanyaGashtold, line 143: 1)  this.calendar._input.value  and this._input.value has the same value, You can make this record shorter:
this.calendar._input.value = this._input.value =  this.settings.isManuallyOverriddenValue ? this.settings.manualInputValue : this.calendar.altInput.value;
2) It is not a good practice to use setTimeout often. We need to try to avoid them where it is possible. If this code must be executed after flatpickr sets the value, can we use it inside some other flatpickr hook, like onValueUpdate or onChange (https://flatpickr.js.org/events/)?

src/widgets/CalendarWidget.js
```diff
@@ -424,20 +421,24 @@ export default class CalendarWidget extends InputWidget {
...
     this.calendar = new Flatpickr(this._input, { ...this.settings, disableMobile: true });
     this.addEventListener(this.calendar.altInput, 'input', (event) => {
-      if (this.settings.allowInput && this.settings.currentValue !== event.target.value) {
-        this.settings.manualInputValue = event.target.value;
-        this.settings.isManuallyOverriddenValue = true;
-        this.settings.currentValue = event.target.value;
-      }
+      // Need to give time for the input mask to update so that the manualInputValue can be properly set
+      setTimeout(() => {
+        if (this.settings.allowInput && this.settings.currentValue !== event.target.value) {
```
TanyaGashtold, line 426: Can we try to avoid setTimeout here as well?

TanyaGashtold (changes requested):

ZenMasterJacob20011:
@TanyaGashtold I didn't realize deleting the comments in my ide would also delete the comments in github as well. I have removed the setTimeouts as requested. I also added some inputEvents for the tests because they would not pass without them. I believe that it is necessary to trigger these input events because manual values require input events to occur for them to work properly

PR #5773, opened 2024-08-26
FIO-7779: update pathing for nested container components and add test

brendanbond:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7779

## Description

**What changed?**

When a nested component is created, the parent's options are passed down with it. For example, an edit grid component that contains a child container component will have a `options.row` parameter passed down and added to the child container component. This causes a problem with pathing - we previously didn't account for container components and naively added a row index to the pathing parameter (via the function `calculateComponentPath`) if the component had a row parameter. This means that the path for any resulting child components of the container would be pathed as 

```editGrid[0].container[0].textField```

rather than

```editGrid[0].container.textField```.

This PR updates the `calculateComponentPath` function to account for container components.

This pathing behavior has been overhauled in the 5.x version of the renderer, so this should be thought of as exclusively a 4.x patch.

n/a

## Dependencies

n/a

Manually, and added an automated test for the root cause

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "fetch-ponyfill": "^7.1.0",
     "i18next": "22.4.12",
     "idb": "^7.1.1",
-    "inputmask": "^5.0.8",
+    "inputmask": "5.0.8",
     "ismobilejs": "^1.1.1",
     "json-logic-js": "^2.0.2",
     "jstimezonedetect": "^1.0.7",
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -81,7 +81,7 @@
     "fetch-ponyfill": "^7.1.0",
     "i18next": "22.4.12",
     "idb": "^7.1.1",
-    "inputmask": "^5.0.8",
+    "inputmask": "5.0.8",
```
brendanbond, line 84: Upgrading to 5.0.9 was causing the tests to fail, so I've pinned it here and asked for a ticket to be created to investigate/upgrade the dependency

TanyaGashtold (approved):

PR #5780, opened 2024-08-30
FIO-8914: fixed an issue where errors list doesnot appear when submitting a PDF form

TanyaGashtold:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8914

## Description

**What changed?**

We should always validate the submission that comes from the iframe, when the form  is submitted, to control the validation errors (add/remove them from the errors list).
Also this PR fixes the portal console error when editing the form with errors list as not all errors have component prop.

Manually with PDF forms

## Checklist:

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -1431,10 +1431,14 @@ export default class Webform extends NestedDataComponent {
         this.checkData(value.data, flags);
         const shouldValidate =
             !flags.noValidate ||
-            flags.fromIFrame ||
+            flags.fromIframe ||
             (flags.fromSubmission && this.rootPristine && this.pristine && flags.changed);
         const errors = shouldValidate
-            ? this.validate(value.data, { ...flags, process: "change" })
+            ? this.validate(value.data, { 
+                ...flags, 
+                noValidate: flags.fromIframe && this.submitted ? false : flags.noValidate, 
+                process: 'change'
+            })
             : [];
         value.isValid = errors.length === 0;
 
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -1436,7 +1436,7 @@ export default class Webform extends NestedDataComponent {
         const errors = shouldValidate
             ? this.validate(value.data, { 
                 ...flags, 
-                noValidate: flags.fromIframe && this.submitted ? false : flags.noValidate, 
+                noValidate: false, 
                 process: 'change'
             })
             : [];
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -3452,7 +3452,7 @@ export default class Component extends Element {
             this.parent.childErrors.push(...errors);
           }
           else {
-            _.remove(this.parent.childErrors, (err) => err.component.key === this.component.key);
+            _.remove(this.parent.childErrors, (err) => (err?.component?.key || err?.context?.key) === this.component.key);
           }
         }
         this.showValidationErrors(errors, data, row, flags);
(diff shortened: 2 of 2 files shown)


lane-formio:
I'd like to note that this passed on the first run. It only failed when I ran it a second time.

src/Webform.js
```diff
@@ -1431,10 +1431,14 @@ export default class Webform extends NestedDataComponent {
         this.checkData(value.data, flags);
         const shouldValidate =
             !flags.noValidate ||
-            flags.fromIFrame ||
+            flags.fromIframe ||
             (flags.fromSubmission && this.rootPristine && this.pristine && flags.changed);
         const errors = shouldValidate
-            ? this.validate(value.data, { ...flags, process: "change" })
+            ? this.validate(value.data, {
```
brendanbond, line 1437: I'm not sure I understand this change. Rather than an ad-hoc reversal of the `noValidate` flag, doesn't this indicate that actually `noValidate` is being set incorrectly upstream?

src/components/_classes/component/Component.js
```diff
@@ -3452,7 +3452,7 @@ export default class Component extends Element {
             this.parent.childErrors.push(...errors);
           }
           else {
-            _.remove(this.parent.childErrors, (err) => err.component.key === this.component.key);
+            _.remove(this.parent.childErrors, (err) => (err?.component?.key || err?.context?.key) === this.component.key);
```
brendanbond, line 3455: This seems fine, although it's another one where I might wonder if we need to update our error shape to include a `component` param for backwards compatibility purposes rather than make the change here.

brendanbond (changes requested):

src/Webform.js
TanyaGashtold: The answer is yes and no at the same time. Our pdf forms work a bit different than simple forms because they always need to be in contact with iFrame. On every change inside the iFrame the PDF form calls setValue that triggers all other processes like change, recalculations, conditions, validations and so on. 
So, when we want to submit the PDF, the PDF form sends the message to iFrame, and the iFrame sends back the submission object. Before to send this submission to the server, the PDF form needs to validate it. First, the PDF form calls the setValue and pass this submission object as an argument there. The setValue iterates through forms components and call setNestedValue for each of them. If there is no value for some component, it sets the default value together with noValidate flag:
image
In my test form, there was no value for number component and this is the place where noValidate turned to true. 
So, to handle it, we need to set noValidate to false here OR to ignore the noValidate flag in the place where the validation MUST be performed. I chose the second option as found it a bit more secure.
Alternative solution for it can be the passing a dirty flag to the setValue for the value that comes from iFrame, like this: 
image

src/components/_classes/component/Component.js
TanyaGashtold: Currently the front-end and server errors have different data structure. Server errors do not have component prop but have short information about component inside the context.
server and front error
If we want to standardize them, I believe we should do this as a part of separate ticket as it will require much more efforts. My current change is just a quick fix for the red error in console due to the lack of proper checks in the code.

src/Webform.js
brendanbond: I guess a follow up question here would be why would there ever be a case that we would call `validate()` with the `noValidate` flag being set to `true`? Couldn't we just always set it to `false`, e.g.
```
        const shouldValidate =
            !flags.noValidate ||
            flags.fromIframe ||
            (flags.fromSubmission && this.rootPristine && this.pristine && flags.changed);
        const errors = shouldValidate
            ? this.validate(value.data, {
                ...flags,
                noValidate: false,
                process: 'change'
            })
            : [];
        value.isValid = errors.length === 0;
```

brendanbond (approved):

PR #5829, opened 2024-09-25
FIO-9059: fixed an issue where the value for textarea with json datatype is not normalized and set correctly

TanyaGashtold:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9059

## Description

**What changed?**

- changed core version to the version with fix for the getModelType that returns any type for textArea with JSON data;
- fixed an issue where only first array element is set as a value for the textarea

## Dependencies

https://github.com/formio/core/pull/156

Manually + tests

## Checklist:


src/components/_classes/component/Component.js
```diff
@@ -2961,12 +2961,20 @@ export default class Component extends Element {
     }
     for (const i in this.refs.input) {
       if (this.refs.input.hasOwnProperty(i)) {
-        this.setValueAt(i, isArray ? value[i] : value, flags);
+        this.setValueAt(i, isArray && !this.isSingleInputValue() ? value[i] : value, flags);
```
brendanbond, line 2964: Isn't the core problem here that I didn't account for JSON text area types in the `normalizeValue` method of the `Multivalue` class? I'd rather just fix that function than spread this type of normalization logic around

brendanbond (changes requested):

src/components/_classes/component/Component.js
TanyaGashtold: This change is not related to the value normalization. When the value is being set, line 2962 iterates though the input references and calls setValueAt for each input. When the value is an array, it takes the " i" element of the array and sets it to the "i" input. But this does not work for textArea with array value. For it, we should set full array into the first input otherwise only the first array element is displayed.

brendanbond (approved):

PR #5832, opened 2024-09-27
FIO-9080 checkbox radio validation error

blakekrammes:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9080

## Description

**What changed?**

A while back, core was updated to use the component `name` property as the `key` for checkbox components with an `inputType` of `'radio'`. `name`, in turn, was then used instead of `key` to generate the component `path`. formio.js was never updated to use `name` instead of `key`, so checkbox–radio components could not be validated. 

This change updates formio.js to use the same helper function that core uses to get the component `key`.

It made sense to plug into core's functionality instead of duplicating the logic; the helper function from core was already imported into formio.js. This also aligns with the desire to move business logic from formio.js to core.

I'm not sure–– are the versions of core and formio.js always synced? If someone was using an older version of core from before the calculation changed with this update from formio.js it would break things.

## Dependencies

n/a

  - One test to assert that `Components.getComponentPath()` returns the correct component path for a checkbox–radio 
     component.
  - One test to assert that a web form renders a checkbox–radio component error correctly when `validation: required` is 
     true for that component.
## Checklist:

### DIFF
--- a/src/components/Components.js
+++ b/src/components/Components.js
@@ -1,6 +1,7 @@
 import Component from './_classes/component/Component';
 import EditFormUtils from './_classes/component/editForm/utils';
 import BaseEditForm from './_classes/component/Component.form';
+import { getComponentKey } from '../utils/utils';
 import _ from 'lodash';
 export default class Components {
   static _editFormUtils = EditFormUtils;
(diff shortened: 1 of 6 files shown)


brendanbond (approved):
Nice work!

blakekrammes:
Thanks! And thank you for your help.

PR #5846, opened 2024-10-06
FIO-9072: required validation triggered on form load

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9072

### Webform.js
- Added explicit submission to flags so that it can be used to make a component dirty or not
- Made it so that if noValidate flag is set to true then the component will not validate. This was a problem because sometimes the noValidate flag was set to undefined causing shouldNotValidate to be true unexpectedly  

### Component.js
- Changed it so that when the dirty flag is set to true it doesn't propagate to sibling components
- Changed it so that the components dirty property is set to true when there is an explicit submission

### NestedComponent.js
- Component now uses a copy of flags so that a change in flags from one component does not propagate throughout other components

### Editgrid.js
- If the Editgrid is dirty then the row should be validated

This solution is probably the closest to what the best solution would be. Unfortunately it is not a small change to make it so that the required validation is not triggered on form init. There is a lot of assumptions made around the dirty flag and it seems as though we were just getting lucky that this hasn't appeared before. A lot of the changes made here fix these assumptions.

N/A

## Dependencies

N/A

Added automated tests and manually tested. Also changed some tests which I will discuss in dev support

## Checklist:

### DIFF
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -795,6 +795,7 @@ export default class Webform extends NestedDataComponent {
      */
     onSetSubmission(submission, flags = {}) {
       this.submissionSet = true;
+      flags.explicitSubmission = structuredClone(submission);
       this.triggerChange(flags);
       this.emit('beforeSetSubmission', submission);
       this.setValue(submission, flags);
@@ -952,7 +953,6 @@ export default class Webform extends NestedDataComponent {
         if (!flags.sanitize) {
             this.mergeData(this.data, submission.data);
         }
-
         submission.data = this.data;
         this._submission = submission;
         return changed;
@@ -1433,14 +1433,14 @@ export default class Webform extends NestedDataComponent {
         }
 
         this.checkData(value.data, flags);
-        const shouldValidate =
-            !flags.noValidate ||
+        let shouldValidate = flags.noValidate ? false :
+            _.isUndefined(flags.noValidate) ||
             flags.fromIframe ||
-            (flags.fromSubmission && this.rootPristine && this.pristine && flags.changed);
+            (flags.fromSubmission && this.rootPristine && this.pristine);
         const errors = shouldValidate
-            ? this.validate(value.data, { 
-                ...flags, 
-                noValidate: false, 
+            ? this.validate(value.data, {
+                ...flags,
+                noValidate: false,
                 process: 'change'
             })
             : [];
--- a/src/Webform.js
+++ b/src/Webform.js
@@ -795,7 +795,7 @@ export default class Webform extends NestedDataComponent {
      */
     onSetSubmission(submission, flags = {}) {
       this.submissionSet = true;
-      flags.explicitSubmission = structuredClone(submission);
+      flags.submission = structuredClone(submission);
       this.triggerChange(flags);
       this.emit('beforeSetSubmission', submission);
       this.setValue(submission, flags);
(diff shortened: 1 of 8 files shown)


src/components/_classes/component/Component.js
```diff
@@ -3388,14 +3387,15 @@ export default class Component extends Element {
...
     }
+    let dirty = this.dirty || flags.dirty
     if (this.options.alwaysDirty) {
-      flags.dirty = true;
+      dirty = true;
     }
-    if (flags.fromSubmission && this.hasValue(data)) {
-      flags.dirty = this.pristine && this.component.protected ? false : true;
+    if (flags.fromSubmission && !_.isUndefined(_.get(flags?.submission?.data, this.key)) && !(this.pristine && this.protected)) {
+      dirty = this.pristine && this.component.protected ? false : true;
```
TanyaGashtold, line 3395: It seems to me that this.protected is always undefined. I was not able to find a place where it is set to true/false.
Why do we need the check for 'protected' and 'pristine' in both line 3395 and line 3394?

TanyaGashtold:
The changes seem logical to me. But I feel like they're a little risky. I want someone else to review them.

TanyaGashtold (changes requested):

src/components/_classes/component/Component.js
```diff
@@ -3388,14 +3387,15 @@ export default class Component extends Element {
...
       return [];
     }
+    let dirty = this.dirty || flags.dirty
     if (this.options.alwaysDirty) {
-      flags.dirty = true;
+      dirty = true;
     }
-    if (flags.fromSubmission && this.hasValue(data)) {
-      flags.dirty = this.pristine && this.component.protected ? false : true;
+    if (flags.fromSubmission && !_.isUndefined(_.get(flags?.submission?.data, this.key)) && !(this.pristine && this.protected)) {
```
travist, line 3394: _.get(submission.data, this.key) not always the way to fetch the correct data for a component. Particularly the Checkbox components configured as Radio components would fail here. Why can we not use "this.hasValue"?  It looks like you can pass data to it... maybe change this to ```this.hasValue(flags?.submission.data)```?

src/components/_classes/nested/NestedComponent.js
```diff
@@ -927,22 +927,24 @@ export default class NestedComponent extends Field {
 
   setNestedValue(component, value, flags = {}) {
     component._data = this.componentContext(component);
+    let componentFlags = {...flags};
```
travist, line 930: Is this necessary anymore now that we are no longer mutating the "dirty" property?  I would prefer to not have to clone if we don't need to.

travist (changes requested):

src/Webform.js
```diff
@@ -795,6 +795,7 @@ export default class Webform extends NestedDataComponent {
      */
     onSetSubmission(submission, flags = {}) {
       this.submissionSet = true;
+      flags.submission = structuredClone(submission);
```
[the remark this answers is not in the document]
ZenMasterJacob20011: This is so that the explicit submission can be used to determine whether or not to show validation. I could not use this.data or this.submission because these are filled with this.empty value. e.g. button components get false, textfield components get and empty string. I need to know the exact submission at the time it was inputted to determine if it was given an empty value or if it was not even a property in the submission json.

src/components/_classes/component/Component.js
ZenMasterJacob20011: hasValue(data) {
    return !_.isUndefined(_.get(data || this.data, this.key));
  }
hasValue will fallback on this.data. I do not want it to fall back on this.data because I want the explicit submission to set if the dirty flag is true. this.data will be populated with this.empty values in the init function of setting up the form. I need absolutely sure that the property was set by whoever defined the submission object in setSubmission(...). This is why I added the submission to the flags object when a submission is set.

PR #5860, opened 2024-10-13
FIO-9072: removed setting dirty flag to true on set submission and changed tests

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9072

## Description

**What changed?**

Removed the dirty flag being set to true on submission. This also required test changes because some tests set submissions and then check for visible errors using class names or message containers

This solution removes the root cause of errors unexpectedly occurring on form load by removing set submission causing components to become dirty

This PR goes directly against a previous test that I changed to should not show errors on ...
- it('Should show errors on setSubmission when providing explicit data values.')

## Dependencies

N/A

One of the automated tests I changed makes sure that set submission does not create visible errors. I also manually tested

## Checklist:

### DIFF
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -3388,14 +3388,12 @@ export default class Component extends Element {
     if (flags.silentCheck) {
       return [];
     }
+    let isDirty = this.dirty || flags.dirty;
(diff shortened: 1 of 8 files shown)


travist:
👍

PR #5902, opened 2024-11-11
FIO-9159: change clearOnHide behavior to track conditionally hidden state rather than mere visibility

brendanbond:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9159

## Description

This commit changes the behavior of the clearOnHide property. Previously, when this flag was present in a component, the renderer would unset the component's value from submission data if it was not visible (so, conditionally hidden, logically hidden, or intentionally hidden via the hidden property). Now, clearOnHide will only unset the component's data if it or its parent is conditionally or logically hidden. The copy of the parameter in the builder has also been changed.

**This constitutes a breaking change.** The clearOnHide property no longer functions the way that it used to. A component with `clearOnHide` enabled will "clear" the component's value if the component or its parent is conditionally hidden.

## Dependencies

[@formio/core#173](https://github.com/formio/core/pull/173)

Manual + automated tests

## Checklist:


src/components/form/Form.js
```diff
@@ -473,11 +473,11 @@ export default class FormComponent extends Component {
   }
 
   hideSubmitButton(component) {
-    const isSubmitButton = (component.type === 'button') &&
-      ((component.action === 'submit') || !component.action);
+    const isSubmitButton = component.type === 'button' && (component.action === 'submit' || !component.action);
 
     if (isSubmitButton) {
       component.hidden = true;
+      component.customConditional = 'show = false';
```
johnformio, line 480: What's up with this?

brendanbond: In nested forms we hide the submit button (to avoid two submit buttons). Previously, it was enough to attach `component.hidden` to the JSON to both (a) hide the button visually and (b) clear its data from the nested form submission (i.e. `{ data: { submit: true } }`). Now that we've changed clear on hide behavior, we need to preserve backwards-compatible button behavior by hiding it visually _and_ clearing its data, which in this case now means it needs to be conditionally hidden.

johnformio (approved):

PR #5913, opened 2024-11-18
FIO-9241: Correctly set current page after conditionally hiding page in sibling nested wizard

blakekrammes:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9241

## Description

**What changed?**

I updated the `onChange()` method in the Wizard class to call `this.setPage()` if the number of total pages in the parent/root-level wizard changes, which updates the current page of the root wizard to be the current page/page (i.e. `this.currentPanel`. `this.setPage()` was not being called `onChange()` for wizards with sub-wizards. I also updated `render()` to set `this.currentPanel` based on `id` and not `key`.

I changed `onChange()` in the way that I did to simplify the logic for when to set the current page and to ensure that the correct page is selected based on a (more) unique `id` and not `key`, which could be easily duplicated between sub-wizard pages.

For the update to `render()`, I made that change since another bug manifested where pages in multiple sub-wizards having the same `key` would cause the current page to jump from one sub-wizard to another. (i.e. `B2` to `C2`) when `A2` was conditionally hidden by the value of `B2` because both `B2` and `C2` have the same `key` of  (I think) `'page2'`. `A2` being a page of sub-wizard form `A`, and the same for pages `B2` and `C2`, in respect to the sub-wizards `B` and `C`.

I removed `this.currentPanels` from the Wizard class in order to simplify the logic in `onChange()` since the current Wizard pages/panels (including those of all sub-wizards) can be accessed via `this.pages`. This might break something, but I'm not sure of a case where it would. All of the tests pass with it removed. See [this](https://github.com/formio/formio.js/pull/5913#discussion_r1853051774) comment for more details.

## Dependencies

n/a

Manual and automated testing.

## Checklist:

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -37,7 +37,6 @@ export default class Wizard extends Webform {
     this.originalComponents = [];
     this.page = 0;
     this.currentPanel = null;
-    this.currentPanels = null;
     this.currentNextPage = 0;
     this._seenPages = [0];
     this.subWizards = [];
@@ -60,14 +59,14 @@ export default class Wizard extends Webform {
 
   getPages(args = {}) {
     const { all = false } = args;
-    const pages = this.hasExtraPages ? this.components : this.pages;
+    const pages = this.hasSubWizards ? this.components : this.pages;
     const filteredPages = pages
       .filter(all ? _.identity : (p, index) => this._seenPages.includes(index));
 
     return filteredPages;
   }
 
-  get hasExtraPages() {
+  get hasSubWizards() {
     return !_.isEmpty(this.subWizards);
   }
 
@@ -216,9 +215,9 @@ export default class Wizard extends Webform {
   render() {
     const ctx = this.renderContext;
 
-    if (this.component.key) {
-      ctx.panels.map(panel => {
-        if (panel.key === this.component.key) {
+    if (this.component.id) {
+      ctx.panels.forEach(panel => {
+        if (panel.id === this.component.id) {
           this.currentPanel = panel;
           ctx.wizardPageTooltip = this.getFormattedTooltip(panel.tooltip);
         }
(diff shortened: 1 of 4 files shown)


src/Wizard.js
```diff
@@ -216,9 +215,9 @@ export default class Wizard extends Webform {
   render() {
     const ctx = this.renderContext;
 
-    if (this.component.key) {
+    if (this.component.id) {
```
blakekrammes, line 218: There was an additional bug here:

If you had multiple/sibling nested wizards and conditional logic controlling the visibility of a nested wizard page, this was setting the currentPanel to the wrong page on render (i.e. both forms have a 'page1' key. This is fixed by using a unique id to set the currentPanel.

src/Wizard.js
```diff
@@ -1006,24 +1005,18 @@ export default class Wizard extends Webform {
...
-      this.establishPages();
-      panels = this.pages.map(page => page.component.key);
+    const panelsUpdated = !_.isEqual(newPanels, currentPanels);
+    
+    if (this.currentPanel?.id && this.pages.length && (!this.hasSubWizards || (this.hasSubWizards && panelsUpdated))) {
+      const newIndex = this.pages.findIndex(page => page.id === this.currentPanel.id);
+      if (newIndex !== -1) this.setPage(newIndex);
     }
-    else {
-      currentPanels = this.currentPanels || this.pages.map(page => page.component.key);
```
blakekrammes, line 1018: `this.currentPanels` was only referenced within `onChange` and seems like an unnecessary piece of state given that the current pages/panels can be accessed via `this.pages`. All the tests pass happily without it.

src/Wizard.js
```diff
@@ -216,9 +215,9 @@ export default class Wizard extends Webform {
   render() {
     const ctx = this.renderContext;
 
-    if (this.component.key) {
-      ctx.panels.map(panel => {
-        if (panel.key === this.component.key) {
+    if (this.component.id) {
+      ctx.panels.forEach(panel => {
```
blakekrammes, line 219: `.forEach` is more idiomatic since we're applying side-effects and not returning a new array.

brendanbond (approved):
nice work

PR #5088, opened 2023-03-06
FIO-6234: added submit button to the webform builder after switching from the w…

RomanLetsuk:
…izard builder

## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-6234

## How has this PR been tested?

Added test.

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] My changes generate no new warnings
- [x] My changes include tests that prove my fix is effective (or that my feature works as intended)
- [x] New and existing unit/integration tests pass locally with my changes
- [ ] Any dependent changes have corresponding PRs that are listed above

PR #5919, opened 2024-11-21
FIO-9361: Fix showing extra submit buttons

mikekotikov:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9361

## Description

**What changed?**

Made better checks for different cases when to add submit button to the form, so to prevent scenarios when couple submit buttons could appear.

It should work for all those cases:
1) changing submit button action on an unsaved form shouldn't cause doubled submit button
2) changing submit button api key on an unsaved form shouldn't cause double submit button
3) switching existing or imported wizard form to a simple form should add submit button
4) imported form shouldn't have doubled submit button
5) copied form shouldn't have doubled submit button

*Use this section to justify your choices*

## Dependencies

Automated tests added

## Checklist:

### DIFF
--- a/src/WebformBuilder.js
+++ b/src/WebformBuilder.js
@@ -1040,15 +1040,19 @@ export default class WebformBuilder extends Component {
     }
     this.keyboardActionsEnabled = keyboardActionsEnabled;
 
-    const isSubmitButton = (comp) => {
-      return (comp.type === 'button') && ((comp.action === 'submit') || !comp.action);
-    }
+    const { display, noAddSubmitButton, noDefaultSubmitButton } = this.options;
+    const { _id, components } = form;
 
-    const isShowSubmitButton = !this.options.noDefaultSubmitButton
-      && (!form.components.length || !form.components.find(comp => isSubmitButton(comp)));
+    const isSubmitButton = ({ type, action }) => type === 'button' && (action === 'submit' || !action);
+    const hasSubmitButton = components.some(isSubmitButton);
+    // Add submit button if form display was switched from wizard
+    // Don't add if there is noAddSubmitButton flag passed, or the form has id, or the form has a submit button already
+    const shouldAddSubmitButton =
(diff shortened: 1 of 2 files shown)


brendanbond (changes requested):
This looks fine at a glance, but I think we'd like some more information on why this regressed? It seems like something fundamental must have changed for this to suddenly be a problem

travist:
@mikekotikov please respond to @brendanbond inquiry.  We would like to understand why this regressed...

mikekotikov:
> This looks fine at a glance, but I think we'd like some more information on why this regressed? It seems like something fundamental must have changed for this to suddenly be a problem

@brendanbond @travist there were couple changes that led to this regression:
1) https://github.com/formio/formio.js/pull/5088 - customer request to add submit btn to the form when switching from wizard
2) https://github.com/formio/formio.js/pull/5818 - spotted by QA that after 1) changes, when changing api key of submit button, double button would appear
The main problem has started with 1) changes since builder was adding submit button when it couldn't find it in form components.

johnformio:
Looking at that [pr](https://github.com/formio/formio.js/pull/5088), could it be because of this change:
```
 else {
      const components = this._form.components
        .filter((component) => component.type !== 'button' || component.action !== 'submit');
      this._form.components = components;
    }
    ```
    Does the `component.action` change when it switches to reset and is included in the component list?

mikekotikov:
> Looking at that [pr](https://github.com/formio/formio.js/pull/5088), could it be because of this change:
> 
> ```
>  else {
>       const components = this._form.components
>         .filter((component) => component.type !== 'button' || component.action !== 'submit');
>       this._form.components = components;
>     }
> ```
> 
> Does the `component.action` change when it switches to reset and is included in the component list?

@johnformio yes, exactly, the action is changing, so the code was thinking there's no submit button in existing compoents and was adding it.

PR #5925, opened 2024-11-25
FIO-9327 Fixed unintentional setting of defaultValue for Select Boxes

antonSoftensity:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9327

## Description

With the changes to the new Validation process, when new values were added to Select Boxes in the Component settings window, emptyValue was called for the Default Component, and the value was incorrect due to the old code that was used to just set Select Boxes options to `false` and make it an empty value. The fix was to set emptyValue to {} always, because it is a better representation of empty value which is never affected by the actual values of the component.

## Dependencies

Tested locally. Tests are passing for the SelectBoxes components. I've decided not to add new tests because the behavior is straightforward and we don't need to test that a component has specific emptyValue, when it is just an empty object all the time.

UPD: added 'selectBoxes' to an exception for a modalEdit test from basic component settings tests, due to the fact that the test for 'closing modal' was failing (during debugging it showed that everything was rendering correctly, just for some reason the handler for 'Close' button wasn't called, which is not connected with the change to 'emptyValue' function in my opinion, at least I couldn't find any evidence of that)

## Checklist:

### DIFF
--- a/src/components/selectboxes/SelectBoxes.js
+++ b/src/components/selectboxes/SelectBoxes.js
@@ -77,12 +77,7 @@ export default class SelectBoxesComponent extends RadioComponent {
   }
 
   get emptyValue() {
-    return this.component.values.reduce((prev, value) => {
-      if (value.value) {
-        prev[value.value] = false;
-      }
-      return prev;
-    }, {});
+    return {};
   }
 
   get defaultValue() {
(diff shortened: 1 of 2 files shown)


travist:
I don't believe this is the correct fix for the problem. The issue is that the "defaultValue" is getting set and does not continually set to the correct value as you are editing the component values in the Builder mode. You are fixing this by just changing the default value of the Select Boxes component, which does not seem like it is fixing the root problem. I am going to investigate this today and will for now close this pr.

PR #5938, opened 2024-12-05
fix select styling

brendanbond:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9453

## Description

Select components in Bootstrap 5 had no static height - it turns out that Bootstrap 5 removed the static height declaration for the `.form-control` class. This PR adds it in the formio global styles for backwards compatibility.

A possible alternative involved adding this style, perhaps inline, to the select component's bootstrap5 template. However, this would not work - choicesjs masks the actual input, so nothing in the template would really have a visual effect. Furthermore, we need to modify the select's parent element, and it the [renderer handles adding the class in question](https://github.com/formio/formio.js/blob/3ff37eb652a83ecc88dac3e3788283d97e7cac27/src/components/_classes/component/Component.js#L3912).

n/a

## Dependencies

n/a

n/a

## Checklist:

### DIFF
--- a/src/sass/formio.form.scss
+++ b/src/sass/formio.form.scss
@@ -31,6 +31,10 @@ $autocomplete-in-dialog-z-index: $dialog-z-index + 1000;
   border-color: #ffeeba;
 }
 
+.form-control {
+  height: calc(1.4em + 0.75rem + 2px);
+}
+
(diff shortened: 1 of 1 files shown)


travist:
+1

PR #5939, opened 2024-12-06
FIO-7937/FIO-7953: Fixes issues where Reset action does not clear values and restore the default ones

alexandraRamanenka:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-7937
https://formio.atlassian.net/browse/FIO-7953

## Description

1.  Call setValue of the Component Modal while resetting value, because it has its own copy of the value that has to be reset separetely (it was leading to modal preview showing the old value after Reset and restoring the value after opening and closing modal window without saving it after cklicking Reset) 
2. Change teh order of calling parent component resetValue and child components reset Value in NestedComponent since otherwise it will restore the deafult value of the parent component, but then all the child compopnents wil reset to the empty value. 
3. redraw the Address component after resetting value as well to reflect the change visually. 
4. Removed custom resetValue method of EditGrid that was calling emptyRows methos after resetting the value (the change was made for Dynamic Wizard fix that is no longer supported) 
5. Made an uncheckValue method for Radio component to use when the user clicks already selected value to uncheck it instead of using resetValue method that should restore the deafult/empty value by design

## Dependencies

*Use this section to describe how you tested your changes; if you haven't included automated tests, justify your reasoning*

## Checklist:


brendanbond (changes requested):
tests around reset are failing

PR #5954, opened 2024-12-22
FIO-9480: ensure parent references are stable before subform creation

brendanbond:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9480

## Description

Sometimes a deeply nested child needs state or data from a distant parent, and needs to crawl up the reference tree to get it. E.g. a calculated value that looks like 
```javascript
const myValue = instance?.root.parent?.root?.parent?.root?.parent?.data?.role;
``` 

Previously these references were not stable, because they were being instantiated _after_ subForm creation. This PR fixes this behavior so that they are passed along with the creation options, and those references will be stable from the start.

n/a

## Dependencies

n/a

## Checklist:


travist:
👍

PR #5966, opened 2025-01-03
FIO-9524: Fix bad request when fetching nested form

mikekotikov:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9524

## Description

**What changed?**

When esignature module is used, the nested form gets initialized with a Formio from it's build. And it doesn't have a set project url that matches currently open project. So the final link to the nested form is malformed and 400 Bad Request happens.
I've tried adjusting the imports of Formio to point to same Formio instance, but didn't succeed with that approach.
So as another solution I've managed to pass the project URL in the options object and use it if it's present.

*Use this section to justify your choices*

## Dependencies

[ngFormio#741](https://github.com/formio/ngFormio/pull/741)

Automated test is added

## Checklist:

### DIFF
--- a/src/components/form/Form.js
+++ b/src/components/form/Form.js
@@ -65,7 +65,7 @@ export default class FormComponent extends Component {
         this.options.project = this.formSrc;
       }
       else {
-        this.formSrc = Formio.getProjectUrl();
+        this.formSrc = this.options.project || Formio.getProjectUrl();
         this.options.project = this.formSrc;
       }
       if (this.component.form) {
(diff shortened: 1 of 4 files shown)


brendanbond (changes requested):
So if I understand this correctly, esignature is bundling a version of Formio with itself. Isn't this the fundamental problem? Why does esignature need to bundle Formio, it should depend on it as a peer dependency (or as an `external` in webpack)

mikekotikov:
> So if I understand this correctly, esignature is bundling a version of Formio with itself. Isn't this the fundamental problem? Why does esignature need to bundle Formio, it should depend on it as a peer dependency (or as an `external` in webpack)

@brendanbond you're right 👍🏼 
Rechecked with @johnformio changes with [omitting formio deps bundling](https://github.com/formio/esignature/pull/9) and the issue is gone. So closing the PRs for this item

PR #6047, opened 2025-03-13
FIO-9813: Add validate when hidden property to the Nested Form if any of the children has it

alexandraRamanenka:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9813

## Description

If there are some components inside hidden nested form that have validateWhenHidden, add this property to the Nested Form component itself, so during the parent's form validation it won't be skipped. 

## Dependencies

https://github.com/formio/core/pull/229

*Use this section to describe how you tested your changes; if you haven't included automated tests, justify your reasoning*

## Checklist:

### DIFF
--- a/src/components/form/Form.js
+++ b/src/components/form/Form.js
@@ -437,6 +437,12 @@ export default class FormComponent extends Component {
       // Iterate through every component and hide the submit button.
       eachComponent(form.components, (component) => {
         this.hideSubmitButton(component);
+        if (component.validateWhenHidden) {
+          this.component.validateWhenHidden = true;
+          // Change original component as well, so it won't cause any custom logic to be continuously triggered
+          // because the original component does not have this property
+          this.originalComponent.validateWhenHidden = true;
+        }
       });
 
       // If the subform is already created then destroy the old one.


src/components/form/Form.js
```diff
@@ -437,6 +437,12 @@ export default class FormComponent extends Component {
       // Iterate through every component and hide the submit button.
       eachComponent(form.components, (component) => {
         this.hideSubmitButton(component);
+        if (component.validateWhenHidden) {
+          this.component.validateWhenHidden = true;
+          // Change original component as well, so it won't cause any custom logic to be continuously triggered
+          // because the original component does not have this property
+          this.originalComponent.validateWhenHidden = true;
```
travist, line 444: It seems to me that we should not mutate the component by adding a property. I would rather add a non-enumerable property using Object.defineProperty.

```
Object.defineProperty(component, 'validateWhenHidden', {
  value: true,
  writable: true,
  enumerable: false
});
```

brendanbond: Discussed with @alexandraRamanenka - we're going to go with a top down approach for nested forms. In other words, the nested form component will add the "validateWhenHidden" property to its definition, and child components will respect that parent property

alexandraRamanenka:
Another solution figured out

PR #6048, opened 2025-03-13
Fixing issues with the data map and unnecessary reset model functions.

travist:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9856

## Description

This changes multiple issues:

 1.) Performance issues - It was identified that a major performance bottleneck was that clearOnHide was being called within the checkComponentConditions method. The issue is that clearOnHide for NestedComponents iterate through all the children to perform clearOnHide for those children. The problem presents itself within the checkData loop where it is already iterating through all of the children and performing these checks, so adding multiple tree iterations is a major performance drain. This solves the problem by implementing a clearComponentOnHide method which is not implemented by the NestedComponent class, and that method is what is called within the checkComponentConditions method.
 
 2.) Data Map data model issues - With the recent data model fixes and streamlining, it became clear we had an issue with Data Map setting data properly. The issue was because there was some zombie code that was resurrected and executed after the recent data model changes that were made. This method is called "restoreComponentsContext" and was introduced @ https://github.com/formio/formio.js/commit/d75812e5b0583c1bbbc4c324504a3ba718b8e2a0 which attempted to solve an old problem where the data models would become "detached" within EditGrid and DataGrids. This issue was solved a while back and no longer was necessary. The reason this clobbered the Data Map was because of this line of code in DataGrid (DataMap extends DataGrid) https://github.com/formio/formio.js/blob/e7c608aeb25ef2aede01b54ffb85858534b7d36c/src/components/datagrid/DataGrid.js#L728.  This code is not compatible with DataMaps.

 3.) valuePropery on Select edit forms was getting reset:  The "valueProperty" is implemented in two different places based on the data source type. One of them "clearOnHide" is set to false, but the other one is true. Due to the streamlined data model fixes, what was happening is that the other valueProperty configuration would be hidden and the data would be reset because it is hidden. The fact that we fixed the bug, exposed this bug.  We can solve this by just making valueProperty not clear when hidden.

 4.) Another issue presented itself with the customDefaultValue and calculatedValues.  There are cases when the these functions are implemented not to set a value, but rather just execute a script. There is now a check to see if the script sets the value using a regex that looks for "value=".  If that is provided then we know we need to set the value to what is returned.

None that we are aware of. All existing tests pass.

## Dependencies

None

This change affects most components that already have many automated tests written to test for the capabilities of the things that were changed in this PR.

## Checklist:

### DIFF
--- a/src/Wizard.js
+++ b/src/Wizard.js
@@ -832,7 +832,10 @@ export default class Wizard extends Webform {
   validateCurrentPage(flags = {}) {
     const components = this.currentPage?.components.map((component) => component.component);
     // Accessing the parent ensures the right instance (whether it's the parent Wizard or a nested Wizard) performs its validation
-    return this.currentPage?.parent.validateComponents(components, this.root.data, flags);
+    if (this.currentPage?.parent) {
+      return this.currentPage?.parent.validateComponents(components, this.root.data, flags);
+    }
+    return this.currentPage?.validateComponents(components, this.root ? this.root.data : this.data, flags);
   }
 
   emitPrevPage() {
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -398,6 +398,14 @@ export default class Component extends Element {
     }
 
     /**
+     * Same as customDefaultValue, but for calculateValue.
+     */
+    this.shouldSetCalculatedValue = true;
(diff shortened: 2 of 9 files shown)


brendanbond (approved):
LGTM

PR #6082, opened 2025-04-06
FIO-9847: time on textfield only changes when timezone included

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9847

## Description

**What changed?**

Added encoded timezone checks to determine if setting a timezone value should use the plain text timezone

Trying to make changes without breaking anything

If a time does not have timezone information attached to it then it will be treated as plain text for textfields

## Dependencies

N/A

manually tested

## Checklist:

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -1750,3 +1750,14 @@ export const interpolateErrors = (component, errors, interpolateFn) => {
     return { ...error, message: unescapeHTML(interpolateFn(toInterpolate, context)), context: { ...context } };
   });
 };
+
+/**
+ * Checks if a string has timezone information encoded in it
+ * Example: 2024-01-01T00:00:00Z -> true
+ * Example: 2011-05-03T00:00:00 -> false
+ * @param {string} value the string value to check
+ * @returns {boolean} if value has encoded timezone
+ */
+export function hasEncodedTimezone(value){
+  return (value.substring(value.length - 1) === 'z' || value.substring(value.length - 1) === 'Z' || value.match(/[+|-][0-9]+:[0-9]+/));
+}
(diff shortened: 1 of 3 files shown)


src/widgets/CalendarWidget.js
```diff
@@ -344,6 +344,12 @@ export default class CalendarWidget extends InputWidget {
       return super.setValue(value);
     }
 
+    // If the component is a textfield that does not have timezone information included in the string value then skip
+    // the timezone offset
+    if(this.component.type === 'textfield' && !(hasEncodedTimezone)(value)){
```
travist, line 349: This is a strange way to call a function.

```
(hasEncodedTimezone)(value)
```

src/utils/utils.js
```diff
@@ -1750,3 +1750,18 @@ export const interpolateErrors = (component, errors, interpolateFn) => {
...
+ * Example: 2024-01-01T00:00:00+03:00 -> true
+ * Example: 2011-05-03T00:00:00 -> false
+ * @param {string} value the string value to check
+ * @returns {boolean} if value has encoded timezone
+ */
+export function hasEncodedTimezone(value){
+  if (typeof value !== 'string'){
+    return false;
+  }
+  return (value.substring(value.length - 1) === 'z' || value.substring(value.length - 1) === 'Z' || value.match(/[+|-][0-9]+:[0-9]+/));
```
travist, line 1766: Can we make the regex more specific?

```
value.match(/[+|-][0-9]{2}:[0-9]{2}/)
```

ZenMasterJacob20011: Made the regex more specific

src/widgets/CalendarWidget.js
ZenMasterJacob20011: removed the parenthesis around the function. Not sure why I did that

brendanbond (approved):

PR #6088, opened 2025-04-11
FIO-9985 fixed appearing extra option with an empty key

HannaKurban:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9985

## Description

*Previously, when submitting a persistent SelectBoxes component with the url dataSrc type, an option with an empty key and a false value could appear in the data. The reason for this behavior was that when adding a new SelectBoxes component with the Url dataSrc, an option with an empty key and a false value was added to the default value. This behavior was fixed and now only the corresponding loaded options are displayed in the list of options.*

*n/a*

## Dependencies

*n/a*

*all tests pass locally*

## Checklist:

### DIFF
--- a/src/components/selectboxes/SelectBoxes.js
+++ b/src/components/selectboxes/SelectBoxes.js
@@ -176,6 +176,9 @@ export default class SelectBoxesComponent extends RadioComponent {
         }
       }
     }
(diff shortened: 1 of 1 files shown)


johnformio (changes requested):
Can you add a test for this please.

johnformio (approved):

PR #6094, opened 2025-04-16
FIO-8539: updated datetime and utils to use dayjs

ZenMasterJacob20011:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-8541

## Description

**What changed?**

Instead of using moment.js we use dayjs now. Also removed zone loading in replacement for dayjs timezone plugin

Slowly transitioning from using moment to dayjs

The Abbreviated named offset for time zone differences can be different. For example CET might instead be GMT+2

## Dependencies

N/A

Manually tested selecting dates, making submissions, and checking all the display ins for viewing submissions in different timezones side by side with the portal application running formio.js without my changes and a local formio.js with my changes

## Checklist:


test/unit/DateTime.unit.js
```diff
@@ -60,7 +60,7 @@ describe('DateTime Component', () => {
 
     const dateTime = form.getComponent('dateTime');
     const response = dateTime.getValueAsString('2025-02-11T10:00:00.000Z', { email: true })
-    assert.equal(response, "2025-02-11 11:00 AM CET");
```
brendanbond, line 63: This is interesting, would you describe this as a breaking change? (the way the dayjs displays time zone data in a date string versus moment)?

brendanbond (changes requested):

test/unit/DateTime.unit.js
ZenMasterJacob20011: This would be worth considering in the change log. It looks like dayjs has different timestamp ending data than moment. Sometimes instead of CET you will see GMT + 3 or something along those lines.

PR #6112, opened 2025-04-30
FIO-9974 fixed Clear on Hide for layout components

HannaKurban:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-9974

## Description

*The reason for the issue of not being able to uncheck the Default Group in the Builder Configurator form is a regression (components inside the Layout Component do not clear values when the ClearOnHide property is set to true). That is, if there is a Layout component (for example, a Panel that can be conditionally hidden) and a nested component in it (for example, a TextField with the ClearOnHide property set to true), then when filling in the value of the nested component and then hiding the parent layout component , the data of the child component is not cleared. The reason for this is that in the `shouldConditionallyClear` method, provided that `this.hasSetValue == true`,  `this._conditionallyClear` property is always [set to `false`](https://github.com/formio/formio.js/blob/master/src/components/_classes/component/Component.js#L826), although `this.parentShouldConditionallyClear()` [should be checked](https://github.com/formio/formio.js/blob/5.1.x/src/components/_classes/component/Component.js#L784).*

*n/a*

## Dependencies

*n/a*

*Automated test was added/ All tests pass locally*

## Checklist:

### DIFF
--- a/src/components/_classes/component/Component.js
+++ b/src/components/_classes/component/Component.js
@@ -823,7 +823,7 @@ export default class Component extends Element {
       this._conditionallyClear = true;
       return this._conditionallyClear;
     }
-    this._conditionallyClear = this.hasSetValue ? false : this.parentShouldConditionallyClear();
+    this._conditionallyClear = this.parentShouldConditionallyClear();
     return this._conditionallyClear;
   }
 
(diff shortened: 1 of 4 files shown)


travist (changes requested):
There has to be a better way to do this other than having to iterate through ALL your children in the deleteValue method. My guess is this will cause performance regressions.

HannaKurban:
The approach was changed and now the parent layout-component check is used instead of the eachComponent method.

PR #6252, opened 2025-10-01
FIO-5687: Fix: Support multiple values in Select component setValue/updateValue in HTML 5 mode

### LINKED ISSUE #5572: Select component with a "Widget type" of "HTML 5" and "Multiple Values" set to true does not allow for multiple values to be selected
**Describe the bug**
When adding a "Select" component to a form with a "Widget type" set to "HTML 5" and "Multiple Values" set to true, it's not possible to select multiple values from the drop down list. It works fine when the "Widget type" is set to "ChoicesJS" and selecting from the dropdown list adds the selected value in the components first line as a removable token.

**Version/Branch**
Tested in both 4.16.0 and 4.17.0-rc.2

**To Reproduce**
1. Create form.
2. Add a "Select" component, in the "Display" property tab set "Widget Type" to "HTML 5", in the "Data" property tab set "Multiple Values" to true and add at least two "Data Source Values".
3. Render the form and try to select one value from the list.
4. Try to select multiple values from the list.

**Expected behavior**
Step 3: The selected item should be displayed in the first line of the component.
Step 4: All the selected items should be displayed in the first line of the component.

**Additional context**
See the following sandbox: 6617ebea28302f197dd16829 for a sample in which the "Select" component is tested with all the possible permutations of "Widget Type" and "Multiple Values" accompanied by a TextField on the right to show the actual selected values from each Select component. We can see that in the fouth row, the one that shows the actual problem, selecting an item in the list would register a value, but it's not possible to select multiple value as expected.

mpetit-boop:
## Link to Jira Ticket

https://formio.atlassian.net/browse/FIO-5687

## Description

**What changed?**

The select component, with HTML 5 as the display type, did not support multiple values when that option was selected.

This PR updates the setValue() and updateValue() logic in the Select component to correctly handle multiple values and improve event handling in HTML5 mode.

- Adjusted setValue() to support arrays of values.
- Updated updateValue() to correctly merge modified values.
- Enhanced attach() to better handle change events for multiple selections.

Fixes #5572.

The conditions were added to ensure that the values are stored in an array if the multiple values option is selected in HTML 5 mode, while minimizing the impact this could have on the existing logic also shared when in ChoicesJS mode.

I do not believe this code can potentially break changes or have any adverse effects on backwards compatibility.

## Dependencies

There are no dependent changes/PRs in other Form.io modules.

This was tested locally, with different select component settings:

- a select using HTML 5 and with the multiple values option enabled
- a select using HTML 5 without the multiple values option
- a select using Choices JS with multiple values option enabled
- a select using Choices JS without the multiple values option


lane-formio:
@mpetit-boop Now there's a merge conflict 😞

PR #6306, opened 2026-04-23
[Snyk] Security upgrade uuid from 9.0.1 to 14.0.0

lane-formio:
snyk-top-banner

#### Snyk changed the following file(s):

- `package.json`

<details>
<summary>⚠️ <b>Warning</b></summary>

```
Failed to update the package-lock.json, please update manually before merging.
```

</details>

#### Vulnerabilities that will be fixed with an upgrade:

|  | Issue | Score | 
:-------------------------:|:-------------------------|:-------------------------
medium severity | Improper Validation of Specified Index, Position, or Offset in Input <br/>[SNYK-JS-UUID-16133035](https://snyk.io/vuln/SNYK-JS-UUID-16133035) | &nbsp;&nbsp;**708**&nbsp;&nbsp; 

#### Breaking Change Risk

Merge Risk: High

> **Notice:** This assessment is enhanced by AI.

---

> [!IMPORTANT]
>
> - Check the changes in this PR to ensure they won't cause issues with your project.
> - Max score is 1000. Note that the real score may have changed since the PR was raised.
> - This PR was automatically created by Snyk using the credentials of a real user.

---

**Note:** _You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs._

For more information: 
🧐 [View latest project report](https://app.snyk.io/org/travist-ulg/project/ddd6c4e1-3edc-4a0f-ac70-76d8e2f7264a?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)
📜 [Customise PR templates](https://docs.snyk.io/scan-using-snyk/pull-requests/snyk-fix-pull-or-merge-requests/customize-pr-templates?utm_source=github&utm_content=fix-pr-template)
🛠 [Adjust project settings](https://app.snyk.io/org/travist-ulg/project/ddd6c4e1-3edc-4a0f-ac70-76d8e2f7264a?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)
📚 [Read about Snyk's upgrade logic](https://docs.snyk.io/scan-with-snyk/snyk-open-source/manage-vulnerabilities/upgrade-package-versions-to-fix-vulnerabilities?utm_source=github&utm_content=fix-pr-template)

---

**Learn how to fix vulnerabilities with free interactive lessons:**

[...]

PR #6335, opened 2026-08-12
Fix: remove uuid dependency by using native crypto.randomUUID (CVE-2026-41907)

dahhou-ilyas:
## Link to Jira Ticket

N/A — external community contribution, no FIO ticket available.
Happy to link one if a maintainer creates it.

Related: closes the gap left by #6306 (Snyk bump, closed unmerged).

## Description

**What changed?**

- Removed the `uuid` dependency from `package.json`.
- `src/providers/storage/indexeddb.js` no longer imports `uuid`; the single v4
  identifier it needs is now produced by a local `generateId()` helper built on
  the native Web Crypto API.

This is the only usage of `uuid` in the package:

```
$ grep -rn "from 'uuid'\|require('uuid')" src/
src/providers/storage/indexeddb.js:1
```

`uuid@9.0.1` is flagged by **CVE-2026-41907** (out-of-bounds write, CVSS 3.1 **7.5 High**).
The advisory covers **all versions < 14.0.0**, and the fix landed **only in 14.0.0**.

A straight version bump is not viable: `uuid@14.0.0` is **ESM-only** — its `package.json`
declares `"type": "module"` and its `exports` map has no `require` condition, so it cannot be
`require()`'d from the CommonJS build in `lib/cjs/`. That is the underlying reason the Snyk
PR #6306 could not be merged, leaving the dependency stuck on a flagged version.

Since `uuid` is used exactly once, and only to generate a v4 identifier, dropping the
dependency altogether resolves the situation at the source rather than deferring it:

- removes this class of advisory permanently, for every consumer of `@formio/js`;
- no ESM/CJS interop problem to work around;
- no bundle-size cost (one fewer transitive dependency).

The generated value remains a fully RFC 4122-compliant v4 UUID (same version and variant
bits), so stored records and their keys are unchanged in shape.

**None.**

- The generated identifier keeps the exact same format (RFC 4122 v4 string), so existing
  IndexedDB records and lookups by `id` are unaffected.
- `crypto.randomUUID()` is only exposed in **secure contexts** (HTTPS / `localhost`), whereas
  IndexedDB also works over plain HTTP. To avoid regressing those deployments, the helper
  falls back to `crypto.getRandomValues()` (available since IE11, no secure-context
  requirement) and formats the bytes manually:

```js
function generateId() {
  if (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function') {
    return crypto.randomUUID();
  }

  const bytes = new Uint8Array(16);
  crypto.getRandomValues(bytes);
  bytes[6] = (bytes[6] & 0x0f) | 0x40; // version 4
  bytes[8] = (bytes[8] & 0x3f) | 0x80; // variant RFC 4122

  const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0')).join('');
  return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
}
```

- `generateId()` is only reached inside `uploadFile()`, after the existing
  `if (!('indexedDB' in window))` guard, so it never runs outside a browser context.

Side note: the previous call passed a `Blob` as uuid's `options` argument
(`uuidv4(blobObject)`). A `Blob` provides neither `random` nor `rng`, so it was silently
ignored and the call behaved as `uuidv4()`. Behaviour is therefore unchanged.

## Dependencies

None. This PR only *removes* a dependency (`uuid`) and adds no new one.

- Verified `uuid` no longer appears anywhere in the source: `grep -rn "uuid" src/` returns nothing.
- Verified the generated identifiers are valid RFC 4122 v4 UUIDs (character 15 is `4`,
  character 20 is one of `8`, `9`, `a`, `b`), in both branches of `generateId()`
  (with and without `crypto.randomUUID` available).
- Exercised the IndexedDB storage provider end to end (upload → download → delete) to confirm
  records are stored and retrieved by the generated `id`.

No automated test was added: the IndexedDB provider has no existing unit-test coverage, and
testing it meaningfully requires a browser environment (IndexedDB + Web Crypto) rather than
the current JSDOM-based unit suite. Happy to add one if you can point me at the preferred
approach.

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,6 @@
     "signature_pad": "^4.2.0",
     "string-hash": "^1.1.3",
     "tippy.js": "^6.3.7",
-    "uuid": "^9.0.0",
     "vanilla-picker": "^2.12.3"
   },
   "devDependencies": {
--- a/src/providers/storage/indexeddb.js
+++ b/src/providers/storage/indexeddb.js
@@ -1,4 +1,27 @@
-import { v4 as uuidv4 } from 'uuid';
+
+/**
+ * Generates an RFC 4122 version 4 UUID.
+ *
+ * Uses the native `crypto.randomUUID()` when available. That API is only exposed
+ * in secure contexts (HTTPS or localhost), whereas IndexedDB also works over plain
+ * HTTP, so a `crypto.getRandomValues()` based fallback preserves the previous
+ * behaviour in those environments.
+ * @returns {string} A randomly generated v4 UUID.
+ */
+function generateId() {
+  if (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function') {
+    return crypto.randomUUID();
+  }
+
+  const bytes = new Uint8Array(16);
+  crypto.getRandomValues(bytes);
+  bytes[6] = (bytes[6] & 0x0f) | 0x40; // version 4
+  bytes[8] = (bytes[8] & 0x3f) | 0x80; // variant RFC 4122
+
+  const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0')).join('');
+  return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
+}
+
 /**
  * indexedDb provider for file storage.
  * @returns {import('./typedefs').FileProvider} The FileProvider interface defined in index.js.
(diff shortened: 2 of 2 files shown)


dahhou-ilyas:
@lane-formio This follows up on #6306, which was closed due to the risk of a breaking change. Rather than bumping `uuid`, this removes the dependency entirely since there is only a single call site. Happy to adjust the approach if you'd prefer another route.

lane-formio:
https://github.com/formio/formio.js/pull/6339

This has been resolved in our latest release 5.5.2.

Closing this PR.

PR #6340, opened 2026-09-07
Fix selectboxes selection validation

### LINKED ISSUE #6338: SelectBoxes.checkComponentValidity returns false without populating the caller's errors array, so required + minSelectedCount is never enforced
**Describe the bug**

`SelectBoxesComponent.checkComponentValidity()` returns `false` from its `min/maxSelectedCount` branch without appending anything to the `errors` array it was handed. Callers gate on that array rather than on the return value, so **a `selectboxes` with both `validate.required` and `validate.minSelectedCount` is never actually enforced**: `Webform.submitForm()` emits `submit` with an empty value, and `Wizard.nextPage()` advances past the page.

Counter-intuitively, `minSelectedCount` is what *breaks* `required` here. Remove `minSelectedCount` and `required` is enforced correctly.

`src/components/selectboxes/SelectBoxes.js#L280-L322` on `main`:

```js
checkComponentValidity(data, dirty, rowData, options, errors = []) {
    const minCount = this.component.validate.minSelectedCount;
    const maxCount = this.component.validate.maxSelectedCount;
    if (!this.shouldSkipValidation(data, rowData, options)) {
        const isValid = this.isValid(data, dirty);
        if (maxCount || minCount) {
            ...
            if (!isValid && maxCount && count > maxCount) {
                this.errors.push({ message });
                this.setCustomValidity(message, dirty);
                return false;                 // <-- `errors` (the caller's array) untouched
            } else if (!isValid && minCount && count < minCount) {
                this.errors.push({ message });
                this.setCustomValidity(message, dirty);
                return false;                 // <-- same
            }
        }
    }
    // the only line that does `allErrors.push(...)`:
    return super.checkComponentValidity(data, dirty, rowData, options, errors);
}
```

Two things combine:

1. Only `super.checkComponentValidity()` appends to the caller's `errors`. Both early returns skip it.
2. The branch is guarded on `!isValid`, and `Component.invalidMessage()` short-circuits with `if (!this.hasInput || (!dirty && this.pristine)) return ''`. So the branch is reached **only when `dirty` is truthy** — which is exactly what the enforcement paths pass (`Wizard.nextPage()` → `validateCurrentPage({ dirty: true })`, `Webform.submitForm()` → `validate(data, { dirty: true, process: 'submit' })`). The flag that is supposed to force validation is what suppresses the error.

Because `this.errors` *is* populated, the UI paints a message while the gate passes — the form displays "Please fix the following errors before submitting" **and emits `submit` in the same interaction**.

**Version/Branch**

Reproduced on `@formio/js@5.5.2` with `@formio/core@2.8.2` (current release), and on `5.4.1` / `2.7.1`. The code is unchanged on `main` — the only edit to this method since 5.4.1 was replacing the hardcoded messages with the `maxSelectItems` / `minSelectItems` i18n keys (#3606).

**To Reproduce**

Self-contained page, stock `formio.form.min.js`, no options and no configuration:

```html
<script src="formio.form.min.js"></script>
<div id="a"></div><div id="b"></div><div id="c"></div>
<script>
  var comp = function (withMin) {
    var v = { required: true };
    if (withMin) { v.minSelectedCount = 1; }
    return {
      type: 'selectboxes', key: 'picks', label: 'Pick at least one', input: true,
      dataSrc: 'values',
      values: [{ label: 'One', value: 'one' }, { label: 'Two', value: 'two' }],
      validate: v
    };
  };
  var submitBtn = { type: 'button', key: 'submit', label: 'Submit', action: 'submit', input: true };

  // A: required + minSelectedCount -> NOT enforced
  Formio.createForm(document.getElementById('a'),
    { display: 'form', components: [comp(true), submitBtn] })
    .then(function (f) {
      window.FA = f;
      f.on('submit', function () { console.log('A: submit fired with', JSON.stringify(f.submission.data.picks)); });
    });

  // B: required only -> enforced correctly
  Formio.createForm(document.getElementById('b'),
    { display: 'form', components: [comp(false), submitBtn] }).then(function (f) { window.FB = f; });

  // C: same component on wizard page 1
  Formio.createForm(document.getElementById('c'), { display: 'wizard', components: [
    { type: 'panel', key: 'p1', title: 'Page 1', components: [comp(true)] },
    { type: 'panel', key: 'p2', title: 'Page 2', components: [] }
  ]}).then(function (f) { window.FC = f; });
</script>
```

1. Select nothing in form **A** and click Submit → the `submit` event fires with `{"one":false,"two":false}` and `submission.state === 'submitted'`, while the alert simultaneously reads "Please fix the following errors before submitting".
2. Select nothing in wizard **C** and click Next → it advances to page 2.
3. Form **B** (no `minSelectedCount`) blocks correctly, which isolates `minSelectedCount` as the trigger.

Direct probe of the two forms, nothing selected:

```js
var probe = function (f) {
  var bucket = [];
  var ok = f.getComponent('picks').checkComponentValidity(f.data, true, null, { dirty: true }, bucket);
  return { returned: ok, pushedIntoCallersArray: bucket.map(function (e) { return e.message; }),
           formValidate: f.validate(f.data, { dirty: true }).map(function (e) { return e.message; }) };
};
probe(window.FA);
// { returned: false, pushedIntoCallersArray: [],                        formValidate: [] }
probe(window.FB);
// { returned: false, pushedIntoCallersArray: ["Pick at least one is required"],
//                                            formValidate: ["Pick at least one is required"] }
```

**Expected behavior**

`checkComponentValidity()` should append to the `errors` array whenever it returns `false`, so the caller's error list agrees with the return value. A required `selectboxes` with no option selected should block both `nextPage()` and `submit()`, with or without `minSelectedCount`.

**Additional context**

Note that `@formio/core`'s `validateMinimumSelectedCount` deliberately returns `null` when nothing is selected ("Should not be triggered if there are no options selected at all"), delegating the zero case to `validateRequired`. That delegation is sound — the override in `SelectBoxes.js` is what prevents `validateRequired`'s result from reaching the caller.

Suggested fix: have both early returns push the message into `errors` before returning, or fall through to `super.checkComponentValidity()` so the standard rules still run.

Workaround for anyone hitting this: drop `minSelectedCount` when the intent is simply "at least one" (`required` alone is enforced correctly), or wrap the prototype:

```js
const proto = Formio.Components.components.selectboxes.prototype;
const original = proto.checkComponentValidity;
const base = Object.getPrototypeOf(proto);
proto.checkComponentValidity = function (data, dirty, rowData, options, errors = []) {
  const before = errors.length;
  const valid = original.call(this, data, dirty, rowData, options, errors);
  if (!valid && errors.length === before) {
    base.checkComponentValidity.call(this, data, dirty, rowData, options, errors);
  }
  return valid;
};
```

This is worth prioritising: on a form where the client is the only validation layer, it silently accepts submissions that fail the form's own required rules.

mansikodgire11:
## Link to Jira Ticket

I am an external open-source contributor and do not have access to Form.io Jira.
The PR template requires a Jira link, so please assign or attach the appropriate Jira ticket ID to this PR.

### What changed?

Fixed SelectBoxes validation so `required`, `minSelectedCount`, and `maxSelectedCount` errors flow through the standard validation pipeline.

Added a regression test for a required SelectBoxes component with `minSelectedCount`.

### Why have you chosen this solution?

The previous implementation returned early and bypassed the standard validation error collection. The fix preserves input-disabling behavior while allowing core validation to produce the correct field errors.

This addresses [formio/formio.js#6338](https://github.com/formio/formio.js/issues/6338).

Fixes #6338


mansikodgire11:
sorry about that, new to open-source contribution

